PR #215, opened 2020-04-29
Tile cache disk

### LINKED ISSUE #210: Use a storage disk for the tile cache
Image tiles are currently stored locally in the tile cache directory. This does not work for a multi-host setup. Instead, a storage disk should be used, which could be cloud storage shared by multiple hosts. The setup can be the same than with image thumbnails which are served directly from the public path of the storage disk. The tile cache disk should still be pruned regularly so it does not contain too many files.

mzur:
Resolves #210 

Run this command after the upgrade:

```
php artisan update-tiled-images
```

It can be run in parallel, too:

```
php artisan update-tiled-images --workers 2 --worker 0
php artisan update-tiled-images --workers 2 --worker 1
```

### DIFF
--- a/app/Console/Commands/UpdateTiledImages.php
+++ b/app/Console/Commands/UpdateTiledImages.php
@@ -1,0 +1,120 @@
+<?php
+
+namespace Biigle\Console\Commands;
+
+use File;
+use Phar;
+use Storage;
+use PharData;
+use SplFileInfo;
+use Biigle\Image;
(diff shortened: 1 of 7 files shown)


mzur:
I decided to stick to zip archives as they are natively supported by VIPS and the only benefit of tar.gz (now that we can't use Swift archive auto extraction) is a slightly smaller archive size.

PR #219, opened 2020-05-20
Migrate to Webpack and ES6 modules

### LINKED ISSUE #213: Modernize JS codebase
Modernize the JS codebase to use modules and ES6 syntax with Laravel Mix/Webpack. Implement the current plugin system as follows:

Migrate everything that is not a Vue ViewModel to a JS module. Use ES6 syntax with `let`/`const` instead of `var`, `functionName()` instead of `function functionName()` and use of `=>` where appropriate. Implement a `mount` helper function that does a similar thing than the `biigle.$viewModel` function.

Export modules for use in plugins in an extra file, e.g. `exports.js` that can look like this:

```js
import {mount, declare, require} from './helpers';
import MyComponent from './components/MyComponent';

window.biigle.$mount = mount;
window.biigle.$declare = declare;
window.biigle.$require = require;

biigle.$declare('core.MyComponent', MyComponent);
// ...
```

Deprecate the `biigle.$component` helper function as it should no longer be required. Use of the function should display a warning in the console. Plugins can use external modules with `biigle.$require('core.MyComponent')`.

In the first step. the `$declare` keys should stay the same so the existing `$require`s still work. In a next step, component names and keys (and everything that is not a function) should be renamed to start with an uppercase character.

To make full use of the new capabilities, merge the OpenLayers rectangle hacks repository into the biigle/annotations codebase, since it can be used with only a few extensions of core modules and classes. Then archive our OpenLayers fork. Maybe upgrade OpenLayers to the latest version in one go (biigle/annotations#134).

mzur:
Resolves #213 

Update the modules:

- [x] [biigle/ananas](https://github.com/biigle/ananas)
- [x] [biigle/annotations](https://github.com/biigle/annotations)
- [x] [biigle/color-sort](https://github.com/biigle/color-sort)
- [x] ~~[biigle/demo](https://github.com/biigle/demo)~~ (no JS)
- [x] [biigle/geo](https://github.com/biigle/geo)
- [x] [biigle/label-trees](https://github.com/biigle/label-trees)
- [x] [biigle/largo](https://github.com/biigle/largo)
- [x] [biigle/laserpoints](https://github.com/biigle/laserpoints)
- [x] [biigle/maia](https://github.com/biigle/maia)
- [x] [biigle/module](https://github.com/biigle/module)
- [x] [biigle/projects](https://github.com/biigle/projects)
- [x] [biigle/reports](https://github.com/biigle/reports)
- [x] [biigle/sync](https://github.com/biigle/sync)
- [x] [biigle/videos](https://github.com/biigle/videos)
- [x] [biigle/volumes](https://github.com/biigle/volumes)

Other tasks:

- [x] Migrate biigle/core
- [x] ~~Fix warnings caused by VueStrap~~
- [x] Check dangling calls to `biigle.$require` and `ol`
- [x] Walk through JS files and validate all features
- [x] Configure ESLint ( #223 )
- [x] Migrate to Vue single file components


mzur:
The warning `This page uses the non standard property “zoom”.` is caused by Bootstrap and can only be fixed by upgrading to Bootstrap 4.

The deprecation notices for `onmozfullscreenchange` and `onmozfullscreenerror` probably come from some dependencies of VueStrap. Maybe we can ditch VueStrap and implement the few required components ourselves, as the repository is no longer maintained.

PR #277, opened 2020-07-15
Video Volumes

### LINKED ISSUE #265: Whole video labels
Users requested the addition of whole video labels (similar to image labels). These are different to the proposed whole frame labels ( #270 ). This is related to biigle/videos#68.

### LINKED ISSUE #276: Video volumes
Some users have a video annotation workflow similar to the image annotation workflow where lots of (short) videos need to be annotated. The following features would facilitate this workflow:

- Whole video labels (like image labels)
- Skipping back and forth through videos in the video annotation tool
- A video overview (like the volume overview) where videos can be browsed, sorted, filtered and labelled

My idea to solve this would be the introduction of "video volumes" as an alternative to single videos. Video volumes are collections of videos just like (image) volumes are collections of images. If added, we need to rename "volumes" to "image volumes" to avoid confusion. Video volumes are displayed in the list of videos on the project overview next to regular (single) videos. If a video volume is clicked, the video volume overview is opened, which is analogous to the image volume overview (including sorting, filtering and setting of video labels but minus features such as Largo, MAIA, ...). There, videos can be browsed and a click on the video opens the video annotation tool. If a video belongs to a video collection, the video annotation tool displays controls to switch to the previous/next video.

mzur:
Run `php artisan migrate` after upgrading to the new version.

Resolves #276
Resolves #265


mzur:
Run `php artisan migrate` after upgrading to the new version.

PR #344, opened 2021-05-07
On Shift+Scroll you get next and previous image

jLeonBi:


resources/assets/js/annotations/annotatorContainer.vue
```diff
@@ -439,18 +439,28 @@ export default {
...
+            const imagesToLoad = 10;
+            // Array with all promises
+            let idsToCache = [];
+
             let previousId = this.imagesIds[this.getPreviousIndex(this.imageIndex)];
-            let nextId = this.imagesIds[this.getNextIndex(this.imageIndex)];
+
+            // load next X ids and push Annotation and Image in array for promise all
+            for (const x of Array(imagesToLoad).keys()) {
+                const currentId = this.imagesIds[this.getNextIndex(this.imageIndex + x)]
```
mzur, line 451: Does this also handle the case where `this.imageIndex + x` is larger than the actual number of images?

resources/views/annotations/show/annotationCanvas.blade.php
```diff
@@ -1,7 +1,7 @@
-<div class="annotation-canvas">
-    <minimap v-if="showMinimap" :extent="extent"></minimap>
-    <div class="annotation-canvas__left-indicators">
-        <scale-line-indicator v-if="showScaleLine" :image="image" :areas="imagesArea" :resolution="resolution" inline-template>
+<div class="annotation-canvas" v-on:wheel="handleScroll">
+    <minimap v-if="showMinimap" :extent="extent" ></minimap>
+    <div class="annotation-canvas__left-indicators" >
+        <scale-line-indicator v-if="showScaleLine" :image="image" :areas="imagesArea" :resolution="resolution" inline-template >
```
mzur, line 4: Were these whitespace changes done on purpose?

mzur (commented):
So this PR implements both the "Shift+Scroll" and the "load x next images" features?

mzur (commented):
Also can you please add the <kbd>Shift</kbd>+Scroll shortcut to [the manual](https://github.com/biigle/core/blob/086357ad2c6ad156503084a9b013ecf771173aef/resources/views/manual/tutorials/annotations/shortcuts.blade.php#L26)?

resources/assets/js/annotations/annotatorContainer.vue
jLeonBi: > So this PR implements both the "Shift+Scroll" and the "load x next images" features?

Unfortunately yes. This was a git mistake on my part. 

> Does this also handle the case where this.imageIndex + x is larger than the actual number of images?

Yes it works fine for one image or less than should be loaded (currently 10). 
I currently change the code, so if the amount of images to load is greater than the amount of images, it only loads as many as there are.

resources/views/annotations/show/annotationCanvas.blade.php
jLeonBi: No i removed them

jLeonBi:
> Also can you please add the Shift+Scroll shortcut to [the manual](https://github.com/biigle/core/blob/086357ad2c6ad156503084a9b013ecf771173aef/resources/views/manual/tutorials/annotations/shortcuts.blade.php#L26)?

I updated the manual, i hope the shortcut instructions are understandable.

resources/assets/js/annotations/annotatorContainer.vue
```diff
@@ -439,18 +439,31 @@ export default {
             Events.$emit('images.change', this.imageId, this.image);
         },
         cachePreviousAndNext() {
+            // currently const number of images to cache
+            let imagesToLoad = 10;
+            if (imagesToLoad > this.allImagesIds.length) {
```
mzur, line 444: I checked and you don't actually need to handle this special case. `getNextIndex` just wraps around if the index gets too large. This is desired behavior, too, as these images are also next in the "queue" and should be cached.

jLeonBi:
> I also updated your branch so it no longer has conflicts to `master`.

Thanks for doing this.

resources/assets/js/annotations/annotatorContainer.vue
```diff
@@ -53,6 +53,7 @@ export default {
             lastCreatedAnnotation: null,
             lastCreatedAnnotationTimeout: null,
             annotationOpacity: 1,
+            cachedImages: 1,
```
mzur, line 56: Could you please rename this to `cachedImagesCount`?

resources/assets/js/annotations/annotatorContainer.vue
```diff
@@ -552,6 +567,9 @@ export default {
                     .finally(this.finishLoading);
             }
         },
+        cachedImages() {
+            this.cachePreviousAndNext();
```
mzur, line 571: This could make use of the `debounce` function so the backend is not spammed with requests for images while the user interacts with the slider. `debounce` waits for a specific time until it is no longer called before a specified callback function is executed. I set 100 ms here but maybe we could even increase that to 500 ms.

```suggestion
            debounce(this.cachePreviousAndNext, 100, 'annotations.cached-image-count.update');
```

resources/assets/js/annotations/stores/settings.js
```diff
@@ -9,6 +9,7 @@ import Settings from '../../core/models/Settings';
 let defaults = {
     // Settings tab.
     annotationOpacity: 1.0,
+    cachedImages: 1,
```
mzur, line 12: This needed to be added, too. I forgot about that myself. It's all quite redundant...

resources/views/annotations/show/annotationCanvas.blade.php
```diff
@@ -1,7 +1,7 @@
-<div class="annotation-canvas">
+<div class="annotation-canvas" v-on:wheel="handleScroll">
     <minimap v-if="showMinimap" :extent="extent"></minimap>
     <div class="annotation-canvas__left-indicators">
-        <scale-line-indicator v-if="showScaleLine" :image="image" :areas="imagesArea" :resolution="resolution" inline-template>
+        <scale-line-indicator v-if="showScaleLine" :image="image" :areas="imagesArea" :resolution="resolution" inline-template >
```
mzur, line 4:
```suggestion
        <scale-line-indicator v-if="showScaleLine" :image="image" :areas="imagesArea" :resolution="resolution" inline-template>
```

resources/views/annotations/show/tabs/settings.blade.php
```diff
@@ -13,6 +13,11 @@
                 <input type="range" min="0" max="1" step="0.1" v-model="annotationOpacity">
             </div>
 
+            <div class="sidebar-tab__section">
+                <h5 title="Set the number of caches images ">Cached Images (<span v-text="cachedImages"></span>)</h5>
+                <input type="range" min="1" max="10" step="1" v-model="cachedImages">
```
mzur, line 18:
```suggestion
                <input type="range" min="1" max="50" step="1" v-model="cachedImages">
```

mzur (changes requested):
Please apply the few remaining changes. I will take care of the rest.

As a side note: I like to follow [this guide](https://chris.beams.io/posts/git-commit/) for writing my git commit messages. This keeps the git history clean and readable. It's just a personal preference, of course.

resources/assets/js/annotations/annotatorContainer.vue
jLeonBi: I just copy and paste your suggestion

mzur (approved):
Looks good, thanks :+1:

I just reduced the maximum number to 50 because the total cache size is 100. If 100 images are loaded with a maximum cache size of 100, the next image (which is the most important one) would be the first image to be removed from the cache (e.g. if the user does one step back). We don't want that.

PR #467, opened 2022-07-06
Fix API-Documentation {id}/files/filter/annotation-user/{id2}

tmoeller4:
Fix documentation for API-call to get all images having annotations of a user


mzur (approved):
Thanks!

PR #524, opened 2022-12-09
Added graphics to the 404 error page

dlangenk:
Added a small image to the 404 error page (and changed wording accordingly) which can be used until we have something better. Related to #2


dlangenk:
Won't implement that. Will be superseded by something more fancy

Issue #496, opened 2022-10-31
Improve project/volume charts for extreme cases


mzur:
Have a look at the project charts of project 555. We should improve the charts to handle such an extreme case more gracefully. Ideas:

- Timeline: Show only the users with 90% of the annotations (if there are more than, say, 5 users). The rest is shown as "other users".
- Label pie chart: Show only the labels with 90%  of the annotations (if there are more than, say, 20 labels). The rest is shown as "other labels".
- NetMap display: Show only the 20 (50/100?) most frequent labels. Display a notice if some labels are not shown because of this.

dlangenk:
I actually like that all the information is shown for the timeline and label pie charts. However, it would be nicer if it would be sorted by abundance. This way you don't have to get rid of information but it would still be well readable.

For the NetMap I totally agree, because it is not readable anymore. We should probably try out if 20/50/100 works best.

dlangenk:
For me this is working (proof via mail to not disclose any information) but I agree that it there is some point after which we have to take measures. Don't know if the library allows for a hierarchical display on hover, i.e. on hover it shows all the names instead of other. If not then a large enough value will also do.

Yes that is exactly what I meant. Sorting the pie pieces by size, i.e. the number of annotations per user.

PR #525, opened 2022-12-09
Improved lbl pie chart

dlangenk:
Partly implements #496.

- sort Label Pie Chart based on the abundance
- activate scrolling legend at the bottom of chart, thus allowing the filtering of the chart.


mzur (approved):
Thanks!

PR #526, opened 2022-12-09
Clone volume

lehecht:
* Cloning files and their (annotation) labels is optional now
* FormRequest is used for request validation


app/Http/Requests/CloneVolume.php
```diff
@@ -0,0 +1,171 @@
...
+
+                    //check file ids
+                    if (!empty($annotationLabelFileIds) &&
+                        !$this->filesBelongToVolume($annotationLabelFileIds, $volume->id, $volume->isImageVolume())) {
+                        $validator->errors()->add('$fileLabelFileIds',
+                            'Unauthorized access to annotation labels that do not belong to the volume');
+                    }
+                } catch (Exception $e) {
+                    $validator->errors()->add('$fileLabelFileIds', $e->getMessage());
+                }
```
mzur, line 141: There is a little misconception here. You expect IDs of the `image_labels`/`video_labels` tables in the `file_label_ids` parameter. Also, you expect IDs of the `image_annotation_labels`/`video_annotation_labels` tables in the `label_ids` parameter. But in both cases you should rather expect IDs of the `labels` table. Here are two examples to visualize this:

Filtering of image labels

```php
$labelIds = $this->input('file_label_ids', []);
// Get all image labels of a volume that a certain label_id
$imageLabels = ImageLabel::join('images', 'images.id', '=', 'image_annotations.image_id')
   ->where('images.volume_id', $volume->id)
   ->whereIn('label_id', $labelIds)
   ->get();
```

Filtering of annotations

```php
$labelIds = $this->input('annotation_label_ids', []);
// Get all annotations of a volume that have labels with a certain label_id
$annotations = ImageAnnotation::join('images', 'images.id', '=', 'image_annotations.image_id')
   ->where('images.volume_id', $volume->id)
   ->with('labels')
   ->whereHas('labels', function ($query) use ($labelIds) {
      $query->whereIn('label_id', $labelIds)
   })
   ->get();
```

(more on [`whereHas()`](https://laravel.com/docs/9.x/eloquent-relationships#querying-relationship-existence))

So there is no need to check if the ImageLabels/VideoLabels/ImageAnnotationLabels/VideoAnnotationLabels actually belong to the volume here. You just filter by the `label_id`. If there are no ImageLabels/VideoLabels/ImageAnnotationLabels/VideoAnnotationLabels with a particular label ID, none will be copied.

mzur (changes requested):
There are two bigger points that I raise in the review comments:

1. The API might be easier to use if you copy all images by default (if no file IDs are given). The same could be done for annotations and file labels (with a additional boolean parameters).

2. The `file_label_ids` and `label_ids` should be the IDs of the `labels` table and not the `image_labels` or `image_annotation_labels` table (which both have the label IDs as `label_id` foreign key).

Also, you should add some test cases for the new behavior.

package-lock.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "core",
+  "name": "biigle",
```
mzur, line 2: This change should be reverted.

app/Http/Controllers/Api/VolumeController.php
```diff
@@ -224,6 +243,37 @@ public function clone($volumeId, $destProjectId, Request $request)
...
+        } else {
+            $labels = $isImageVol ? ImageAnnotationLabel::find($labelIds) : VideoAnnotationLabel::find($labelIds);
+            //get annotation ids
+            $annotationIds = $labels->pluck('annotation_id');
+            //get annotations
+            $annotations = $isImageVol ? ImageAnnotation::find($annotationIds) : VideoAnnotation::find($annotationIds);
+            return $annotations->filter(function ($label) use ($fileIds, $fileType) {
+                //filter annotations whose files were not copied
+                return in_array($label->$fileType, $fileIds);
+            })->pluck('id')->toArray();
```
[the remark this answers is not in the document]
lehecht: Is a sorting by id also needed here?

app/Http/Requests/CloneVolume.php
lehecht: So the whole withValidator function can be deleted?

app/Http/Controllers/Api/VolumeController.php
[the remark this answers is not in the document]
mzur: No I don't think so. But please check if the returned values are what you expect. In the old version and my suggested alternative you get annotation IDs. But maybe you expect annotation label IDs? Please update the docblock accordingly.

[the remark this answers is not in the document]
mzur: If I see that correctly, this is basically the same query than in `copyImageAnnotation()` and `copyVideoAnnotation()` below. Do you really need the query both times?

app/Http/Requests/CloneVolume.php
mzur: Not completely. It still makes sense to check if the `file_ids` all belong to the volume.

app/Http/Controllers/Api/VolumeController.php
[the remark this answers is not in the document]
lehecht: Well, if I have not forgotten anything, then all copyVideo and copyImage methods are almost equal except for their file types. I thought that was the way it was supposed to be. But I can merge them, if you want.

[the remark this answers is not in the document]
mzur: No, it's fine to have separate methods for images and videos. What I meant was that the whole `filterLabels()` method may be unnecessary. You do basically the same query that is done in `filterLabels()` again in `copyImageAnnotation()` and `copyVideoAnnotation()`.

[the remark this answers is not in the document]
lehecht: Ah yes, I deleted it already because filtering is more efficient with your approach.

app/Http/Controllers/Api/VolumeController.php
```diff
@@ -157,22 +158,27 @@ public function update(UpdateVolume $request)
...
      *
-     * @apiParam {Number} id The volume ID.
-     * @apiParam {Number} project_id The target project ID.
-     * @apiParam (image IDs that can be used for cloning) {Array} imageIds selected subset of image IDs from original volume. Only images from this subset will be cloned.
-     * @apiParam (video IDs that can be used for cloning) {Array} videoIds selected subset of video IDs from original volume. Only videos from this subset will be cloned.
+     * @apiParam {Number} id The volume id.
+     * @apiParam {Number} project_id The target project id.
+     * @apiParam {string} name volume name of cloned volume.
+     * @apiParam (file ids) {Array} only_files ids of files which should be cloned. If empty all files are cloned.
+     * @apiParam {bool} clone_annotations Switch to clone annotation labels.
```
mzur, line 178:
```suggestion
     * @apiParam {bool} clone_annotations Set to `true` to also clone annotations.
```

app/Http/Controllers/Api/VolumeController.php
```diff
@@ -157,22 +158,27 @@ public function update(UpdateVolume $request)
...
+     * @apiParam (annotation label ids) {Array} only_annotation_labels ids of annotation labels which should be cloned. If empty all labels are cloned.
+     * @apiParam {bool} clone_file_labels Switch to clone file labels.
```
mzur, line 180:
```suggestion
     * @apiParam {bool} clone_file_labels Set to `true` to also clone image/video labels.
```

app/Http/Controllers/Api/VolumeController.php
```diff
@@ -157,22 +158,27 @@ public function update(UpdateVolume $request)
...
-     * @apiParam {Number} id The volume ID.
-     * @apiParam {Number} project_id The target project ID.
-     * @apiParam (image IDs that can be used for cloning) {Array} imageIds selected subset of image IDs from original volume. Only images from this subset will be cloned.
-     * @apiParam (video IDs that can be used for cloning) {Array} videoIds selected subset of video IDs from original volume. Only videos from this subset will be cloned.
+     * @apiParam {Number} id The volume id.
+     * @apiParam {Number} project_id The target project id.
+     * @apiParam {string} name volume name of cloned volume.
+     * @apiParam (file ids) {Array} only_files ids of files which should be cloned. If empty all files are cloned.
+     * @apiParam {bool} clone_annotations Switch to clone annotation labels.
+     * @apiParam (annotation label ids) {Array} only_annotation_labels ids of annotation labels which should be cloned. If empty all labels are cloned.
```
mzur, line 179:
```suggestion
     * @apiParam (annotation label ids) {Array} only_annotation_labels Label IDs to filter cloned annotations. If empty, all annotations are cloned. Only has an effect if `clone_annotations` is `true`.
```

app/Http/Controllers/Api/VolumeController.php
```diff
@@ -157,22 +158,27 @@ public function update(UpdateVolume $request)
...
+     * @apiParam {bool} clone_file_labels Switch to clone file labels.
+     * @apiParam (file label ids) {Array} only_file_labels ids of file labels which should be cloned. If empty all labels are cloned.
```
mzur, line 181:
```suggestion
     * @apiParam (file label ids) {Array} only_file_labels Label IDs to filter cloned file labels. If empty, all file labels are cloned. Only has an effect if `clone_file_labels` is `true`.
```

app/Http/Controllers/Api/VolumeController.php
```diff
@@ -256,18 +276,53 @@ private function copyImages($volume, $copy, $selectedImageIds)
...
+        } else {
+            $imageAnnotationLabelIds = $selectedLabelIds;
+        }
+
         $chunkSize = 100;
         $newImageIds = $copy->images()->orderBy('id')->pluck('id');
         $volume->images()
+            ->with([
+                'annotations' => fn($q) => $q->whereIn('id', $usedAnnotationIds),
+                'annotations.labels' => fn($q) => $q->whereIn('label_id', $imageAnnotationLabelIds),
```
mzur, line 314: I feel that we also need an `orderBy('id')` here.
```suggestion
                'annotations' => fn($q) => $q->whereIn('id', $usedAnnotationIds)->orderBy('id'),
                'annotations.labels' => fn($q) => $q->whereIn('label_id', $imageAnnotationLabelIds)->orderBy('id'),
```

app/Http/Controllers/Api/VolumeController.php
```diff
@@ -370,19 +428,52 @@ private function copyVideos($volume, $copy, $selectedVideoIds)
...
+        } else {
+            $videoAnnotationLabelIds = $selectedLabelIds;
+        }
+
         $chunkSize = 100;
         $newVideoIds = $copy->videos()->orderBy('id')->pluck('id');
         $volume->videos()
+            ->with([
+                'annotations' => fn($q) => $q->whereIn('id', $usedAnnotationIds),
+                'annotations.labels' => fn($q) => $q->whereIn('label_id', $videoAnnotationLabelIds),
```
mzur, line 465:
```suggestion
                'annotations' => fn($q) => $q->whereIn('id', $usedAnnotationIds)->orderBy('id'),
                'annotations.labels' => fn($q) => $q->whereIn('label_id', $videoAnnotationLabelIds)->orderBy('id'),
```

mzur (changes requested):
Except for the tests, only minor things are left. Please add the following test cases:

- clone volume with image/video labels and image/video annotations but without `clone_file_labels = true` and without `clone_annotations = true` to check if the file labels and annotations really aren't cloned (can be added to `testCloneVolume` plus one new test case for the videos)

- clone volume images/videos with `clone_file_labels = true` and a non-empty `only_file_labels` to check if the file label filtering works (2 new test cases, can be adapted from `testCloneVolumeImages` and `testCloneVolumeVideos` but without the extra tests for the file attributes)

- clone volume image/video annotations with `clone_annotations = false` and a non-empty `only_annotation_labels` to check if the annotation label filtering works (2 new test cases, can be adapted from `testCloneVolumeImageAnnotations` and `testCloneVolumeVideoAnnotations` but without the extra tests for the annotation attributes)

tests/php/Jobs/CloneImagesOrVideosTest.php
```diff
@@ -0,0 +1,458 @@
...
+        );
+
+        $ignore = ['id', 'annotation_id'];
+        $this->assertEquals(
+            $oldAnnotationLabel->makeHidden($ignore)->toArray(),
+            $newAnnotationLabel->makeHidden($ignore)->toArray()
+
+        );
+
+        // second test
```
mzur, line 398: Please do this in a separate test function.

app/Jobs/CloneImagesOrVideos.php
```diff
@@ -0,0 +1,480 @@
...
+
+
+    /**
+     * Create a new job instance.
+     *
+     * @param CloneVolume $request containing the project, volume, new volume name and ids of files and labels.
+     *
+     * @return void
+     */
+    public function __construct($request)
```
mzur, line 93: I think it makes sense to expect the `CloneVolume` request here. If it causes problems in the tests, use `new CloneVolume(...)` there instead of `new Request(...)`.

```suggestion
    public function __construct(CloneVolume $request)
```

app/Http/Controllers/Api/VolumeController.php
```diff
@@ -186,284 +182,32 @@ public function update(UpdateVolume $request)
...
 
-            $project->addVolumeId($copy->id);
+            // If too many files should be created, do this asynchronously in the
+            // background. Else the script will run in the 30 s execution timeout.
+            $job = new CloneImagesOrVideos($request);
+            if (($nbrFiles === 0 && $volume->files()->count() > 0) || $nbrFiles > $createSyncLimit) {
+                Queue::pushOn('high', $job);
+            } else {
+                Queue::connection('sync')->push($job);
+            }
```
mzur, line 202: I think we sould always push the job on the normal (non-sync) queue. With websockets we now have a way to deal with this in the UI. This wasn't possible when the async volume creation was implemented.

app/Http/Controllers/Api/VolumeController.php
```diff
@@ -186,284 +182,32 @@ public function update(UpdateVolume $request)
...
-            if ($volume->hasIfdo()) {
-                $this->copyIfdoFile($volumeId, $copy->id);
-            }
+            $nbrFiles = count($request->input('only_files', []));
 
-            $project->addVolumeId($copy->id);
+            // If too many files should be created, do this asynchronously in the
+            // background. Else the script will run in the 30 s execution timeout.
+            $job = new CloneImagesOrVideos($request);
+            if (($nbrFiles === 0 && $volume->files()->count() > 0) || $nbrFiles > $createSyncLimit) {
```
mzur, line 198: I also don't really grok the `if` statement. It's false if the volume is empty or the files should be filtered and the number of filtered files is below the limit? But since the job should be pushed to the regular queue anyway, we don't need this any more.

app/Jobs/CloneImagesOrVideos.php
```diff
@@ -0,0 +1,480 @@
...
+     * @var bool
+     **/
+    public bool $cloneFileLabels;
+
+    /**
+     * Array containing file label ids.
+     *
+     * @var array
+     **/
+    public array $onlyFileLabels;
```
mzur, line 83: Please also add this:

```php
/**
 * Ignore this job if the project or volume does not exist any more.
 *
 * @var bool
 */
protected $deleteWhenMissingModels = true;
```

The project and volume models will be serialized when the job is pushed to the queue. Later when the job is processed and one of the models doesn't exist any more (because it was deleted in the meantime), there would be an error if the variable above is not set to `true`. With the variable `true` the job will simply be deleted in that case.

app/Jobs/CloneImagesOrVideos.php
```diff
@@ -0,0 +1,480 @@
...
+            }
+
+            //save ifdo-file if exist
+            if ($volume->hasIfdo()) {
+                $this->copyIfdoFile($volume->id, $copy->id);
+            }
+
+            $project->addVolumeId($copy->id);
+
+            $copy->flushThumbnailCache();
```
mzur, line 149: Why is this needed?

tests/php/Http/Controllers/Api/VolumeControllerTest.php
```diff
@@ -257,243 +246,8 @@ public function testCloneVolume()
...
-
-        $response = $this->post("/api/v1/volumes/{$volume->id}/clone-to/{$project->id}");
-        $response->assertStatus(302);
-        $copy = $project->volumes()->first();
-
-        $this->assertNotNull($copy->getIfdo());
-        $this->assertTrue($copy->hasIfdo());
-        $this->assertEquals($volume->getIfdo(), $copy->getIfdo());
+        $response = $this->postJson("/api/v1/volumes/{$volume->id}/clone-to/{$project->id}");
+        $response->assertStatus(200);
```
mzur, line 250: Please also test here if the `CloneImagesOrVideos` job was submitted.

app/Jobs/CloneImagesOrVideos.php
```diff
@@ -0,0 +1,480 @@
...
+use Biigle\VideoLabel;
+use Biigle\Volume;
+use \Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Queue\InteractsWithQueue;
+use Illuminate\Queue\SerializesModels;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Storage;
+use Ramsey\Uuid\Uuid;
+
+class CloneImagesOrVideos extends Job implements ShouldQueue
```
mzur, line 24: You could also call the job just `CloneVolume`.

tests/php/Jobs/CloneImagesOrVideosTest.php
```diff
@@ -0,0 +1,458 @@
...
+    public function testCloneVolume()
+    {
+        $volume = $this->volume(
+            ['created_at' => '2022-11-09 14:37:00',
+                'updated_at' => '2022-11-09 14:37:00',])->fresh(); // Use fresh() to load even the null fields.
+        // The target project.
+        $project = ProjectTest::create();
+
+        $this->beAdmin();
+        $project->addUserId($this->admin()->id, Role::adminId());
```
mzur, line 36: You don't need this any more. This is only required for the authorization checks when the API requests are tested (in `VolumeControllerTest`).

tests/php/Jobs/CloneImagesOrVideosTest.php
```diff
@@ -0,0 +1,458 @@
...
+
+        $volume = VolumeTest::create(
+            ['created_at' => '2023-01-09 14:37:00',
+                'updated_at' => '2023-01-09 14:37:00',
+                'media_type_id' => MediaType::videoId()])->fresh(); // Use fresh() to load even the null fields.
+        // The target project.
+        $project = ProjectTest::create();
+
+        $this->beAdmin();
+        $project->addUserId($this->admin()->id, Role::adminId());
```
mzur, line 76:
```suggestion
```

tests/php/Jobs/CloneImagesOrVideosTest.php
```diff
@@ -0,0 +1,458 @@
...
+
+        $oldVideo = VideoTest::create([
+            'filename' => 'a321123.jpg',
+            'taken_at' => [Carbon::now()->setTimezone('Europe/Lisbon')],
```
mzur, line 80:
```suggestion
            'taken_at' => Carbon::now()->setTimezone('Europe/Lisbon'),
```

tests/php/Jobs/CloneImagesOrVideosTest.php
```diff
@@ -0,0 +1,458 @@
...
+        $volume = $this->volume([
+            'media_type_id' => MediaType::imageId(),
+            'created_at' => '2022-11-09 14:37:00',
+            'updated_at' => '2022-11-09 14:37:00',
+        ])->fresh(); // Use fresh() to load even the null fields.
+        // The target project.
+        $project = ProjectTest::create();
+
+        $this->beAdmin();
+        $project->addUserId($this->admin()->id, Role::adminId());
```
mzur, line 109:
```suggestion
```

tests/php/Jobs/CloneImagesOrVideosTest.php
```diff
@@ -0,0 +1,458 @@
...
+            'tiled' => true])->fresh();
+        ImageLabelTest::create(['image_id' => $oldImage->id]);
+        $oldImageLabel = $oldImage->labels()->first();
+
+        $request = new Request(['project' => $project, 'volume' => $volume, 'clone_file_labels' => true]);
+
+        Queue::fake();
+        $this->expectsEvents('volume.cloned');
+        with(new CloneImagesOrVideos($request))->handle();
+        Queue::assertPushed(ProcessNewVolumeFiles::class);
```
mzur, line 126: Maybe also check in `testCloneVolume` that the job was *not* pushed?

tests/php/Jobs/CloneImagesOrVideosTest.php
```diff
@@ -0,0 +1,458 @@
...
+        $volume = VolumeTest::create([
+            'media_type_id' => MediaType::imageId(),
+            'created_at' => '2022-11-09 14:37:00',
+            'updated_at' => '2022-11-09 14:37:00',
+        ])->fresh(); // Use fresh() to load even the null fields.
+        // The target project.
+        $project = ProjectTest::create();
+
+        $this->beAdmin();
+        $project->addUserId($this->admin()->id, Role::adminId());
```
mzur, line 164:
```suggestion
```

tests/php/Jobs/CloneImagesOrVideosTest.php
```diff
@@ -0,0 +1,458 @@
...
+        $request = new Request(['project' => $project, 'volume' => $volume, 'clone_file_labels' => true,
+            'only_file_labels' => [$l2->label_id, $l3->label_id]]);
+
+        with(new CloneImagesOrVideos($request))->handle();
+
+        $copy = $project->volumes()->first();
+        $newImage = $copy->images()->first();
+        $newImageLabel = $newImage->labels()->get();
+
+        $this->assertEquals(2, $newImageLabel->count());
```
mzur, line 181: You should also check if the label IDs of the cloned image labels are correct. E.g.:
```php
$newImageLabels = $newImage->labels()->pluck('label_id');
$this->assertCount(2, $newImageLabels);
$this->assertContains($l2->label_id, $newImageLabels);
$this->assertContains($l3->label_id, $newImageLabels);
```

tests/php/Jobs/CloneImagesOrVideosTest.php
```diff
@@ -0,0 +1,458 @@
...
+        $volume = $this->volume([
+            'created_at' => '2022-11-09 14:37:00',
+            'updated_at' => '2022-11-09 14:37:00',
+            'media_type_id' => MediaType::videoId()
+        ])->fresh(); // Use fresh() to load even the null fields.
+        // The target project.
+        $project = ProjectTest::create();
+
+        $this->beAdmin();
+        $project->addUserId($this->admin()->id, Role::adminId());
```
mzur, line 195:
```suggestion
```

tests/php/Jobs/CloneImagesOrVideosTest.php
```diff
@@ -0,0 +1,458 @@
...
+
+        $oldVideo = VideoTest::create([
+            'filename' => 'a.jpg',
+            'taken_at' => [Carbon::now()->setTimezone('Europe/Lisbon')],
```
mzur, line 199:
```suggestion
            'taken_at' => Carbon::now()->setTimezone('Europe/Lisbon'),
```

tests/php/Jobs/CloneImagesOrVideosTest.php
```diff
@@ -0,0 +1,458 @@
...
+        $volume = VolumeTest::create([
+            'created_at' => '2022-11-09 14:37:00',
+            'updated_at' => '2022-11-09 14:37:00',
+            'media_type_id' => MediaType::videoId()
+        ])->fresh(); // Use fresh() to load even the null fields.
+        // The target project.
+        $project = ProjectTest::create();
+
+        $this->beAdmin();
+        $project->addUserId($this->admin()->id, Role::adminId());
```
mzur, line 250:
```suggestion
```

tests/php/Jobs/CloneImagesOrVideosTest.php
```diff
@@ -0,0 +1,458 @@
...
+        $request = new Request(['project' => $project, 'volume' => $volume, 'clone_file_labels' => true,
+            'only_file_labels' => [$l2->label_id, $l3->label_id]]);
+
+        with(new CloneImagesOrVideos($request))->handle();
+
+        $copy = $project->volumes()->first();
+        $newVideo = $copy->videos()->first();
+        $newVideoLabels = $newVideo->labels()->get();
+
+        $this->assertEquals(2, $newVideoLabels->count());
```
mzur, line 267: Same than with image labels above. Please check the `label_id`, too.

tests/php/Jobs/CloneImagesOrVideosTest.php
```diff
@@ -0,0 +1,458 @@
...
+            'updated_at' => '2022-11-09 14:37:00',
+        ])->fresh(); // Use fresh() to load even the null fields.
+        // The target project.
+        $project = ProjectTest::create();
+
+        $oldImage = ImageTest::create(['volume_id' => $volume->id])->fresh();
+        $oldAnnotation = ImageAnnotationTest::create(['image_id' => $oldImage->id]);
+        $oldAnnotationLabel = ImageAnnotationLabelTest::create(['annotation_id' => $oldAnnotation->id]);
+        $this->beAdmin();
+        $project->addUserId($this->admin()->id, Role::adminId());
```
mzur, line 284:
```suggestion
```

tests/php/Jobs/CloneImagesOrVideosTest.php
```diff
@@ -0,0 +1,458 @@
...
+            'only_annotation_labels' => [$l2->label_id, $l3->label_id]]);
+
+        with(new CloneImagesOrVideos($request))->handle();
+
+        $copy = $project->volumes()->first();
+        $newImage = $copy->images()->first();
+        $newAnnotation = $newImage->annotations()->first();
+        $newAnnotationLabels = $newAnnotation->labels()->get();
+
+        $this->assertEquals(2, $newAnnotationLabels->count());
```
mzur, line 343: Please also check the `label_id` of the annotation labels (similar to image/video labels above).

tests/php/Jobs/CloneImagesOrVideosTest.php
```diff
@@ -0,0 +1,458 @@
...
+        ])->fresh(); // Use fresh() to load even the null fields.
+        // The target project.
+        $project = ProjectTest::create();
+
+        $oldVideo = VideoTest::create(['volume_id' => $volume->id])->fresh();
+        $oldAnnotation = VideoAnnotationTest::create(['video_id' => $oldVideo->id]);
+        $oldAnnotationLabel = VideoAnnotationLabelTest::create(['annotation_id' => $oldAnnotation->id]);
+
+        $this->beAdmin();
+        $project->addUserId($this->admin()->id, Role::adminId());
```
mzur, line 362:
```suggestion
```

tests/php/Jobs/CloneImagesOrVideosTest.php
```diff
@@ -0,0 +1,458 @@
...
+            'only_annotation_labels' => [$l2->label_id, $l3->label_id]]);
+
+        with(new CloneImagesOrVideos($request))->handle();
+
+        $copy = $project->volumes()->first();
+        $newVideo = $copy->videos()->first();
+        $newAnnotation = $newVideo->annotations()->first();
+        $newAnnotationLabels = $newAnnotation->labels()->get();
+
+        $this->assertEquals(2, $newAnnotationLabels->count());
```
mzur, line 423: Please also check the `label_id` of the annotation labels (similar to image/video labels above).

tests/php/Jobs/CloneImagesOrVideosTest.php
```diff
@@ -0,0 +1,458 @@
...
+        Storage::fake('ifdos');
+        $csv = __DIR__."/../../files/image-ifdo.yaml";
+        $file = new UploadedFile($csv, 'ifdo.yaml', 'application/yaml', null, true);
+        $volume->saveIfdo($file);
+
+        // The target project.
+        $project = ProjectTest::create();
+
+        $this->beAdmin();
+        $project->addUserId($this->admin()->id, Role::adminId());
```
mzur, line 444:
```suggestion
```

mzur (changes requested):
Most comments are related to the new job class. I think this can be merged after one final iteration. Then we can focus on the UI.

app/Jobs/CloneImagesOrVideos.php
lehecht: If I delete addVolmeId(), my tests fail with "Error: Call to a member function videos()/images() on null".
But the flushThumbnailCache method was unnecessary, so I deleted it.

lehecht: If I delete addVolumeId(), my tests fail with "Error: Call to a member function videos() on null".
But the last line is unecessary. I will remove it.

mzur: Yes, I was just referring to `flushThumbnailCache`. Thanks!

app/Jobs/CloneImagesOrVideos.php
```diff
@@ -0,0 +1,480 @@
...
+                $this->copyVideos($volume, $copy, $onlyFiles);
+                if ($cloneAnnotations) {
+                    $this->copyVideoAnnotation($volume, $copy, $onlyFiles, $onlyAnnotationLabels);
+                }
+                if ($cloneFileLabels) {
+                    $this->copyVideoLabels($volume, $copy, $onlyFiles, $onlyFileLabels);
+                }
+            }
+            if ($copy->files()->exists()) {
+                ProcessNewVolumeFiles::dispatch($copy);
```
mzur, line 139: I just remembered another thing that is missing here.

While the annotations are cloned correctly (if there are any), the biigle/largo module also needs to generate thumbnail images for the annotations. This can be done in a similar way than the postprocessing of a [volume import through the admin panel](https://github.com/biigle/sync/blob/7430b7eb898768a58842a44b173d139594d10809/src/Jobs/PostprocessVolumeImport.php#L50-L68). It checks if the job classes of the biigle/largo module are there (i.e. the module is installed) and then pushes one job for each newly imported annotation.

Similarly, [the tests](https://github.com/biigle/sync/blob/7430b7eb898768a58842a44b173d139594d10809/tests/Jobs/PostprocessVolumeImportTest.php#L26-L52) require the biigle/largo module to be installed. They use `markTestSkipped()` if biigle/largo is not there.

app/Jobs/CloneImagesOrVideos.php
```diff
@@ -136,7 +143,7 @@ public function handle()
                 }
             }
             if ($copy->files()->exists()) {
-                ProcessNewVolumeFiles::dispatch($copy);
+                (new PostProcessingVolumeCloning($copy))->handle();
```
mzur, line 146: Why create a separate job class and then call it (synchronously) here? I suggest that you just add a new method to this job (e.g. `$this->postProcessCloning($copy)`) which includes all the code of `PostProcessingVolumeCloning::handle`.

app/Jobs/CloneImagesOrVideos.php
```diff
@@ -2,6 +2,7 @@
 
 namespace Biigle\Jobs;
 
+use AdvancedJsonRpc\Request;
```
mzur, line 5:
```suggestion
use Illuminate\Http\Request;
```

routes/web.php
```diff
@@ -254,6 +254,10 @@
             'as'   => 'volume',
             'uses' => 'VolumeController@index',
         ]);
+        $router->get('clone/{id}',[
+            'as' => 'clone-volume',
+            'uses' => 'VolumeController@clone'
+        ]);
```
mzur, line 260: This should be added in the next PR for the UI. Please remove it here.

tests/php/Http/Controllers/Api/VolumeControllerTest.php
```diff
@@ -220,34 +221,44 @@ public function testCloneVolume()
...
-            'updated_at' => '2022-11-09 14:37:00',])->fresh();
+                'updated_at' => '2022-11-09 14:37:00',])->fresh();
         $project = ProjectTest::create();
 
         $this->doTestApiRoute('POST', "/api/v1/volumes/{$volume->id}/clone-to/{$project->id}");
 
         $this->be($project->creator);
-        $this->postJson("/api/v1/volumes/{$volume->id}/clone-to/{$project->id}")
+        $this->postJson("/api/v1/volumes/{$volume->id}/clone-to/{$project->id}",
+            ['volume' => $volume, 'project' => $project])
```
mzur, line 231: The additional parameters shouldn't be required. The volume and project are fetched by their ID in the `CloneVolume` request.
```suggestion
        $this->postJson("/api/v1/volumes/{$volume->id}/clone-to/{$project->id}")
```

tests/php/Http/Controllers/Api/VolumeControllerTest.php
```diff
@@ -220,34 +221,44 @@ public function testCloneVolume()
...
             // No update permissions in the source project.
             ->assertStatus(403);
 
         $this->beAdmin();
-        $this->postJson("/api/v1/volumes/{$volume->id}/clone-to/{$project->id}")
+        $this->postJSon("/api/v1/volumes/{$volume->id}/clone-to/{$project->id}",
+            ['volume' => $volume, 'project' => $project])
```
mzur, line 237:
```suggestion
        $this->postJon("/api/v1/volumes/{$volume->id}/clone-to/{$project->id}")
```

tests/php/Http/Controllers/Api/VolumeControllerTest.php
```diff
@@ -220,34 +221,44 @@ public function testCloneVolume()
...
 
         $project->addUserId($this->admin()->id, Role::adminId());
+
         Cache::flush();
-        $this->postJson("/api/v1/volumes/{$volume->id}/clone-to/{$project->id}")
+
+        Queue::fake();
+
+        $this->postJson("/api/v1/volumes/{$volume->id}/clone-to/{$project->id}",
+            ['volume' => $volume, 'project' => $project])
```
mzur, line 248:
```suggestion
        $this->postJson("/api/v1/volumes/{$volume->id}/clone-to/{$project->id}")
```

tests/php/Http/Controllers/Api/VolumeControllerTest.php
```diff
@@ -220,34 +221,44 @@ public function testCloneVolume()
...
 
         // The target project.
         $project = ProjectTest::create();
 
         $this->beAdmin();
         $project->addUserId($this->admin()->id, Role::adminId());
 
-        $response = $this->postJson("/api/v1/volumes/{$volume->id}/clone-to/{$project->id}");
+        $response = $this->postJson("/api/v1/volumes/{$volume->id}/clone-to/{$project->id}",
+            ['volume' => $volume, 'project' => $project]);
```
mzur, line 259:
```suggestion
        $response = $this->postJson("/api/v1/volumes/{$volume->id}/clone-to/{$project->id}");
```

tests/php/Jobs/CloneImagesOrVideosTest.php
```diff
@@ -62,19 +62,16 @@ public function testCloneVolume()
...
+    }
 
-        // test for video
-
+    public function testCloneVideoVolume(){
         $volume = VolumeTest::create(
-            ['created_at' => '2023-01-09 14:37:00',
-                'updated_at' => '2023-01-09 14:37:00',
+            ['created_at' => '2022-01-09 14:37:00',
+                'updated_at' => '2022-01-09 14:37:00',
```
mzur, line 70: Just wondering: Is there any special reason for this change?

tests/php/Jobs/CloneImagesOrVideosTest.php
```diff
@@ -105,11 +103,8 @@ public function testCloneVolumeImages()
         // The target project.
         $project = ProjectTest::create();
 
-        $this->beAdmin();
-        $project->addUserId($this->admin()->id, Role::adminId());
-
         $oldImage = ImageTest::create([
-            'filename' => 'a.jpg',
+            'filename' => 'j.jpg',
```
mzur, line 107: Here, too: Why the change?

mzur (changes requested):
Almost finished :+1:

tests/php/Jobs/CloneImagesOrVideosTest.php
lehecht: No, I forgot to undo the changes. I encountered a strange error and changed the data just to make sure that it wasn't caused by it.

mzur: It's not important in that case :wink:

mzur:
The maintainer of the PHP linter Psalm was [very quick](https://github.com/vimeo/psalm/pull/9329) with a fix for the issue that you discovered. I've updated the dependency in your branch. There is no need for the `@psalm-suppress` commenty any more.

routes/web.php
mzur, line 260: The route is still there. Please remove it for this PR.

mzur (changes requested):
Only one thing slipped through. Then everything is ready to be merged.

mzur (approved):
:+1:

PR #560, opened 2023-03-11
Clone volume UI

lehecht:


resources/assets/js/volumes/cloneForm.vue
```diff
@@ -0,0 +1,190 @@
...
+                    label.selected = true : label.selected = false);
+            }
+
+            this.fileLabelTrees = fileLabelTrees;
+            this.annotationLabelTrees = annotationLabelTrees;
+        },
+        initializeFileList(ids, filenames) {
+            const nbrFiles = ids.length;
+            if (nbrFiles > 0) {
+                this.cloneFiles = true;
```
mzur, line 137: The variable name is misleading. It should rather be `filterFiles`, correct?

mzur, line 137: This is also the same with the file labels and annotation labels above. If I select "filter images/videos", then add a filter query and get a file list and then uncheck "filter images/videos" again, the file list is still there as hidden form fields.

resources/views/volumes/clone.blade.php
```diff
@@ -0,0 +1,195 @@
...
+        biigle.$declare('volume', {!!$volume!!});
+        biigle.$declare('name', '{!!old('name',$volume->name)!!}');
+        biigle.$declare('fileLabelTrees', '{!!$labelTrees!!}');
+        biigle.$declare('selectedFilesIds', '{!! collect(old('only_files',[])) !!}');
+        biigle.$declare('selectedFiles', '{!! collect(old('selected_files',[])) !!}');
+        biigle.$declare('fileLabelIds', '{!! collect(old('only_file_labels',[])) !!}');
+        biigle.$declare('annotationLabelTrees', '{!!$labelTrees!!}');
+        biigle.$declare('cloneFileLabels', {{old('clone_file_labels',false)}});
+        biigle.$declare('cloneAnnotations', {{old('clone_annotations',false)}});
+        biigle.$declare('annotationLabelIds', '{!! collect(old('only_annotation_labels', [])) !!}');
```
mzur, line 17: In our last meeting we agreed that adding the label trees as JSON strings and using `JSON.parse()` would be useful because this way we get two independent label trees objects and the label `selected` state would not be shared across them. But this is wrong (I see now), you don't need strings and `JSON.parse()` at all for this. It's enough to define both objects as separate varaibles here. Also, you don't need strings and `JSON.parse()` for any of the other variables:

```suggestion
        biigle.$declare('fileLabelTrees', {!!$labelTrees!!});
        biigle.$declare('selectedFilesIds', {!! collect(old('only_files',[])) !!});
        biigle.$declare('selectedFiles', {!! collect(old('selected_files',[])) !!});
        biigle.$declare('fileLabelIds', {!! collect(old('only_file_labels',[])) !!});
        biigle.$declare('annotationLabelTrees', {!!$labelTrees!!});
        biigle.$declare('cloneFileLabels', {{old('clone_file_labels',false)}});
        biigle.$declare('cloneAnnotations', {{old('clone_annotations',false)}});
        biigle.$declare('annotationLabelIds', {!! collect(old('only_annotation_labels', [])) !!});
```

resources/assets/js/volumes/cloneForm.vue
```diff
@@ -0,0 +1,190 @@
...
+        this.cloneUrlTemplate = biigle.$require('cloneUrlTemplate');
+        this.selectedProjectId = Number(location.href.split('=')[1]);
+        this.annotationLabelIds = JSON.parse(biigle.$require('annotationLabelIds'));
+        this.fileLabelIds = JSON.parse(biigle.$require('fileLabelIds'));
+        this.cloneFileLabels = biigle.$require('cloneFileLabels');
+        this.cloneAnnotationLabels = biigle.$require('cloneAnnotations');
+        let fileLabelTrees = JSON.parse(biigle.$require('fileLabelTrees'));
+        let annotationLabelTrees = JSON.parse(biigle.$require('annotationLabelTrees'));
+        let ids = JSON.parse(biigle.$require('selectedFilesIds'));
+        let filenames = JSON.parse(biigle.$require('selectedFiles'));
```
mzur, line 177: See my other comment.

```suggestion
        this.annotationLabelIds = biigle.$require('annotationLabelIds');
        this.fileLabelIds = biigle.$require('fileLabelIds');
        this.cloneFileLabels = biigle.$require('cloneFileLabels');
        this.cloneAnnotationLabels = biigle.$require('cloneAnnotations');
        let fileLabelTrees = biigle.$require('fileLabelTrees');
        let annotationLabelTrees = biigle.$require('annotationLabelTrees');
        let ids = biigle.$require('selectedFilesIds');
        let filenames = biigle.$require('selectedFiles');
```

resources/views/volumes/clone.blade.php
```diff
@@ -0,0 +1,195 @@
...
+                        <label>New volume destination project</label>
+                        <typeahead class="typeahead--block" :items="getProjects"
+                                   placeholder="Select destination project"
+                                   title="Select project to clone volume to"
+                                   v-on:select="setProject" :clear-on-select="false"
+                                   :value="setDefaultProject" required></typeahead>
+                    </div>
+
+                    <div class="checkbox">
+                        <label><input type="checkbox" id="files" v-model="cloneFiles">
```
mzur, line 54: You should add help text blocks for all of the checkboxes. See the reports form for an [example](https://github.com/biigle/reports/blob/5f842b59f29974e010bc03e14c0d84a2eb8aaed4/src/resources/views/volumeReports.blade.php#L105-L107).

When "filter images" is checked, you should also display a text that explains how the "*" wildcard can be used ([example text](https://github.com/biigle/core/blob/3ec7819bbadc889895c2eb5561b05520bafd6c8b/resources/assets/js/volumes/stores/filters.js#L177)).

resources/views/volumes/clone.blade.php
```diff
@@ -0,0 +1,195 @@
...
+                                               placeholder="video*.mp4" v-model="filePattern" required
+                                               v-on:keydown.enter="loadFilesMatchingPattern">
+                                    @endif
+                                </div>
+                            </div>
+                            <div class="panel-body">
+                                <ul class="list-group files-list" v-cloak>
+                                    <li v-for="file in selectedFiles" class="list-group-item"><span
+                                            class="text-muted">#<span v-text="file.id"></span></span> <span
+                                            v-text="file.filename"></span></li>
```
mzur, line 90: You should show a text if no files matched a submitted pattern.

resources/views/volumes/clone.blade.php
```diff
@@ -0,0 +1,195 @@
...
+                            Filter <span>
+                            @if($volume->isImageVolume())
+                                    images
+                                @else
+                                    videos
+                                @endif
+                            </span>
+                        </label>
+                    </div>
+                    <div v-if="cloneFiles">
```
mzur, line 64: This element "flashes" when the page loads, i.e. it is briefly shown until Vue is initialized to hide it again. Use [`v-cloak`](https://vuejs.org/api/built-in-directives.html#v-cloak) to hide the element from the beginning:

```suggestion
                    <div v-if="cloneFiles" v-cloak>
```

resources/views/volumes/clone.blade.php
```diff
@@ -0,0 +1,195 @@
...
+                                </ul>
+                            </div>
+                            <div class="form-group{{ $errors->has('clone_files') ? ' has-error' : '' }}">
+                                @if($errors->has('clone_files'))
+                                    <span class="help-block">{{ $errors->first('clone_files') }}</span>
+                                @endif
+                            </div>
+                        </div>
+                    </div>
+                    <div v-cloak>
```
mzur, line 100: You should not use v-cloak here because the checkboxes should always be visible. Instead, use `v-cloak` on the child elements that use `v-if` and should be hidden initially.

resources/views/volumes/clone.blade.php
```diff
@@ -0,0 +1,195 @@
...
+                                    <span class="help-block">{{ $errors->first('only_annotation_labels') }}</span>
+                                @endif
+                            </div>
+                        </div>
+                    </div>
+                </div>
+                <div class="form-group">
+                    <input type="submit" class="btn btn-success pull-right" :disabled="cannotSubmit" value="Clone">
+                    <a class="btn btn-default pull-right" :disabled="loading" type="button"
+                       href="{{\Illuminate\Support\Facades\URL::previous()}}">Cancel</a>
```
mzur, line 179: There is an alias for this facade in Laravel, so you don't have to give the full namespace.

```suggestion
                       href="{{URL::previous()}}">Cancel</a>
```

resources/views/volumes/clone.blade.php
```diff
@@ -0,0 +1,195 @@
...
+                    <input v-for="id in annotationLabelIds" type="hidden" name="only_annotation_labels[]"
+                           v-bind:value="id">
+                    <input v-for="id in fileLabelIds" type="hidden" name="only_file_labels[]" v-bind:value="id">
+                    <input v-for="file in selectedFiles" type="hidden" name="only_files[]" v-bind:value="file.id">
+                    <input v-for="file in selectedFiles" type="hidden" name="selected_files[]"
+                           v-bind:value="file.filename">
```
mzur, line 185: This is no actual argument for the API, you only need it to fill the "filter files" list again, right? I rather wouldn't send this information back and forth, as it's only required in case of a validation error. The file IDs should be enough to initialize the file list again. In this case, you can perform the `VolumeApi.queryFilenames()` request again inside `initializeFileList()` and then re-populate the file list after the filenames were fetched that way.

resources/views/volumes/clone.blade.php
```diff
@@ -0,0 +1,195 @@
...
+                    <input type="hidden" name="clone_annotations" v-if="cloneAnnotationLabels" v-bind:value=1>
+                    <input type="hidden" name="clone_annotations" v-else v-bind:value=0>
```
mzur, line 187: There should be no need for this. You should be able to directly utilize the visible checkbox:

```html
<input type="checkbox" id="annotations" name="clone_annotations" v-model="cloneAnnotationLabels">
```

If this doesn't work, we should discuss it in our meeting.

resources/views/volumes/clone.blade.php
```diff
@@ -0,0 +1,195 @@
...
+                    <input type="hidden" name="clone_file_labels" v-if="cloneFileLabels" v-bind:value=1>
+                    <input type="hidden" name="clone_file_labels" v-else v-bind:value=0>
```
mzur, line 189: Same than above.

resources/views/volumes/clone.blade.php
mzur, line 179: There should be a little space between the buttons. You could either only `pull-right` the submit button and leave cancel on the left or do:

```suggestion
                    <span class="pull-right">
                        <a class="btn btn-default" :disabled="loading" type="button"
                           href="{{\Illuminate\Support\Facades\URL::previous()}}">Cancel</a>
                        <input type="submit" class="btn btn-success" :disabled="cannotSubmit" value="Clone">
                    </span>
```

resources/views/volumes/clone.blade.php
```diff
@@ -0,0 +1,195 @@
...
+
+@section('content')
+    <div class="container">
+        <div class="col-sm-8 col-sm-offset-2 col-lg-6 col-lg-offset-3">
+            <div class="form-group">
+                <h2 class="row">Clone volume "{{$volume->name}}"</h2>
+                <br>
+            </div>
+            <form id="clone-volume-form" class="clearfix" role="form" method="POST" :action="getCloneUrl"
+                  enctype="multipart/form-data" v-on:submit="startLoading" v-on:keydown.enter.prevent>
```
mzur, line 30: Why this?

```suggestion
                  enctype="multipart/form-data" v-on:submit="startLoading">
```

mzur (changes requested):
We're getting there :+1:

resources/assets/js/volumes/cloneForm.vue
```diff
@@ -0,0 +1,190 @@
...
+                    }, handleErrorResponse);
+                VolumeApi.queryFilesWithFilename({id: this.id, pattern: this.filePattern})
+                    .then((response2) => {
+                        let ids = response2.body;
+                        this.setMatchedFiles(ids.map(id => {
+                            return {id: id, filename: id2filenames[id]}
+                        }));
+                    }, handleErrorResponse)
+                    .finally(this.finishLoading);
+            }
```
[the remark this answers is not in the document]
lehecht: I don't know why, but you solution always results in error "Server didn't respond, sorry".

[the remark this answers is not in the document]
mzur: You can try to remove the `handleErrorResponse` error handlers and then take a look what error pops up in the browser console.

resources/views/volumes/clone.blade.php
lehecht: Because otherwise, if you want to search for specific files and hit enter then the form is submitted.

mzur: I see. This must be fixed in the typeahead component. If you can, delete this attribute of the `<form>` in your next update of the PR and I will take care of the typeahead component later.

resources/assets/js/volumes/cloneForm.vue
```diff
@@ -0,0 +1,190 @@
...
+            cloneUrlTemplate: "",
+        };
+    },
+    computed: {
+        getProjects() {
+            return this.destinationProjects;
+        },
+        selectedFileLabels() {
+            let selection = this.flatLabels(this.fileLabelTrees).filter((label) => label.selected);
+            this.setFileLabels(selection.map((label) => label.id));
```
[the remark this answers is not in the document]
lehecht: If I use the first approach, my old values are empty.

mzur (changes requested):
Another round but then I think we're done :+1:

tests/php/Http/Controllers/Views/Volumes/VolumeCloneControllerTest.php
```diff
@@ -2,50 +2,47 @@
...
+        // if project creator and volume creator are the same, the volume can be cloned
+        $this->be($this->project()->creator);
+        $response = $this->get("volumes/clone/{$id}");
+        $response->assertStatus(200);
 
         $this->beAdmin();
         $response = $this->get("volumes/clone/{$id}");
         $response->assertStatus(200);
+
+        $response = $this->get('volumes/clone/999');
```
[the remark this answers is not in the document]
lehecht: I just copied it from the testEdit method in the views/volumeController. Should it be changed there too?

resources/views/volumes/clone.blade.php
```diff
@@ -47,116 +47,122 @@
...
                     </div>
 
-                    <div class="checkbox">
-                        <label><input type="checkbox" id="files" v-model="cloneFiles">
-                            Filter <span>
-                            @if($volume->isImageVolume())
-                                    images
+                    <div class="checkbox" v-cloak>
+                        <label><input type="checkbox" id="files" v-model="filterFiles">
+                            <span v-text="'Filter '+getFileType(false)+'s'"></span>
```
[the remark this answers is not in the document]
lehecht: Should I change it only at this point or everywhere I switch between video and image?

resources/views/volumes/clone.blade.php
```diff
@@ -47,116 +47,122 @@
...
                                 <label><input type="checkbox" class="checkbox" id="restrictFileLabels"
                                               v-model="restrictFileLabels">
-                                    @if($volume->isImageVolume())
-                                        Restrict image labels (<span v-text="selectedFileLabelsCount"></span> labels
-                                        selected)
-                                    @else
-                                        Restrict video labels (<span v-text="selectedFileLabelsCount"></span> labels
-                                        selected)
-                                    @endif
+                                    <span v-text="'Restrict '+getFileType(false)+' labels '+'('+selectedFileLabelsCount+' labels selected)'"></span>
```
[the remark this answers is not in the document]
lehecht: I will change it to "filter file labels" to keep it consistent.

tests/php/Http/Controllers/Views/Volumes/VolumeCloneControllerTest.php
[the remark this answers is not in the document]
mzur: That would be nice, yes. I used to do it like this at first until the test suite got larger. Now these issues mostly pup up as errors in the GitHub actions that run the tests in random order. Sometimes such a test is executed at the end of a test run and suddenly fails.

resources/assets/js/volumes/cloneForm.vue
```diff
@@ -70,11 +70,11 @@ export default {
         getCloneUrl() {
             return this.cloneUrlTemplate.replace(':pid', this.selectedProjectId);
         },
-    },
-    methods: {
-        setDefaultProject() {
+        defaultProjectName() {
```
mzur, line 73: This is still more like the `selectedProjectName` rather than the `defaultProjectName` if I'm not mistaken.

resources/views/volumes/clone.blade.php
```diff
@@ -93,83 +95,76 @@ class="text-muted">#<span v-text="file.id"></span></span> <span
...
                             </div>
 
-                            <div v-if="cloneFileLabels" v-cloak>
+                            <div v-if="filterFileLabels" v-cloak>
                                 <label><input type="checkbox" class="checkbox" id="restrictFileLabels"
                                               v-model="restrictFileLabels">
-                                    <span v-text="'Restrict '+getFileType(false)+' labels '+'('+selectedFileLabelsCount+' labels selected)'"></span>
-                                    <div class="form-group{{ $errors->has('only_file_labels') ? ' has-error' : '' }}">
+                                    <span
+                                        v-text="'Restrict file labels '+'('+selectedFileLabelsCount+' labels selected)'"></span>
```
mzur, line 114: This should become "filter file labels", right? Also, an alternative implementation could be:

```html
Filter file labels (<span v-text="selectedFileLabelsCount"></span> labels selected)
```

resources/views/volumes/clone.blade.php
```diff
@@ -47,18 +47,17 @@
...
+                                   :value="defaultProjectName" required></typeahead>
                     </div>
 
-                    <div class="checkbox" v-cloak>
+                    <div class="checkbox">
                         <label><input type="checkbox" id="files" v-model="filterFiles">
-                            <span v-text="'Filter '+getFileType(false)+'s'"></span>
-                            <div class="form-group{{ $errors->has('clone_files') ? ' has-error' : '' }}">
+                            <span>Filter files</span>
+                            <div class="form-group{{ $errors->has('clone_files') ? ' has-error' : '' }}" v-cloak>
```
mzur, line 56: Only use `v-cloak` on elements with `v-if` :wink: So this should be added to the `<span v-if="filterFiles">` help block below. Otherwise, the default help text will be flashing on page load.

Also in Bootstrap 3, a checkbox does not need a `form-group`. Instead, it is wrapped in a `has-error` (see [here](https://getbootstrap.com/docs/3.3/css/#forms-control-validation)).

resources/views/volumes/clone.blade.php
```diff
@@ -93,83 +95,76 @@ class="text-muted">#<span v-text="file.id"></span></span> <span
...
                         <div class="checkbox">
-                            <div v-cloak>
+                            <div>
                                 <label><input type="checkbox" class="checkbox" id="fileLabels"
-                                              v-model="cloneFileLabels" name="clone_file_labels" value="1">
-                                    <span v-text="'Clone '+getFileType(false)+' labels'"></span>
-                                    <div class="form-group{{ $errors->has('clone_file_labels') ? ' has-error' : '' }}">
+                                              v-model="filterFileLabels" name="clone_file_labels" value="1">
+                                    <span>Clone file labels</span>
+                                    <div class="form-group{{ $errors->has('clone_file_labels') ? ' has-error' : '' }}" v-cloak>
```
mzur, line 102: No `form-group` required for a checkbox. The `has-error` could be added to the the `<div>` at line 96.

Also there is no `v-if` here, so no `v-cloak` is required.

Also any empty `<div>` in the template (e.g. line 98) is unnecessary.

mzur (approved):
While I've added another round of comments, I approve this PR. I'll add a final commit to this PR to show what I meant with my comments. Then I'll merge. Thanks a lot :+1:

PR #593, opened 2023-06-07
Fix multiple projects dropdown in image annotation tool

lehecht:
Css class 'annotations-breadcrumb' prevents displaying project dropdown because of 'overflow: hidden;' property.

Move `@include(...)` line for dropdown outside the css class scope and wrap it in own div element with own css class.

### DIFF
--- a/resources/assets/sass/annotations/main.scss
+++ b/resources/assets/sass/annotations/main.scss
@@ -22,7 +22,7 @@
 }
 
 // Assign dropdown an own css class due to bug #583
-.annotations-project-dd{
+.annotations-project-dd {
     display: inline-block;
     vertical-align: bottom;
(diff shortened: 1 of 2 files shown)


resources/assets/sass/annotations/main.scss
```diff
@@ -21,6 +21,13 @@
     vertical-align: bottom;
 }
 
+// Assign dropdown an own css class due to bug #583
+.annotations-project-dd{
```
mzur, line 25:
```suggestion
.annotations-project-dd {
```

mzur (dismissed):
Thanks! I suggested a small change for the coding style which I will apply myself. I'll also update the title of the PR to be more descriptive because it will be used to generate the changelog later.

Issue #441, opened 2022-03-30
Handle WoRMS errors more gracefully


mzur:
WoRMS is in maintenance mode at the moment. This produces a `SoapFault` error [here](https://github.com/biigle/core/blob/557bfb2fdf1d083cf2fe44e87796d25e8428b598/app/Services/LabelSourceAdapters/WormsAdapter.php#L91), which is not caught. Handle these kinds of errors more gracefully and return a "WoRMS is currently unavailable" response (maybe even code 503?).

mzur:
Clarification: WoRMS can be used to import labels from the external WoRMS database into a label tree:

![Screen Shot 2023-06-02 at 09 04 57](https://github.com/biigle/core/assets/2457311/0b9b80da-98e5-40a2-bd8e-c151debfc58b)

mzur:
@lehecht I have to reopen this issue because the error also appeared at this location:

https://github.com/biigle/core/blob/b370415195a05e845fe59daa10ccb1be4267dc2e/app/Http/Controllers/Api/LabelTreeLabelController.php#L50

It also occurred to me that the `SoapFault` is specific to the WoRMS adapter because other adapters might not use SOAP at all. So it's probably better to handle the exception directly in the adapter.

Here:

https://github.com/biigle/core/blob/b370415195a05e845fe59daa10ccb1be4267dc2e/app/Services/LabelSourceAdapters/WormsAdapter.php#L142-L158

By wrapping the whole block in a `try/catch` and throwing a `Symfony\Component\HttpKernel\Exception\ServiceUnavailableHttpException` with an appropriate message.

And here:

https://github.com/biigle/core/blob/b370415195a05e845fe59daa10ccb1be4267dc2e/app/Services/LabelSourceAdapters/WormsAdapter.php#L91

By also throwing a `ServiceUnavailableHttpException` with the message.

You should revert #595 and add new test cases to `WormsAdapterTest`.

PR #595, opened 2023-06-09
Fix bug #441 by catching WoRMS SoapFault exception

lehecht:


lehecht:
> Thanks! I have some small suggestions.
> 
> Also, can you please add a [test case](https://github.com/biigle/core/blob/master/tests/php/Http/Controllers/Api/LabelSourceControllerTest.php) for this? You can use similar code than in `testFind()` and test the exception with:
> 
> ```
> $mock = Mockery::mock();
> $mock->shouldReceive('find')
>     ->once()
>     ->andThrow(\SoapFault::class);
> ```

Should I create a new test method or just add it to the existing one?

mzur:
> Should I create a new test method or just add it to the existing one?

Please add a new method.

mzur (approved):
Thanks!

PR #597, opened 2023-06-13
Fix the server error caused by attaching a label to an existing video annotation

### LINKED ISSUE #410: Server error on double click to attach video annotation label
It seems to be possible to trigger a server error (i.e. a database unique constraint violation) with a double click when attaching a label to an existing video annotation. Catch these errors and return a correct 400 response. This issue was handled elaborately for image annotations with a [double check](https://github.com/biigle/core/blob/06ffbbda2ca1b4e47cb38b1520d40553286dc94e/app/Http/Controllers/Api/ImageAnnotationLabelController.php#L173-L193). Also implement such a check for video annotations. Ideally, this could share the same logic.

lehecht:
Attaching a label to an existing video annotation can cause a sever error when using the double click. Use the same strategy as in ImageAnnotationLabelController to fix that bug.

Resolves #410

### DIFF
--- a/app/Http/Controllers/Api/VideoAnnotationLabelController.php
+++ b/app/Http/Controllers/Api/VideoAnnotationLabelController.php
@@ -6,6 +6,8 @@ use Biigle\Http\Controllers\Api\Controller;
 use Biigle\Http\Requests\DestroyVideoAnnotationLabel;
 use Biigle\Http\Requests\StoreVideoAnnotationLabel;
 use Biigle\VideoAnnotationLabel;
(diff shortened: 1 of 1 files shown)


mzur (approved):
Thanks!

mzur:
@lehecht you can also add "Resolves #410" to the commit message. This way you could write a more descriptive commit message title that says what was done without having to look into the GitHub issue. Also, GitHub recognizes the keyword and will automatically link the issue with the pull request (as it has done after my edit to your PR description). When the PR is merged, the issue will automatically be closed.

PR #611, opened 2023-06-20
Resolve #558 by updating the view

### LINKED ISSUE #558: Fix handling of images that fail to load
If individual images of a volume cannot be displayed, the current behavior of the UI could lead to incorrect annotations. As it is, the annotation tool advances to the next image, displays the error that the image failed to load *but keeps displaying the previous image*. This could make users assume that the image succeeded to load. If they now create new annotations, the annotations will be created *on the image that failed to load*. This should not be possible.

To fix this:
- clear the main view of the annotation tool if an image could not be loaded (it should look like if the annotation tool was directly opened for that image)
- disallow the creation of new annotations

Also check if there is a similar issue in the video annotation tool and fix it there.

lehecht:
Use Vue's conditional rendering for displaying the annotatorContainer.

Resolves #558

### DIFF
--- a/resources/assets/js/annotations/annotatorContainer.vue
+++ b/resources/assets/js/annotations/annotatorContainer.vue
@@ -86,6 +86,8 @@ export default {
             userUpdatedVolareResolution: false,
             userId: null,
             crossOriginError: false,
+            // Toggles the annotation container view for images that could not be loaded,
+            // see https://github.com/biigle/core/issues/558.
             showImage: true
         };
     },
--- a/resources/assets/js/annotations/annotatorContainer.vue
+++ b/resources/assets/js/annotations/annotatorContainer.vue
@@ -635,7 +635,7 @@ export default {
             this.annotationFilters[1].annotations = annotations;
         },
         image(image) {
-            this.crossOriginError = image.crossOrigin;
+            this.crossOriginError = image?.crossOrigin;
         },
     },
     created() {
--- a/resources/assets/js/annotations/annotatorContainer.vue
+++ b/resources/assets/js/annotations/annotatorContainer.vue
@@ -92,6 +92,15 @@ export default {
         };
     },
     computed: {
+        canAdd(){
+            return this.isEditor && !this.imageLoadingError;
+        },
+        canModify(){
+            return this.canAdd;
+        },
+        canDelete(){
+            return this.canAdd;
+        },
         imageId() {
             return this.imagesIds[this.imageIndex];
         },
--- a/resources/assets/js/annotations/annotatorContainer.vue
+++ b/resources/assets/js/annotations/annotatorContainer.vue
@@ -86,6 +86,7 @@ export default {
             userUpdatedVolareResolution: false,
             userId: null,
             crossOriginError: false,
+            showImage: true
         };
     },
     computed: {
@@ -545,6 +546,7 @@ export default {
             if (message instanceof CrossOriginError) {
                 this.crossOriginError = true;
             } else {
+                this.showImage = false;
                 Messages.danger(message);
             }
         },
@@ -577,6 +579,7 @@ export default {
     watch: {
         imageId(id) {
             if (id) {
+                this.showImage = true;
                 this.startLoading();
                 this.crossOriginError = false;
                 Vue.Promise.all(this.getImageAndAnnotationsPromises(id))
--- a/resources/assets/js/annotations/annotatorContainer.vue
+++ b/resources/assets/js/annotations/annotatorContainer.vue
@@ -166,6 +166,7 @@ export default {
             if (args) {
                 this.image = args[0];
                 this.annotations = args[1];
+                this.imageLoadingError = false;
             } else {
                 // This might happen if there was an error loading the image or the
                 // annotations.
@@ -581,7 +582,6 @@ export default {
     watch: {
         imageId(id) {
             if (id) {
-                this.imageLoadingError = false;
                 this.startLoading();
                 this.crossOriginError = false;
                 Vue.Promise.all(this.getImageAndAnnotationsPromises(id))
--- a/resources/assets/js/annotations/annotatorContainer.vue
+++ b/resources/assets/js/annotations/annotatorContainer.vue
@@ -572,7 +572,7 @@ export default {
             }
 
             this.startLoading();
-            this.crossOriginError = this.image?.crossOrigin;
+            this.crossOriginError = false;
 
             try {
                 let [image, annotations] = await Vue.Promise.all(this.getImageAndAnnotationsPromises(id));
--- a/resources/assets/js/annotations/annotatorContainer.vue
+++ b/resources/assets/js/annotations/annotatorContainer.vue
@@ -554,15 +554,6 @@ export default {
         handleClosedTab() {
             Settings.delete('openTab');
         },
-        handleLoadingError(message) {
-            if (message instanceof CrossOriginError) {
-                this.crossOriginError = true;
-            } else {
-                this.imageLoadingError = true;
-                this.annotations = [];
-                Messages.danger(message);
-            }
-        },
         createSampledAnnotation() {
             this.$refs.canvas.createSampledAnnotation();
         },
@@ -590,22 +581,37 @@ export default {
         },
     },
     watch: {
-        imageId(id) {
-            if (id) {
-                this.startLoading();
-                this.crossOriginError = false;
-                Vue.Promise.all(this.getImageAndAnnotationsPromises(id))
-                    .then(this.setCurrentImageAndAnnotations)
-                    .then(this.maybeUpdateAnnotationMode)
-                    .then(this.maybeShowTilingInProgressMessage)
-                    .catch(this.handleLoadingError)
-                    .finally(() => {
-                        this.updateUrlSlug();
-                        this.emitImageChanged();
-                        // When everything is loaded, pre-fetch the data of the next and
-                        // previous images so they can be switched fast.
-                        this.cachePreviousAndNext();
-                        this.finishLoading()});
+        async imageId(id) {
+            if (!id) {
+                return;
+            }
+
+            this.startLoading();
+            this.crossOriginError = image?.crossOrigin;
+
+            try {
+                let [image, annotations] = await Vue.Promise.all(this.getImageAndAnnotationsPromises(id));
+                this.imageLoadingError = false;
+                this.image = image;
+                this.annotations = annotations;
+                this.maybeUpdateAnnotationMode();
+                this.maybeShowTilingInProgressMessage();
+            } catch (e) {
+                if (e instanceof CrossOriginError) {
+                    this.crossOriginError = true;
+                } else {
+                    this.imageLoadingError = true;
+                    this.image = null;
+                    this.annotations = [];
+                    Messages.danger(e);
+                }
+            } finally {
+                this.updateUrlSlug();
+                this.emitImageChanged();
+                // When everything is loaded, pre-fetch the data of the next and
+                // previous images so they can be switched fast.
+                this.cachePreviousAndNext();
+                this.finishLoading();
             }
         },
         cachedImagesCount() {
--- a/resources/assets/js/annotations/annotatorContainer.vue
+++ b/resources/assets/js/annotations/annotatorContainer.vue
@@ -572,7 +572,7 @@ export default {
             }
 
             this.startLoading();
-            this.crossOriginError = image?.crossOrigin;
+            this.crossOriginError = this.image?.crossOrigin;
 
             try {
                 let [image, annotations] = await Vue.Promise.all(this.getImageAndAnnotationsPromises(id));
--- a/resources/assets/js/annotations/annotatorContainer.vue
+++ b/resources/assets/js/annotations/annotatorContainer.vue
@@ -156,6 +156,9 @@ export default {
         },
     },
     methods: {
+        getImageLoadingError(){
+            return this.imageLoadingError;
+        },
         getImageAndAnnotationsPromises(id) {
             return [
                 ImagesStore.fetchAndDrawImage(id),
@@ -587,15 +590,16 @@ export default {
                 this.crossOriginError = false;
                 Vue.Promise.all(this.getImageAndAnnotationsPromises(id))
                     .then(this.setCurrentImageAndAnnotations)
-                    .then(this.updateUrlSlug)
                     .then(this.maybeUpdateAnnotationMode)
-                    .then(this.emitImageChanged)
                     .then(this.maybeShowTilingInProgressMessage)
-                    // When everything is loaded, pre-fetch the data of the next and
-                    // previous images so they can be switched fast.
-                    .then(this.cachePreviousAndNext)
                     .catch(this.handleLoadingError)
-                    .finally(this.finishLoading);
+                    .finally(() => {
+                        this.updateUrlSlug();
+                        this.emitImageChanged();
+                        // When everything is loaded, pre-fetch the data of the next and
+                        // previous images so they can be switched fast.
+                        this.cachePreviousAndNext();
+                        this.finishLoading()});
             }
         },
         cachedImagesCount() {
--- a/resources/assets/js/annotations/annotatorContainer.vue
+++ b/resources/assets/js/annotations/annotatorContainer.vue
@@ -86,19 +86,16 @@ export default {
             userUpdatedVolareResolution: false,
             userId: null,
             crossOriginError: false,
-            // Toggles the annotation container view for images that could not be loaded,
-            // see https://github.com/biigle/core/issues/558.
-            imageLoadingError: false
         };
     },
     computed: {
-        canAdd(){
-            return this.isEditor && !this.imageLoadingError;
+        canAdd() {
+            return this.isEditor && (this.image !== null);
         },
-        canModify(){
+        canModify() {
             return this.canAdd;
         },
-        canDelete(){
+        canDelete() {
             return this.canAdd;
         },
         imageId() {
--- a/resources/assets/js/annotations/annotatorContainer.vue
+++ b/resources/assets/js/annotations/annotatorContainer.vue
@@ -168,18 +168,6 @@ export default {
                 AnnotationsStore.fetchAnnotations(id),
             ];
         },
-        setCurrentImageAndAnnotations(args) {
-            if (args) {
-                this.image = args[0];
-                this.annotations = args[1];
-                this.imageLoadingError = false;
-            } else {
-                // This might happen if there was an error loading the image or the
-                // annotations.
-                this.image = null;
-                this.annotations = [];
-            }
-        },
         updateUrlSlug() {
             UrlParams.setSlug(this.imageId, -2);
         },
--- a/resources/assets/js/annotations/annotatorContainer.vue
+++ b/resources/assets/js/annotations/annotatorContainer.vue
@@ -88,7 +88,7 @@ export default {
             crossOriginError: false,
             // Toggles the annotation container view for images that could not be loaded,
             // see https://github.com/biigle/core/issues/558.
-            showImage: true
+            imageLoadingError: false
         };
     },
     computed: {
@@ -548,7 +548,7 @@ export default {
             if (message instanceof CrossOriginError) {
                 this.crossOriginError = true;
             } else {
-                this.showImage = false;
+                this.imageLoadingError = true;
                 Messages.danger(message);
             }
         },
@@ -581,7 +581,7 @@ export default {
     watch: {
         imageId(id) {
             if (id) {
-                this.showImage = true;
+                this.imageLoadingError = false;
                 this.startLoading();
                 this.crossOriginError = false;
                 Vue.Promise.all(this.getImageAndAnnotationsPromises(id))
--- a/resources/assets/js/annotations/annotatorContainer.vue
+++ b/resources/assets/js/annotations/annotatorContainer.vue
@@ -165,9 +165,6 @@ export default {
         },
     },
     methods: {
-        getImageLoadingError(){
-            return this.imageLoadingError;
-        },
         getImageAndAnnotationsPromises(id) {
             return [
                 ImagesStore.fetchAndDrawImage(id),
--- a/resources/assets/js/annotations/annotatorContainer.vue
+++ b/resources/assets/js/annotations/annotatorContainer.vue
@@ -576,7 +576,6 @@ export default {
 
             try {
                 let [image, annotations] = await Vue.Promise.all(this.getImageAndAnnotationsPromises(id));
-                this.imageLoadingError = false;
                 this.image = image;
                 this.annotations = annotations;
                 this.maybeUpdateAnnotationMode();
@@ -585,7 +584,6 @@ export default {
                 if (e instanceof CrossOriginError) {
                     this.crossOriginError = true;
                 } else {
-                    this.imageLoadingError = true;
                     this.image = null;
                     this.annotations = [];
                     Messages.danger(e);
--- a/resources/assets/js/annotations/annotatorContainer.vue
+++ b/resources/assets/js/annotations/annotatorContainer.vue
@@ -550,6 +550,7 @@ export default {
                 this.crossOriginError = true;
             } else {
                 this.imageLoadingError = true;
+                this.annotations = [];
                 Messages.danger(message);
             }
         },
--- a/resources/assets/js/annotations/components/annotationCanvas.vue
+++ b/resources/assets/js/annotations/components/annotationCanvas.vue
@@ -723,7 +723,7 @@ export default {
             if (!state) {
                 this.resetInteractionMode();
             } else {
-                this.modifyInteraction.setActive(state && this.isDefaultInteractionMode);
+                this.modifyInteraction.setActive(this.isDefaultInteractionMode);
             }
         },
         canDelete(state) {
--- a/resources/assets/js/annotations/components/annotationCanvas.vue
+++ b/resources/assets/js/annotations/components/annotationCanvas.vue
@@ -304,18 +304,18 @@ export default {
                 multi: true
             });
 
-                // Map to detect which features were changed between modifystart and
-                // modifyend events of the modify interaction.
-                this.featureRevisionMap = {};
-                this.modifyInteraction = new ModifyInteraction({
-                    features: this.selectInteraction.getFeatures(),
-                    // The Shift key must be pressed to delete vertices, so that new
-                    // vertices can be drawn at the same position of existing
-                    // vertices.
-                    deleteCondition: function (event) {
-                        return shiftKeyOnlyCondition(event) && singleClickCondition(event);
-                    },
-                });
+            // Map to detect which features were changed between modifystart and
+            // modifyend events of the modify interaction.
+            this.featureRevisionMap = {};
+            this.modifyInteraction = new ModifyInteraction({
+                features: this.selectInteraction.getFeatures(),
+                // The Shift key must be pressed to delete vertices, so that new
+                // vertices can be drawn at the same position of existing
+                // vertices.
+                deleteCondition: function (event) {
+                    return shiftKeyOnlyCondition(event) && singleClickCondition(event);
+                },
+            });
         },
         updateMapSize() {
             this.mapSize = this.map.getSize();
@@ -715,14 +715,12 @@ export default {
         },
         canModify(state) {
             this.modifyInteraction.setActive(state && this.isDefaultInteractionMode);
-
         },
         canDelete(state) {
             if (state) {
                 Keyboard.on('Delete', this.deleteSelectedAnnotations, 0, this.listenerSet);
                 Keyboard.on('Backspace', this.deleteLastCreatedAnnotation, 0, this.listenerSet);
-            }
-            else {
+            } else {
                 Keyboard.off('Delete', this.deleteSelectedAnnotations, 0, this.listenerSet);
                 Keyboard.off('Backspace', this.deleteLastCreatedAnnotation, 0, this.listenerSet);
             }
--- a/resources/assets/js/annotations/components/annotationCanvas.vue
+++ b/resources/assets/js/annotations/components/annotationCanvas.vue
@@ -715,17 +715,16 @@ export default {
                 this.modifyInteraction.setActive(defaultMode);
             }
         },
-        canModify(state){
-            if(!state){
-                this.modifyInteraction.setActive(false);
-            }
+        canModify(state) {
+            this.modifyInteraction.setActive(state && this.isDefaultInteractionMode);
+
         },
-        canDelete(state){
-            if(state){
+        canDelete(state) {
+            if (state) {
                 Keyboard.on('Delete', this.deleteSelectedAnnotations, 0, this.listenerSet);
                 Keyboard.on('Backspace', this.deleteLastCreatedAnnotation, 0, this.listenerSet);
             }
-            else{
+            else {
                 Keyboard.off('Delete', this.deleteSelectedAnnotations, 0, this.listenerSet);
                 Keyboard.off('Backspace', this.deleteLastCreatedAnnotation, 0, this.listenerSet);
             }
--- a/resources/assets/js/annotations/components/annotationCanvas.vue
+++ b/resources/assets/js/annotations/components/annotationCanvas.vue
@@ -304,7 +304,7 @@ export default {
                 multi: true
             });
 
-            if (this.canModify) {
+            // if (this.canModify) {
                 // Map to detect which features were changed between modifystart and
                 // modifyend events of the modify interaction.
                 this.featureRevisionMap = {};
@@ -317,7 +317,7 @@ export default {
                         return shiftKeyOnlyCondition(event) && singleClickCondition(event);
                     },
                 });
-            }
+            // }
         },
         updateMapSize() {
             this.mapSize = this.map.getSize();
@@ -715,6 +715,24 @@ export default {
                 this.modifyInteraction.setActive(defaultMode);
             }
         },
+        canModify(state){
+            if(state){
+                this.modifyInteraction.setActive(true);
+            }
+            else{
+                this.modifyInteraction.setActive(false);
+            }
+        },
+        canDelete(state){
+            if(state){
+                Keyboard.on('Delete', this.deleteSelectedAnnotations, 0, this.listenerSet);
+                Keyboard.on('Backspace', this.deleteLastCreatedAnnotation, 0, this.listenerSet);
+            }
+            else{
+                Keyboard.off('Delete', this.deleteSelectedAnnotations, 0, this.listenerSet);
+                Keyboard.off('Backspace', this.deleteLastCreatedAnnotation, 0, this.listenerSet);
+            }
+        }
     },
     created() {
         this.declareNonReactiveProperties();
@@ -746,16 +764,10 @@ export default {
         Keyboard.on('ArrowLeft', this.handlePrevious, 0, this.listenerSet);
         Keyboard.on('Escape', this.resetInteractionMode, 0, this.listenerSet);
 
-        if (this.canModify) {
-            this.modifyInteraction.on('modifystart', this.handleFeatureModifyStart);
-            this.modifyInteraction.on('modifyend', this.handleFeatureModifyEnd);
-            this.map.addInteraction(this.modifyInteraction);
-        }
-
-        if (this.canDelete) {
-            Keyboard.on('Delete', this.deleteSelectedAnnotations, 0, this.listenerSet);
-            Keyboard.on('Backspace', this.deleteLastCreatedAnnotation, 0, this.listenerSet);
-        }
+        this.modifyInteraction.on('modifystart', this.handleFeatureModifyStart);
+        this.modifyInteraction.on('modifyend', this.handleFeatureModifyEnd);
+        this.map.addInteraction(this.modifyInteraction);
+        this.modifyInteraction.setActive(false);
     },
     mounted() {
         this.map.setTarget(this.$el);
--- a/resources/assets/js/annotations/components/annotationCanvas.vue
+++ b/resources/assets/js/annotations/components/annotationCanvas.vue
@@ -729,6 +729,8 @@ export default {
         canDelete(state) {
             if (!state) {
                 this.resetInteractionMode();
+                Keyboard.off('Delete', this.deleteSelectedAnnotations, 0, this.listenerSet);
+                Keyboard.off('Backspace', this.deleteLastCreatedAnnotation, 0, this.listenerSet);
             } else {
                 Keyboard.on('Delete', this.deleteSelectedAnnotations, 0, this.listenerSet);
                 Keyboard.on('Backspace', this.deleteLastCreatedAnnotation, 0, this.listenerSet);
--- a/resources/assets/js/annotations/components/annotationCanvas.vue
+++ b/resources/assets/js/annotations/components/annotationCanvas.vue
@@ -304,7 +304,6 @@ export default {
                 multi: true
             });
 
-            // if (this.canModify) {
                 // Map to detect which features were changed between modifystart and
                 // modifyend events of the modify interaction.
                 this.featureRevisionMap = {};
@@ -317,7 +316,6 @@ export default {
                         return shiftKeyOnlyCondition(event) && singleClickCondition(event);
                     },
                 });
-            // }
         },
         updateMapSize() {
             this.mapSize = this.map.getSize();
--- a/resources/assets/js/annotations/components/annotationCanvas.vue
+++ b/resources/assets/js/annotations/components/annotationCanvas.vue
@@ -638,6 +638,7 @@ export default {
             if (!image) {
                 this.map.removeLayer(this.tiledImageLayer);
                 this.map.removeLayer(this.imageLayer);
+                this.resetInteractionMode();
             } else if (image.tiled === true) {
                 if (!oldImage || oldImage.tiled !== true) {
                     this.map.removeLayer(this.imageLayer);
--- a/resources/assets/js/annotations/components/annotationCanvas.vue
+++ b/resources/assets/js/annotations/components/annotationCanvas.vue
@@ -716,10 +716,7 @@ export default {
             }
         },
         canModify(state){
-            if(state){
-                this.modifyInteraction.setActive(true);
-            }
-            else{
+            if(!state){
                 this.modifyInteraction.setActive(false);
             }
         },
--- a/resources/assets/js/annotations/components/annotationCanvas.vue
+++ b/resources/assets/js/annotations/components/annotationCanvas.vue
@@ -719,22 +719,28 @@ export default {
                 this.resetInteractionMode();
             }
         },
-        canModify(state) {
-            if (!state) {
-                this.resetInteractionMode();
-            } else {
-                this.modifyInteraction.setActive(this.isDefaultInteractionMode);
-            }
+        canModify: {
+            handler(state) {
+                if (!state) {
+                    this.resetInteractionMode();
+                } else {
+                    this.modifyInteraction.setActive(this.isDefaultInteractionMode);
+                }
+            },
+            immediate: true
         },
-        canDelete(state) {
-            if (!state) {
-                this.resetInteractionMode();
-                Keyboard.off('Delete', this.deleteSelectedAnnotations, 0, this.listenerSet);
-                Keyboard.off('Backspace', this.deleteLastCreatedAnnotation, 0, this.listenerSet);
-            } else {
-                Keyboard.on('Delete', this.deleteSelectedAnnotations, 0, this.listenerSet);
-                Keyboard.on('Backspace', this.deleteLastCreatedAnnotation, 0, this.listenerSet);
-            }
+        canDelete: {
+            handler(state) {
+                if (!state) {
+                    this.resetInteractionMode();
+                    Keyboard.off('Delete', this.deleteSelectedAnnotations, 0, this.listenerSet);
+                    Keyboard.off('Backspace', this.deleteLastCreatedAnnotation, 0, this.listenerSet);
+                } else {
+                    Keyboard.on('Delete', this.deleteSelectedAnnotations, 0, this.listenerSet);
+                    Keyboard.on('Backspace', this.deleteLastCreatedAnnotation, 0, this.listenerSet);
+                }
+            },
+            immediate: true
         }
     },
     created() {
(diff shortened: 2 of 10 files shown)


resources/assets/js/annotations/annotatorContainer.vue
```diff
@@ -86,6 +86,9 @@ export default {
             userUpdatedVolareResolution: false,
             userId: null,
             crossOriginError: false,
+            // Toggles the annotation container view for images that could not be loaded,
+            // see https://github.com/biigle/core/issues/558.
+            showImage: true
```
mzur, line 91: I think I'd like a variable named `imageLoadingError` better (similar to the existing `crossOriginError`, `false` by default). This describes better what the variable is for.

resources/views/annotations/show.blade.php
```diff
@@ -63,6 +63,7 @@
                 <button type="button" class="btn btn-default" v-on:click="dismissCrossOriginError">Continue at own risk</button>
             </div>
         </message-curtain>
+        <div v-show="showImage">
```
mzur, line 66: You can add `v-show="!imageLoadingError"` directly to the attributes of the `<annotation-canvas>`. Then you don't need an additional wrapper.

mzur (changes requested):

mzur:
Thanks! There are still a couple of things I'd like to improve for the error handling:

1. If the whole `<annotation-canvas>` is made invisible, the users also don't see the controls to move to the next/previous image any more. Instead, these controls should be visible and only the image should not be shown. It's similar to what you got before if you directly open a missing image. Maybe you can just pass `null` as the `image` attribute of the `<annotation-canvas>`?

2. If the control buttons at the bottom remain visible, the add/modify/delete buttons could be hidden. The `<annotation-canvas>` already [has attributes](https://github.com/biigle/core/blob/b4242255608bcddc2b6f9fc0598ea39515d631bd/resources/views/annotations/show.blade.php#L67-L69) to control this. However, the visibility of the buttons is currently checked only [in the backend](https://github.com/biigle/core/blob/b4242255608bcddc2b6f9fc0598ea39515d631bd/resources/views/annotations/show/annotationCanvas.blade.php#L27) when the page is loaded. To dynamically show/hide the buttons, you have to update the template with the approproate `v-if` attributes.

3. `imageLoadingError` is reset before the image loading is initiated. It can happen that there is a quick flash of the old image while the new image loads (and fails) until the `<annotation-canvas>` is hidden again. Instead, the variable could be reset in [`setCurrentImageAndAnnotations`](https://github.com/biigle/core/blob/b4242255608bcddc2b6f9fc0598ea39515d631bd/resources/assets/js/annotations/annotatorContainer.vue#L583), for example, or you add a new method for this.

4. This is a new issue: If the image fails to load, the URL is not updated because [`updateUrlSlug`](https://github.com/biigle/core/blob/b4242255608bcddc2b6f9fc0598ea39515d631bd/resources/assets/js/annotations/annotatorContainer.vue#L584) is not called. This should be done regardless if the loading is successful or not. I think it can even be just called in the watcher of `imageId` and doesn't have to be included in the promise chain at all.

lehecht:
> 2\. f the control buttons at the bottom remain visible, the add/modify/delete buttons could be hidden. The `<annotation-canvas>` already [has attributes](https://github.com/biigle/core/blob/b4242255608bcddc2b6f9fc0598ea39515d631bd/resources/views/annotations/show.blade.php#L67-L69) to control this. However, the visibility of the buttons is currently checked only [in the backend](https://github.com/biigle/core/blob/b4242255608bcddc2b6f9fc0598ea39515d631bd/resources/views/annotations/show/annotationCanvas.blade.php#L27) when the page is loaded. To dynamically show/hide the buttons, you have to update the template with the approproate `v-if` attributes.

So the lower button group should completely disappear in that case, right?

mzur:
No it should look like as if you are a project "guest", showing the next/previous buttons and the measure instrument:

image

mzur:
It's the same than what the [backend currently does](https://github.com/biigle/core/blob/b4242255608bcddc2b6f9fc0598ea39515d631bd/resources/views/annotations/show/annotationCanvas.blade.php#L27) woth `@can('add-annotation', $image)` but dynamically in the frontend. After the change we probably don't even need the `@can` in the backend any more.

lehecht:
> 4\. This is a new issue: If the image fails to load, the URL is not updated because [`updateUrlSlug`](https://github.com/biigle/core/blob/b4242255608bcddc2b6f9fc0598ea39515d631bd/resources/assets/js/annotations/annotatorContainer.vue#L584) is not called. This should be done regardless if the loading is successful or not. I think it can even be just called in the watcher of `imageId` and doesn't have to be included in the promise chain at all.

Should I create a new issue for it or can I commit the changes to this issue?

mzur:
You can commit it also to this PR.

lehecht:
> It's the same than what the [backend currently does](https://github.com/biigle/core/blob/b4242255608bcddc2b6f9fc0598ea39515d631bd/resources/views/annotations/show/annotationCanvas.blade.php#L27) with `@can('add-annotation', $image)` but dynamically in the frontend. After the change we probably don't even need the `@can` in the backend any more.

I'm still a bit confused by what you meant. If I write v-if before ":can-edit/delete/add=...", the whole button group disappears. So what exactly should I change in the show.blade and annotationCanvas.blade ?

mzur:
You have to replace `isEditor` with something else that not only checks if the current user is an editor but also if the current image loaded correctly (e.g. with a computed property):
https://github.com/biigle/core/blob/b4242255608bcddc2b6f9fc0598ea39515d631bd/resources/views/annotations/show.blade.php#L67-L69

Then you can remove the `@can` and instead add `v-if="canAdd"` to the drawing controls:
https://github.com/biigle/core/blob/b4242255608bcddc2b6f9fc0598ea39515d631bd/resources/views/annotations/show/annotationCanvas.blade.php#L28

And `v-if="canModify"` to the edit buttons (except the delete button):
https://github.com/biigle/core/blob/b4242255608bcddc2b6f9fc0598ea39515d631bd/resources/views/annotations/show/annotationCanvas.blade.php#L48-L51

And finally `v-if="canDelete"` to the delete button:
https://github.com/biigle/core/blob/b4242255608bcddc2b6f9fc0598ea39515d631bd/resources/views/annotations/show/annotationCanvas.blade.php#L52-L53

You also have to handle the case to display the single measure instrument in case of `v-if="!canAdd"`:
https://github.com/biigle/core/blob/b4242255608bcddc2b6f9fc0598ea39515d631bd/resources/views/annotations/show/annotationCanvas.blade.php#L56-L58

To really make it shine you can add something like this to all elements that now have `v-if`:
```
@cannot('add-annotation') v-cloak @endcannot
```
This will prevent a quick flash of the elements in case the user is not an editor and should never see them.

lehecht:
Thanks!
If I add v-if to the edit and delete control buttons, then a bar is created between the next and lineal button. 
It comes from `<div class="btn-group edit-controls">`. My suggest would be to create two separate divs with the class for the edit and delete buttons. There I could add the corresponding v-ifs directly.

mzur:
> My suggest would be to create two separate divs with the class for the edit and delete buttons. There I could add the corresponding v-ifs and the bar is not shown any more.

This would also result in a different UI than before (because the delete button will have it's own "group"). The UI should not be changed by this, even if it makes the code uglier. Instead, you could add `v-if="canModify || canDelete"` to the single `div`.

resources/assets/js/annotations/annotatorContainer.vue
```diff
@@ -581,20 +592,20 @@ export default {
...
-                    .then(this.cachePreviousAndNext)
                     .catch(this.handleLoadingError)
-                    .finally(this.finishLoading);
+                    .finally(() => {
+                        this.updateUrlSlug();
+                        this.emitImageChanged();
+                        // When everything is loaded, pre-fetch the data of the next and
+                        // previous images so they can be switched fast.
+                        this.cachePreviousAndNext();
+                        this.finishLoading()});
```
mzur, line 608: I feel this is getting too convoluted with the additional logic. Maybe it's time to introduce async/await (for the first time) for better readability. How do you like something like this?

```js
async imageId(id) {
    if (!id) {
        return;
    }

    this.startLoading();
    this.crossOriginError = false;

    try {
        let [image, annotations] = await Vue.Promise.all(this.getImageAndAnnotationsPromises(id));
        this.imageLoadingError = false;
        this.image = image;
        this.annotations = annotations;
        this.maybeUpdateAnnotationMode();
        this.maybeShowTilingInProgressMessage();
    } catch (e) {
        if (e instanceof CrossOriginError) {
            this.crossOriginError = true;
        } else {
            this.imageLoadingError = true;
            this.image = null;
            this.annotations = [];
            Messages.danger(e);
        }
    } finally {
        this.updateUrlSlug();
        this.emitImageChanged();
        // When everything is loaded, pre-fetch the data of the next and
        // previous images so they can be switched fast.
        this.cachePreviousAndNext();
        this.finishLoading();
    }
},
```

This makes `setCurrentImageAndAnnotations()` and `handleLoadingError()` unnecessary. Also, I think you can read/understand it better. Or is it just me because I want to use async/await? :smiley: 

This may also have other implications that I will discuss in the other comments.

resources/views/annotations/show.blade.php
```diff
@@ -64,11 +64,10 @@
...
         <annotation-canvas
-            v-show="!imageLoadingError"
-            :can-add="isEditor"
-            :can-modify="isEditor"
-            :can-delete="isEditor"
-            :image="image"
+            :can-add="canAdd"
+            :can-modify="canModify"
+            :can-delete="canDelete"
+            :image="!imageLoadingError ? image : null"
```
mzur, line 70: With the proposed method above, `image` will be automatically `null` on error.

```suggestion
            :image="image"
```

resources/views/annotations/show/annotationCanvas.blade.php
```diff
@@ -44,18 +43,18 @@
                     @mixin('imageAnnotationPolygonTools')
                 </control-button>
             </div>
-            <div class="btn-group edit-controls">
-                <control-button icon="fa-tag" title="Attach the currently selected label to existing annotations 𝗟" :active="isAttaching" v-on:click="toggleAttaching">
+            <div class="btn-group edit-controls" v-if="canModify || canDelete">
```
mzur, line 46: The `v-cloak` can also be added here so you don't need it on all of the children.

resources/views/annotations/show/annotationCanvas.blade.php
```diff
@@ -44,18 +43,18 @@
...
-                <control-button v-else icon="fa-trash" title="Delete selected annotations 𝗗𝗲𝗹" :disabled="modifyInProgress||!hasSelectedAnnotations" v-on:click="deleteSelectedAnnotations"></control-button>
+                <control-button v-if="canModify" @cannot('add-annotation') v-cloak @endcannot icon="fa-arrows-alt" title="Move selected annotations 𝗠" :active="isTranslating" v-on:click="toggleTranslating" :disabled="modifyInProgress"></control-button>
+                <control-button v-if="hasLastCreatedAnnotation && canDelete" @cannot('add-annotation') v-cloak @endcannot icon="fa-undo" title="Delete the last drawn annotation 𝗕𝗮𝗰𝗸𝘀𝗽𝗮𝗰𝗲" v-on:click="deleteLastCreatedAnnotation"></control-button>
+                <control-button v-else-if="canDelete" @cannot('add-annotation') v-cloak @endcannot icon="fa-trash" title="Delete selected annotations 𝗗𝗲𝗹" :disabled="modifyInProgress||!hasSelectedAnnotations" v-on:click="deleteSelectedAnnotations"></control-button>
             </div>
-        @else
-            <div class="btn-group drawing-controls">
+
+            <div class="btn-group drawing-controls" v-if="!canAdd" @cannot('add-annotation') v-cloak @endcannot>
```
mzur, line 55: I only think of this now: If there is no image, nothing can be measured, of course! So the button can be hidden on error:

```suggestion
            <div class="btn-group drawing-controls" v-if="!canAdd && image" @cannot('add-annotation') v-cloak @endcannot>
```

mzur (changes requested):
This is more complicated than I thought but it's getting really nice!

I've added a suggestion for a refactored watcher method using async/await. Please have a look if you like it. If you use the updated method, you also need to update the `image` watcher method because now `this.image` can be null (`?` was added):

```js
image(image) {
    this.crossOriginError = image?.crossOrigin;
},
```

Thinking about it, maybe `this.imageLoadingError === true` may be equivalent to `this.image === null`, so maybe we don't need the additional `imageLoadingError` variable at all. Before, `this.image` couldn't become `null` once any image was loaded (which wasn't really intended) but with the new async/await function it can become `null` again.

resources/assets/js/annotations/annotatorContainer.vue
lehecht: maybeUpdateAnnotationMode() expects a 'data' argument. This was provided by the setCurrentImageAndAnnotations() promise. Now you cannot draw annotations because "[Vue warn]: Error in callback for watcher "isDefaultInteractionMode": "TypeError: this.modifyInteraction is undefined"". 
I don't get why this error occurs because 'data' was undefined even before this change.

resources/views/annotations/show/annotationCanvas.blade.php
```diff
@@ -43,16 +43,16 @@
...
-                <control-button v-else-if="canDelete" @cannot('add-annotation') v-cloak @endcannot icon="fa-trash" title="Delete selected annotations 𝗗𝗲𝗹" :disabled="modifyInProgress||!hasSelectedAnnotations" v-on:click="deleteSelectedAnnotations"></control-button>
+                <control-button v-if="canModify" icon="fa-arrows-alt" title="Move selected annotations 𝗠" :active="isTranslating" v-on:click="toggleTranslating" :disabled="modifyInProgress"></control-button>
+                <control-button v-if="hasLastCreatedAnnotation && canDelete" icon="fa-undo" title="Delete the last drawn annotation 𝗕𝗮𝗰𝗸𝘀𝗽𝗮𝗰𝗲" v-on:click="deleteLastCreatedAnnotation"></control-button>
+                <control-button v-else-if="canDelete" icon="fa-trash" title="Delete selected annotations 𝗗𝗲𝗹" :disabled="modifyInProgress||!hasSelectedAnnotations" v-on:click="deleteSelectedAnnotations"></control-button>
             </div>
 
-            <div class="btn-group drawing-controls" v-if="!canAdd" @cannot('add-annotation') v-cloak @endcannot>
+            <div class="btn-group drawing-controls" v-if="!canAdd & image" @cannot('add-annotation') v-cloak @endcannot>
```
mzur, line 55:
```suggestion
            <div class="btn-group drawing-controls" v-if="!canAdd && image" @cannot('add-annotation') v-cloak @endcannot>
```

resources/assets/js/annotations/annotatorContainer.vue
mzur: This is unrelated to the `data` argument. `maybeUpdateAnnotationMode()` can also work without it. I'll respond to the error in my new review.

mzur (changes requested):
The changes are fine (only a missing `&`).

The new error happens because `canAdd`, `canModify` and `canDelete` are now initially `false` by default. The `annotationCanvas` component can't handle this, yet. It does not create some interactions:

https://github.com/biigle/core/blob/ac2f57f98b791eed61f8e71bd0f8994b7ddad141/resources/assets/js/annotations/components/annotationCanvas.vue#L307-L320

...or does not enable some things in the `created()` hook:

https://github.com/biigle/core/blob/ac2f57f98b791eed61f8e71bd0f8994b7ddad141/resources/assets/js/annotations/components/annotationCanvas.vue#L749-L758

You should have a look at all places where `canAdd`, `canModify` and `canDelete` occur in the `annotationCanvas.vue` and also in all "subcomponents" in the `annotationCanvas` directory. The new behavior should be that all interactions etc. should be created. But then there are watchers for `canAdd`, `canModify` and `canDelete` that dynamically enable or disable the interactions (with `setActive()` or `Keyboard.on()` and `Keyboard.off()`).

lehecht:
If canAdd, canModify and canDelete is initially false, can I delete the corresponding code in created() and mounted() ? Or should I leave it there?

mzur:
Everything that's creating something (e.g. the ModifyInteraction) or setting event handlers to the interaction should do this independently of the `can*` variables. Only the affected interactions should be initially switched off (with `setActive(false)`). The watchers for `can*` then enable/disable the interactions and enable/disable the keyboard event listeners later.

lehecht:
How should I deal with this.$watch(...) methods? Often they are used to change the can* variables or contain nested methods that do. If I just use the vue-watcher to set the state, then the $watch methods could overwrite them easily.

mzur:
Can you please give an example where a watcher changes the can* variables?

lehecht:
The [magicWandInteraction](https://github.com/biigle/core/blob/38fb572cde3569b210515024b118a57fbb3a5223/resources/assets/js/annotations/components/annotationCanvas/magicWandInteraction.vue#L68) state is set in toggleMagicWandInteraction, which depends on canAdd and isMagicWanding.

mzur:
You could move the `if` inside the watcher method like this:
```js
watch: {
   isMagicWanding(isMagicWanding) {
      if (this.canAdd) {
         this.toggleMagicWandInteraction(isMagicWanding);
      }
   },
},
```

Is there any instance where this wouldn't work?

lehecht:
So the setter may be dependent on can* variables? Because your wrote before:
> [...] setting event handlers to the interaction should do this independently of the `can*` variables
Maybe I misunderstood you?

mzur:
> [...] setting event handlers to the interaction should do this independently of the can* variables

What I meant was that something like this:

https://github.com/biigle/core/blob/38fb572cde3569b210515024b118a57fbb3a5223/resources/assets/js/annotations/components/annotationCanvas.vue#L749-L753

should be done regardless of whether `canModify` is `true` or `false`. All interactions should be created and all (interaction) event handlers should be attached. You then need watchers for can* to enable or disable interactions and keyboard event handlers.

From a quick glance at your recent commits, you seem to have done exactly what I meant. Are the changes ready to review, yet?

In the specific case of the Magic Wand interaction, you could also do `if (magicWandInteraction && this.canAdd)` [here](https://github.com/biigle/core/blob/38fb572cde3569b210515024b118a57fbb3a5223/resources/assets/js/annotations/components/annotationCanvas/magicWandInteraction.vue#L26), which may be a little more elegant.

lehecht:
My changes are ready to be reviewed.

resources/assets/js/annotations/components/annotationCanvas.vue
```diff
@@ -715,6 +715,21 @@ export default {
                 this.modifyInteraction.setActive(defaultMode);
             }
         },
+        canModify(state){
+            if(!state){
+                this.modifyInteraction.setActive(false);
+            }
```
mzur, line 721: I think you need something like this:

```suggestion
            this.modifyInteraction.setActive(state && this.isDefaultInteractionMode);
```

Otherwise the interaction is not enabled initially.

resources/assets/js/annotations/components/annotationCanvas/drawInteractions.vue
```diff
@@ -100,16 +100,29 @@ export default {
...
+        canAdd(state) {
+            if (state) {
+                Keyboard.on('a', this.drawPoint, 0, this.listenerSet);
+                Keyboard.on('s', this.drawRectangle, 0, this.listenerSet);
+                Keyboard.on('d', this.drawCircle, 0, this.listenerSet);
+                Keyboard.on('Shift+d', this.drawEllipse, 0, this.listenerSet);
+                Keyboard.on('f', this.drawLineString, 0, this.listenerSet);
+                Keyboard.on('g', this.drawPolygon, 0, this.listenerSet);
+            }
+            else {
```
mzur, line 117:
```suggestion
            } else {
```

resources/assets/js/annotations/components/annotationCanvas/attachLabelInteraction.vue
```diff
@@ -64,30 +64,37 @@ export default {
                 this.resetInteractionMode();
             }
         },
+        canAdd(state){
```
mzur, line 67:
```suggestion
        canAdd(state) {
```

resources/assets/js/annotations/components/annotationCanvas/attachLabelInteraction.vue
```diff
@@ -64,30 +64,37 @@ export default {
...
+            if(state){
```
mzur, line 68:
```suggestion
            if (state) {
```

resources/assets/js/annotations/components/annotationCanvas/attachLabelInteraction.vue
```diff
@@ -64,30 +64,37 @@ export default {
...
+                Keyboard.on('l', this.toggleAttaching, 0, this.listenerSet);
+                Keyboard.on('Shift+l', this.toggleSwapping, 0, this.listenerSet);
+            }
+            else{
```
mzur, line 72:
```suggestion
            } else {
```

resources/assets/js/annotations/components/annotationCanvas.vue
```diff
@@ -715,6 +715,21 @@ export default {
...
+        },
+        canDelete(state){
```
mzur, line 723:
```suggestion
        canDelete(state) {
```

resources/assets/js/annotations/components/annotationCanvas.vue
```diff
@@ -715,6 +715,21 @@ export default {
...
+            if(state){
```
mzur, line 724:
```suggestion
            if (state) {
```

resources/assets/js/annotations/components/annotationCanvas.vue
```diff
@@ -715,6 +715,21 @@ export default {
...
+                Keyboard.on('Delete', this.deleteSelectedAnnotations, 0, this.listenerSet);
+                Keyboard.on('Backspace', this.deleteLastCreatedAnnotation, 0, this.listenerSet);
+            }
+            else{
```
mzur, line 728:
```suggestion
            } else {
```

resources/assets/js/annotations/components/annotationCanvas/attachLabelInteraction.vue
```diff
@@ -64,30 +64,37 @@ export default {
...
+                attachLabelInteraction.setActive(false);
+                swapLabelInteraction.setActive(false);
```
mzur, line 74: This could lead to an invalid state (see my review comment).

resources/assets/js/annotations/components/annotationCanvas/drawInteractions.vue
```diff
@@ -100,16 +100,29 @@ export default {
...
+                drawInteraction?.setActive(false);
```
mzur, line 118: This could lead to an invalid state (see my review comment).

resources/assets/js/annotations/components/annotationCanvas/magicWandInteraction.vue
```diff
@@ -60,18 +60,30 @@ export default {
             }
         }
     },
+    watch:{
+        isMagicWanding(isMagicWanding) {
+            if (this.canAdd) {
+                this.toggleMagicWandInteraction(isMagicWanding);
+            }
```
mzur, line 67: No need for the `if` here because it's handled in `toggleMagicWand()`.

```suggestion
            this.toggleMagicWandInteraction(isMagicWanding);
```

resources/assets/js/annotations/components/annotationCanvas/magicWandInteraction.vue
```diff
@@ -60,18 +60,30 @@ export default {
...
+        },
+        canAdd(state){
```
mzur, line 69:
```suggestion
        canAdd(state) {
```

resources/assets/js/annotations/components/annotationCanvas/magicWandInteraction.vue
```diff
@@ -60,18 +60,30 @@ export default {
...
+            if(state){
```
mzur, line 70:
```suggestion
            if (state) {
```

resources/assets/js/annotations/components/annotationCanvas/magicWandInteraction.vue
```diff
@@ -60,18 +60,30 @@ export default {
...
+                Keyboard.on('Shift+g', this.toggleMagicWand, 0, this.listenerSet);
+            }
+            else{
```
mzur, line 73:
```suggestion
            } else {
```

resources/assets/js/annotations/components/annotationCanvas/magicWandInteraction.vue
```diff
@@ -60,18 +60,30 @@ export default {
...
+                magicWandInteraction.setActive(false);
```
mzur, line 74: This could lead to an invalid state (see my review comment).

resources/assets/js/annotations/components/annotationCanvas/magicWandInteraction.vue
```diff
@@ -60,18 +60,30 @@ export default {
...
             this.$watch('image', this.maybeUpdateMagicWandSnapshot);
             this.$watch('image', this.maybeSetMagicWandLayer);
-            this.$watch('isMagicWanding', this.toggleMagicWandInteraction);
-        }
     },
     mounted() {
         // Initialize the magic wand interaction here because we have to wait for
         // the non-reactive properties of annotationCanvas to be initialized.
-        if (this.canAdd) {
+        // if (this.canAdd) {
```
mzur, line 86: Please remove the comments and unindent the block.

resources/assets/js/annotations/components/annotationCanvas/polygonBrushInteraction.vue
```diff
@@ -100,33 +100,44 @@ export default {
...
-            Keyboard.on('t', this.togglePolygonFill, 0, this.listenerSet);
-        }
-
-        if (this.canAdd || this.canModify) {
-            this.$watch('interactionMode', this.toggleCurrentInteraction);
+    watch: {
+        interactionMode(mode) {
+            if (this.canAdd || this.canModify) {
+                this.toggleCurrentInteraction(mode);
+            }
```
mzur, line 107: No need for `if` here because it's already handled in `toggleCurrentInteraction()`.

```suggestion
            this.toggleCurrentInteraction(mode);
```

resources/assets/js/annotations/components/annotationCanvas/polygonBrushInteraction.vue
```diff
@@ -100,33 +100,44 @@ export default {
...
+                Keyboard.off('e', this.togglePolygonBrush, 0, this.listenerSet);
+            }
+        },
+        canModify(state) {
+            if (state) {
+                Keyboard.on('r', this.togglePolygonEraser, 0, this.listenerSet);
+                Keyboard.on('t', this.togglePolygonFill, 0, this.listenerSet);
+            }
+            else {
+                shiftClickSelectInteraction.setActive(false);
```
mzur, line 123: This could lead to an invalid state (see my review comment).

I think you can already use `resetInteractionMode()` here because the `toggle*` methods already handle `can*`.

resources/assets/js/annotations/components/annotationCanvas/polygonBrushInteraction.vue
```diff
@@ -100,33 +100,44 @@ export default {
...
+        interactionMode(mode) {
+            if (this.canAdd || this.canModify) {
+                this.toggleCurrentInteraction(mode);
+            }
+        },
+        canAdd(state) {
+            if (state) {
+                Keyboard.on('e', this.togglePolygonBrush, 0, this.listenerSet);
+            }
+            else {
```
mzur, line 113:
```suggestion
            } else {
```

resources/assets/js/annotations/components/annotationCanvas/polygonBrushInteraction.vue
```diff
@@ -100,33 +100,44 @@ export default {
...
+                Keyboard.off('e', this.togglePolygonBrush, 0, this.listenerSet);
+            }
+        },
+        canModify(state) {
+            if (state) {
+                Keyboard.on('r', this.togglePolygonEraser, 0, this.listenerSet);
+                Keyboard.on('t', this.togglePolygonFill, 0, this.listenerSet);
+            }
+            else {
```
mzur, line 122:
```suggestion
            } else {
```

resources/assets/js/annotations/components/annotationCanvas/polygonBrushInteraction.vue
```diff
@@ -100,33 +100,44 @@ export default {
...
+            condition(e) {
+                return click(e) && shiftKeyOnly(e);
+            },
+            style: Styles.highlight,
+            layers: [this.annotationLayer],
+            features: this.selectInteraction.getFeatures(),
+            multi: true,
+        });
+        shiftClickSelectInteraction.on('select', this.handleFeatureSelect);
+        shiftClickSelectInteraction.setActive(false);
```
mzur, line 140: This is not required because the interaction is not added to the map initially. However, the current implementation here is bad because once the `shiftClickSelectInteraction` is added to the map (in `toggleCurrentInteraction()`) it is never removed again. Instead, it should be created and removed just like the `ModifyPolygonBrushInteraction` in `toggleCurrentInteraction`. Maybe you can implement this change right in this PR, too.

resources/assets/js/annotations/components/annotationCanvas/translateInteraction.vue
```diff
@@ -30,21 +30,27 @@ export default {
...
+                translateInteraction.setActive(false);
+                Keyboard.off('m', this.toggleTranslating, 0, this.listenerSet);
+            }
+        }
     },
     mounted() {
         // Initialize the translate interaction here because we have to wait for
         // the non-reactive properties of annotationCanvas to be initialized.
-        if (this.canModify) {
             translateInteraction = new TranslateInteraction({
```
mzur, line 46: Please fix the indentation.

resources/assets/js/annotations/components/annotationCanvas/translateInteraction.vue
```diff
@@ -30,21 +30,27 @@ export default {
                 translateInteraction.setActive(translating);
             }
         },
+        canModify(state){
```
mzur, line 33:
```suggestion
        canModify(state) {
```

resources/assets/js/annotations/components/annotationCanvas/translateInteraction.vue
```diff
@@ -30,21 +30,27 @@ export default {
...
+            if(state){
```
mzur, line 34:
```suggestion
            if (state) {
```

resources/assets/js/annotations/components/annotationCanvas/translateInteraction.vue
```diff
@@ -30,21 +30,27 @@ export default {
...
+                Keyboard.on('m', this.toggleTranslating, 0, this.listenerSet);
+            }
+            else{
```
mzur, line 37:
```suggestion
            } else {
```

resources/assets/js/annotations/components/annotationCanvas/translateInteraction.vue
```diff
@@ -30,21 +30,27 @@ export default {
...
+                translateInteraction.setActive(false);
```
mzur, line 38: This could lead to an invalid state (see my review comment).

resources/assets/js/annotations/components/annotationCanvas/polygonBrushInteraction.vue
lehecht: That kind of implementation also occurs in [translateInteractions](https://github.com/biigle/core/blob/fdde8273c2f2f3dbc5b4cafb2c368219d960fd82/resources/assets/js/annotations/components/annotationCanvas/translateInteraction.vue#L38-L41) and in [magicWandInteraction](https://github.com/biigle/core/blob/fdde8273c2f2f3dbc5b4cafb2c368219d960fd82/resources/assets/js/annotations/components/annotationCanvas/magicWandInteraction.vue#L75-L82). Should the interactions be deleted there too?

resources/assets/js/annotations/components/annotationCanvas/translateInteraction.vue
lehecht: For the next time, could you use the code from my branch, when you highlight something here? Sometimes it's quiet confusing to understand what you really mean, if lines of code are highlighted which don't exist on my branch any more.

resources/assets/js/annotations/components/annotationCanvas/polygonBrushInteraction.vue
mzur: No, the places you have linked add the interactions once to the map and then work with `setActive()` to enable and disable the interactions. This is fine and different to the polygon brush I was talking about. It uses `addInteraction()` and `removeInteraction()`. The `shiftClickSelectInteraction` is only ever added and never removed, which is the problem here.

However, looking at it again, the `shiftClickSelectInteraction` is disabled initially (`setActive(false)`) and never enabled in the file. This must be fixed. You can then work with `setActive()` instead of `add`/`removeInteraction()` here, too. So you can add `shiftClickSelectInteraction` to the map in `mounted()` but disable it and then conditionally disable/enable it in `toggleCurrentInteraction()`.

resources/assets/js/annotations/components/annotationCanvas/translateInteraction.vue
mzur: You mean the links in my [main comment](https://github.com/biigle/core/pull/611#pullrequestreview-1510599393)? I'll try to remember this next time!

If you mean *this* comment, the highlighted line should be in your branch. :thinking:

lehecht: No, I mean this comment. The if-statement in line 37 does not exist on my branch [img-fail-view.](https://github.com/biigle/core/blob/49c058e090eea1a886e18c1c3f9281a427a26b7c/resources/assets/js/annotations/components/annotationCanvas/translateInteraction.vue#L43-L54)

mzur: Ah I see. Line 37 was deleted in your branch, as is shown above. My comment refers to line 38/46 which must be unindented because the `if` is no longer there.

resources/assets/js/annotations/components/annotationCanvas/drawInteractions.vue
```diff
@@ -101,29 +101,25 @@ export default {
...
-                Keyboard.off('f', this.drawLineString, 0, this.listenerSet);
-                Keyboard.off('g', this.drawPolygon, 0, this.listenerSet);
-            }
+        canAdd: {
+            handler(state) {
+                if (!state) {
+                    this.resetInteractionMode();
+                }
+            },
+            immediate: true
```
mzur, line 113: This watcher would be required for all the other interactions, too (e.g. the `attachlabelInteraction`). However, the reset would be called many times then, even though it's required only once. So what about moving this into the main file `annotationCanvas.vue`? It could just be called for `canAdd`, `canModify` and `canDelete`. These don't need to be immediate watchers, too (since the initial state is the default interaction mode).

resources/assets/js/annotations/components/annotationCanvas/drawInteractions.vue
```diff
@@ -82,7 +82,7 @@ export default {
                 drawInteraction = undefined;
             }
 
-            if (this.isDrawing) {
+            if (this.isDrawing && this.canAdd) {
```
mzur, line 85: I think `isDrawing` can only be true if `canAdd` is true, too. This is handled in `draw()`. Please check if the additional constraint is required here.

resources/assets/js/annotations/components/annotationCanvas/translateInteraction.vue
mzur, line 1: Please move the `if (this.canModify)` from the `isTranslating` watcher to the `toggleTranslating()` method to be consistent with the other files.

resources/assets/js/annotations/components/annotationCanvas/drawInteractions.vue
```diff
@@ -101,29 +101,25 @@ export default {
...
+        canAdd: {
+            handler(state) {
+                if (!state) {
+                    this.resetInteractionMode();
+                }
+            },
+            immediate: true
         }
     },
+    mounted() {
```
mzur, line 116:
```suggestion
    created() {
```

resources/assets/js/annotations/components/annotationCanvas/measureInteraction.vue
```diff
@@ -30,7 +30,7 @@ export default {
         toggleMeasuring() {
             if (this.isMeasuring) {
                 this.resetInteractionMode();
-            } else {
+            } else if (this.canModify){
```
mzur, line 33:
```suggestion
            } else if (this.canModify) {
```

mzur, line 33: Should this rather be `if (this.image)`?

resources/assets/js/annotations/components/annotationCanvas/polygonBrushInteraction.vue
```diff
@@ -25,13 +25,13 @@ export default {
         },
         isUsingPolygonFill() {
             return this.interactionMode === 'polygonFill';
-        },
+        }
```
mzur, line 28: I prefer trailing commas.

```suggestion
        },
```

resources/assets/js/annotations/components/annotationCanvas/polygonBrushInteraction.vue
```diff
@@ -99,32 +99,16 @@ export default {
                 this.map.addInteraction(shiftClickSelectInteraction);
             }
         },
+        toggleShiftClickSelectInteraction(mode) {
+            shiftClickSelectInteraction.setActive(this.canModify && (mode === 'polygonEraser' || mode === 'polygonFill'));
```
mzur, line 103: This is probably better (even though I didn't do it myself in `toggleCurrentInteraction()`):

```suggestion
            shiftClickSelectInteraction.setActive(this.canModify && (this.isUsingPolygonEraser || this.isUsingPolygonFill));
```

resources/assets/js/annotations/components/annotationCanvas/polygonBrushInteraction.vue
mzur, line 1: Please remove the:
```
this.map.addInteraction(shiftClickSelectInteraction);
```
(2x) from `toggleCurrentInteraction()` and add it to `mounted()` instead.

mzur, line 1: The logic here could also be refactored to three different watchers for `isUsingPolygonBrush`, `isUsingPolygonEraser` and `isUsingPolygonFill`, each calling a `togglePolygon*Interaction()` method that contains the respective section of `toggleCurrentInteraction()`.

This way, there would be no need for the large `if` and second check for `can*` currently in `toggleCurrentInteraction()`.

But this is just an optional idea. Feel free to dismiss it if you want :wink:

resources/assets/js/annotations/annotatorContainer.vue
```diff
@@ -575,21 +566,35 @@ export default {
...
-                    .then(this.cachePreviousAndNext)
-                    .catch(this.handleLoadingError)
-                    .finally(this.finishLoading);
+        async imageId(id) {
+            if (!id) {
+                return;
+            }
+
+            this.startLoading();
+            this.crossOriginError = this.image?.crossOrigin;
```
mzur, line 575: Why not this?

```suggestion
            this.crossOriginError = false;
```

resources/assets/js/annotations/components/annotationCanvas.vue
```diff
@@ -715,6 +713,20 @@ export default {
...
+        canModify(state) {
+            this.modifyInteraction.setActive(state && this.isDefaultInteractionMode);
+
+        },
+        canDelete(state) {
+            if (state) {
+                Keyboard.on('Delete', this.deleteSelectedAnnotations, 0, this.listenerSet);
+                Keyboard.on('Backspace', this.deleteLastCreatedAnnotation, 0, this.listenerSet);
+            }
+            else {
```
mzur, line 725:
```suggestion
            } else {
```

resources/assets/js/annotations/components/annotationCanvas/magicWandInteraction.vue
```diff
@@ -60,30 +60,30 @@ export default {
...
+            source: this.annotationSource,
+            style: Styles.editing,
+            indicatorPointStyle: Styles.editing,
+            indicatorCrossStyle: Styles.cross,
+            simplifyTolerant: 0.1,
+        });
+        magicWandInteraction.on('drawend', this.handleNewFeature);
+        magicWandInteraction.setActive(false);
+        this.map.addInteraction(magicWandInteraction);
+        Keyboard.on('Shift+g', this.toggleMagicWand, 0, this.listenerSet);
```
mzur, line 86: This line should be moved back to `created()`.

resources/assets/js/annotations/components/annotationCanvas/polygonBrushInteraction.vue
```diff
@@ -99,34 +99,32 @@ export default {
...
+            style: Styles.highlight,
+            layers: [this.annotationLayer],
+            features: this.selectInteraction.getFeatures(),
+            multi: true,
+        });
+        shiftClickSelectInteraction.on('select', this.handleFeatureSelect);
+        shiftClickSelectInteraction.setActive(false);
+        Keyboard.on('r', this.togglePolygonEraser, 0, this.listenerSet);
+        Keyboard.on('t', this.togglePolygonFill, 0, this.listenerSet);
+        Keyboard.on('e', this.togglePolygonBrush, 0, this.listenerSet);
```
mzur, line 127: Adding of the keyboard event listeners should be moved back to `created()` again.

mzur (changes requested):
Almost done, I think :wink:

resources/assets/js/annotations/components/annotationCanvas.vue
```diff
@@ -715,17 +713,16 @@ export default {
...
         },
-        canDelete(state){
-            if(state){
+        canDelete(state) {
+            if (state) {
                 Keyboard.on('Delete', this.deleteSelectedAnnotations, 0, this.listenerSet);
                 Keyboard.on('Backspace', this.deleteLastCreatedAnnotation, 0, this.listenerSet);
             }
-            else{
+            else {
```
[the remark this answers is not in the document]
lehecht: That would be really nice! Usually I work with an auto formatter which is why my code looks not that nice currently.

resources/assets/js/annotations/components/annotationCanvas/drawInteractions.vue
lehecht: Without it the 'select a label' info box is shown to a user with no adding permission, if the short-cut is used.

resources/assets/js/annotations/components/annotationCanvas/magicWandInteraction.vue
lehecht: You mean line 86? Otherwise magicWandInteraction is not initialized yet and an exception will be thrown.

resources/assets/js/annotations/components/annotationCanvas/polygonBrushInteraction.vue
lehecht: Should toggleCurrentInteraction still contain
```
if (currentInteraction) {
                brushRadius = currentInteraction.getBrushRadius();
                this.map.removeInteraction(currentInteraction);
                currentInteraction = null;
            }
```
?

resources/assets/js/annotations/components/annotationCanvas.vue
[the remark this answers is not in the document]
mzur: It seems like this should be possible with ESLint, which we already have. We only need the appropriate configuration.

resources/assets/js/annotations/components/annotationCanvas/polygonBrushInteraction.vue
mzur: `toggleCurrentInteraction()` will be replaced by the `togglePolygon*Interaction()` methods, so they need this block where appropriate. If the block should not be duplicated, you could put it into a `resetCurrentInteraction()` method or something similar.

resources/assets/js/annotations/components/annotationCanvas/magicWandInteraction.vue
mzur: Yes, only the single line.

mzur (changes requested):
Only a few minor things now.

I've tested the UI and most things seemed to work fine. Only the Magic SAM interaction of (biigle/magic-sam) did not handle the initial state correctly. Here is a PR for the fix: https://github.com/biigle/magic-sam/pull/6

resources/assets/js/annotations/components/annotationCanvas.vue
```diff
@@ -713,19 +714,25 @@ export default {
...
+            if (!state) {
+                this.resetInteractionMode();
+            } else {
                 Keyboard.on('Delete', this.deleteSelectedAnnotations, 0, this.listenerSet);
                 Keyboard.on('Backspace', this.deleteLastCreatedAnnotation, 0, this.listenerSet);
             }
-            else {
-                Keyboard.off('Delete', this.deleteSelectedAnnotations, 0, this.listenerSet);
-                Keyboard.off('Backspace', this.deleteLastCreatedAnnotation, 0, this.listenerSet);
-            }
```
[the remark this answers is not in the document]
lehecht: Ah, I thought you meant that resetInteractionMode() would handle this. But why not turn off the other keyboard listeners in the created methods?

lehecht:
> Only a few minor things now.
> 
> I've tested the UI and most things seemed to work fine. Only the Magic SAM interaction of (biigle/magic-sam) did not handle the initial state correctly. Here is a PR for the fix: [biigle/magic-sam#6](https://github.com/biigle/magic-sam/pull/6)

So should I merge the changes from that branch into my branch? Or what should I do?

resources/assets/js/annotations/components/annotationCanvas.vue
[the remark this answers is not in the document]
mzur: > But why not turn off the other keyboard listeners in the created methods?

Because their listening methods ignore the event in case the required ability (`can*`) is not true. At least they should :smiley:

mzur:
You don't have to do anything with biigle/magic-sam. It's a module in a separate repository so this must be handled in a separate PR. I will merge both PRs at the same time and publish the releases for biigle/core and biigle/magic-sam.

Issue #552, opened 2023-02-17
Fix unselect of video anotation in another frame


mzur:
When a single-frame video annotation is selected and then another annotation is selected in another frame (in the video, not in the timeline), the previously selected annotation stays selected. Instead, the first annotation should be unselected in this case.

[Screencast from 17.02.2023 15:04:41.webm](https://user-images.githubusercontent.com/2457311/219676935-68301607-71a3-4f36-a6a0-2cde1f54fc3f.webm)

mzur:
You can click on it in the video or in the timeline at the bottom. In the example video above, the first annotation is selected with a click on the "handle" in the timeline and the second one is selected with a click on the circle in the video.

PR #622, opened 2023-07-03
Annotation selection bug

lehecht:
Resolves bug #552 by using selectInteraction features and annotations array to obtain all selected and deselected annotations.

### DIFF
--- a/resources/assets/js/videos/components/videoScreen.vue
+++ b/resources/assets/js/videos/components/videoScreen.vue
@@ -452,10 +452,18 @@ export default {
         extractAnnotationFromFeature(feature) {
             return feature.get('annotation');
         },
-        handleFeatureSelect(e) {
+        handleFeatureSelect() {
+            // Don't use event.deselected because is not updated for certain cases,
+            // see https://github.com/biigle/core/issues/552
+
+            // New selected annotations
+            let selected = this.selectInteraction.getFeatures().getArray().map(this.extractAnnotationFromFeature);
+            // Old selected annotations
+            let deselected = this.annotations.filter((a) => {return a.selected;})
+                                             .filter((a) => {return !selected.includes(a)});
             this.$emit('select',
-                e.selected.map(this.extractAnnotationFromFeature),
-                e.deselected.map(this.extractAnnotationFromFeature),
+                selected,
+                deselected,
                 this.video.currentTime
             );
         },
--- a/resources/assets/js/videos/components/videoScreen.vue
+++ b/resources/assets/js/videos/components/videoScreen.vue
@@ -453,7 +453,7 @@ export default {
             return feature.get('annotation');
         },
         handleFeatureSelect() {
-            // Don't use event.deselected because is not updated for certain cases,
+            // Don't use event.deselected here because it uses selections in current frame only,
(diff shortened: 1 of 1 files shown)


resources/assets/js/videos/components/videoScreen.vue
```diff
@@ -452,10 +452,18 @@ export default {
         extractAnnotationFromFeature(feature) {
             return feature.get('annotation');
         },
-        handleFeatureSelect(e) {
+        handleFeatureSelect() {
+            // Don't use event.deselected here because it uses selections in current frame only,
+            // see https://github.com/biigle/core/issues/552
+
+            // New selected annotations
+            let selected = this.selectInteraction.getFeatures().getArray().map(this.extractAnnotationFromFeature);
```
mzur, line 460: This should be equivalent to `e.selected.map(this.extractAnnotationFromFeature)`. Why not use it?

resources/assets/js/videos/components/videoScreen.vue
```diff
@@ -452,10 +452,18 @@ export default {
...
+            // Old selected annotations
+            let deselected = this.annotations.filter((a) => {return a.selected;})
+                                             .filter((a) => {return !selected.includes(a)});
```
mzur, line 463: You can use syntactic sugar:

```suggestion
            let deselected = this.annotations.filter(a => a.selected)
                                             .filter(a => !selected.includes(a));
```

mzur (changes requested):
That's almost exactly what I was talking about in my email! I have only minor comments.

resources/assets/js/videos/components/videoScreen.vue
lehecht: If you use this, then you cannot select multiple annotations in one frame with shift any more.

resources/assets/js/videos/components/videoScreen.vue
```diff
@@ -452,10 +452,18 @@ export default {
...
         },
-        handleFeatureSelect(e) {
+        handleFeatureSelect() {
+            // New selected annotations
+            // Don't use event.selected here because then shift select won't work in current frame.
+            let selected = this.selectInteraction.getFeatures().getArray().map(this.extractAnnotationFromFeature);
+            // Old selected annotations
+            // Don't use event.deselected here because it uses selections in current frame only,
+            // see https://github.com/biigle/core/issues/552.
+            let deselected = this.annotations.filter(a => a.selected)
```
mzur, line 462:
```suggestion
            let deselected = this.annotations.filter(a => a.isSelected)
```

This must be `isSelected` because `selected` stores the time if the annotation is selected and the time could be `0` which would be false here. I should rather have called it `selectedTime` :man_shrugging:

resources/assets/js/videos/components/videoScreen.vue
mzur: Ah I see, it's because `e.selected` contains only the newly selected annotations and any other selected annotations would be deselected by the new logic for `deselected` again.

mzur (commented):
Thanks! I'll resolve the remaining changes myself.

mzur (approved):
Well, this was more complicated than it looked at first (again)...

PR #625, opened 2023-07-12
Fix bug with unchanged label color after swap

lehecht:
Annotations were not updated because the annotation watcher was not triggered by the color change. Add a refreshAnnotation method which replaces the old annotation by the new one in the annotationSource.

### DIFF
--- a/resources/assets/js/annotations/annotatorContainer.vue
+++ b/resources/assets/js/annotations/annotatorContainer.vue
@@ -354,7 +354,7 @@ export default {
             if (this.isEditor) {
                 if (annotation.labels.length > 1) {
                     AnnotationsStore.detachLabel(annotation, annotationLabel)
-                        .then(() => this.refreshAnnotation(annotation))
+                        .then(() => this.refreshSingleAnnotation(annotation))
                         .catch(handleErrorResponse);
                 } else if (confirm('Detaching the last label of an annotation deletes the whole annotation. Do you want to delete the annotation?')) {
                     this.handleDeleteAnnotation(annotation);
@@ -450,8 +450,8 @@ export default {
                     .catch(handleErrorResponse);
             }
         },
-        refreshAnnotation(annotation){
-            this.$refs.canvas.refreshAnnotation(annotation);
+        refreshSingleAnnotation(annotation){
+            this.$refs.canvas.refreshSingleAnnotation(annotation);
         },
         handleAttachAllSelected() {
             this.selectedAnnotations.forEach(this.handleAttachLabel);
--- a/resources/assets/js/annotations/annotatorContainer.vue
+++ b/resources/assets/js/annotations/annotatorContainer.vue
@@ -353,10 +353,11 @@ export default {
         handleDetachAnnotationLabel(annotation, annotationLabel) {
             if (this.isEditor) {
                 if (annotation.labels.length > 1) {
-                    return AnnotationsStore.detachLabel(annotation, annotationLabel)
+                    AnnotationsStore.detachLabel(annotation, annotationLabel)
+                        .then(() => this.refreshAnnotation(annotation))
                         .catch(handleErrorResponse);
                 } else if (confirm('Detaching the last label of an annotation deletes the whole annotation. Do you want to delete the annotation?')) {
-                    return this.handleDeleteAnnotation(annotation);
+                    this.handleDeleteAnnotation(annotation);
                 }
             }
         },
@@ -370,7 +371,7 @@ export default {
             // Mark for deletion so the annotation is immediately removed from
             // the canvas. See https://github.com/biigle/annotations/issues/70
             Vue.set(annotation, 'markedForDeletion', true);
-            return AnnotationsStore.delete(annotation)
+            AnnotationsStore.delete(annotation)
                 .catch(function (response) {
                     annotation.markedForDeletion = false;
                     handleErrorResponse(response);
@@ -443,10 +444,7 @@ export default {
                 this.handleAttachLabel(annotation, label)
                     .then(() => {
                         if (lastLabel) {
-                            this.handleDetachAnnotationLabel(annotation, lastLabel)
-                                .then(() => {
-                                    this.refreshAnnotation(annotation);
-                                });
+                            this.handleDetachAnnotationLabel(annotation, lastLabel);
                         }
                     })
                     .catch(handleErrorResponse);
--- a/resources/assets/js/annotations/annotatorContainer.vue
+++ b/resources/assets/js/annotations/annotatorContainer.vue
@@ -353,10 +353,10 @@ export default {
         handleDetachAnnotationLabel(annotation, annotationLabel) {
             if (this.isEditor) {
                 if (annotation.labels.length > 1) {
-                    AnnotationsStore.detachLabel(annotation, annotationLabel)
+                    return AnnotationsStore.detachLabel(annotation, annotationLabel)
                         .catch(handleErrorResponse);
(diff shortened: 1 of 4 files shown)


lehecht:
I was a little too quick with this PR. I forgot to check the video annotations. When I'm finished, I will let you know.

mzur:
There are also too many commits in this PR. Maybe you have to update your branches?

mzur:
I see 67 commits in this PR but only one implements the code changes. Somehow all newer commits from `master` are shown, too. You can fix this by starting a new branch from `master` and cherry-picking the single commit. I think you can even reuse this PR with a force push:

```bash
# Start fresh from master.
git checkout master
# Create a new branch with a different name.
git checkout -b lsb
# Cherry-pick the commit to the new branch.
git cherry-pick 28e448a
# Delete the old branch.
git branch -D label-swap-bug
# Rename the new branch to the old name.
git branch -m label-swap-bug
# Force push to update this PR.
git push -fu origin label-swap-bug
```

lehecht:
You can review my changes now.

lehecht:
> 
> ```shell
> # Start fresh from master.
> git checkout master
> # Create a new branch with a different name.
> git checkout -b lsb
> # Cherry-pick the commit to the new branch.
> git cherry-pick 28e448a


Should I do it like this every time a branch is outdated?

mzur:
> You can review my changes now.

In the future, you can also click on "request review" at the top right of a PR to let me know I should have a look. That should be easier for you.

> Should I do it like this every time a branch is outdated?

No, it was only a problem because you seem to have created the branch from an outdated `master`, then updated the `master` and merged the changes into the branch. Only then you have added the commit that actually implemented the changes in the branch. Usually you only merge the other way around (from the branch to `master`) when the PR is merged. So most of the time, if you don't merge anything into your branch, it should be fine.

resources/assets/js/annotations/components/annotationCanvas.vue
```diff
@@ -596,6 +596,17 @@ export default {
...
+        refreshAnnotation(annotation) {
+            let source = this.annotationSource;
+
+            let newFeature = this.createFeature(annotation);
+            let oldFeature = source.getFeatureById(annotation.id)
+            
+            source.removeFeature(oldFeature);
+            source.addFeature(newFeature);
+            
+
```
mzur, line 609:
```suggestion
```

resources/assets/js/annotations/annotatorContainer.vue
```diff
@@ -353,10 +353,10 @@ export default {
         handleDetachAnnotationLabel(annotation, annotationLabel) {
             if (this.isEditor) {
                 if (annotation.labels.length > 1) {
-                    AnnotationsStore.detachLabel(annotation, annotationLabel)
+                    return AnnotationsStore.detachLabel(annotation, annotationLabel)
                         .catch(handleErrorResponse);
```
mzur, line 357: I think you only need this:

```suggestion
                    return AnnotationsStore.detachLabel(annotation, annotationLabel)
                        .then(() => this.refreshAnnotation(annotation))
                        .catch(handleErrorResponse);
```

This will refresh the annotation if the label is swapped *and* if a label is deleted (which currently still won't update the color). No need to return all the promises and update `handleSwapLabel()`.

resources/assets/js/videos/videoContainer.vue
```diff
@@ -425,7 +425,9 @@ export default {
             this.attachAnnotationLabel(annotation)
                 .then(() => {
                     if (lastLabel) {
-                        this.detachAnnotationLabel(annotation, lastLabel);
+                        this.detachAnnotationLabel(annotation, lastLabel)
+                            .then(Events.$emit('video.swap', annotation));
```
mzur, line 429: I would prefer to do it the same than in the image annotation tool here. Add a "hard" method call to `this.$refs.videoScreen.refreshAnnotation()` to the `detachAnnotationLabel()` method and be done with it. While I've implemented this global `Event` bus at some point, I wouldn't use it any more because you never know from where the event originated and where the listeners are.

resources/assets/js/videos/components/videoScreen/annotationPlayback.vue
```diff
@@ -125,6 +126,15 @@ export default {
                 this.updateGeometry(feature, time);
             });
         },
+        refreshAnnotation(annotation) {
```
mzur, line 129: Since there also is a `refreshAnnotations` method here, I would rename the methods to `refreshSingleAnnotation` and `refreshAllAnnotations` to make the distinction more clear.

resources/assets/js/videos/videoContainer.vue
mzur, line 429: As an aside: Line 429 would not work correctly because the event would be dispatched immediately and not after the promise is resolved. This would be correct:

```suggestion
                            .then(() => Events.$emit('video.swap', annotation));
```

mzur (changes requested):
It does what it should but I would structure it a bit differently. This way we can fix the same issue when a label is deleted, too.

resources/assets/js/videos/components/videoScreen/annotationPlayback.vue
```diff
@@ -7,6 +7,7 @@ import Point from '@biigle/ol/geom/Point';
 import Polygon from '@biigle/ol/geom/Polygon';
 import Rectangle from '@biigle/ol/geom/Rectangle';
 import {getRoundToPrecision} from '../../utils';
+import Events from '../../../core/events';
```
mzur, line 10:
```suggestion
```

resources/assets/js/videos/components/videoScreen/annotationPlayback.vue
```diff
@@ -226,10 +236,11 @@ export default {
         },
     },
     created() {
-        this.$on('refresh', this.refreshAnnotations);
+        Events.$on('video.swap', this.refreshAnnotation);
```
mzur, line 239:
```suggestion
```

mzur (commented):
Looks good, thanks! Two lines are no longer needed. I will remove them myself.

mzur (approved):

dlangenk:
@mzur Is this already live? Because the color does not change and the old labels are not "really" deleted, i.e. they still appear in the annotations list in the sidebar. However, if you click the delete button (the little x) sometimes an error occurs: No query results for model [Biigle\ImageAnnotationLabel] 15900345.

mzur:
The label swapping only works for your own labels. If the annotation has a label of a different user, your label is attached and the other label is kept, too. If you want to replace labels of other users you can use Largo (with the "force" button pressed) or you have to detach the other label with the "x".

I don't know why you get the error. Can you provide steps to reproduce it?

dlangenk:
Shouldn't this follow the same logic as Largo, with experts and admins etc.? 

Not sure about reproducing it but the best bet is:

1. Label swap a lot of labels from another user
2. Manually click on the little 'x' (quite quickly) of the "old" labels.
3. error

mzur:
> Shouldn't this follow the same logic as Largo, with experts and admins etc.?

It does. Largo has the "force" button that needs to be active so experts/admins can delete or replace labels of other users. The label swap tool doesn't have this button. We could add another button next to the label swap button (with the same icon but with a red background) that experts/admins can use to swap labels of other users. What do you think?

> Not sure about reproducing it but the best bet is: [...]

This would make sense. If the requests to swap the labels are still pending, you still see the labels in the sidebar. When you try to delete them, the backend eventually catches up and the old labels no longer exist.

PR #627, opened 2023-07-14
Fix highlighting for unchanged values in label tree

lehecht:
Highlighting for unchanged values didn't work because the watcher was not triggered. Replace the watcher by an click and key event to listen for input submission.


mzur (commented):
Thanks!

mzur (approved):

PR #629, opened 2023-07-17
Fix unhandled error for undefinded events

lehecht:
Catch undefined events for label children/siblings and return false.


mzur:
Can you please add something like "resolves #xxx" in the pull request description as well so the PR is automatically linked with the issue?

PR #630, opened 2023-07-17
Prevent linking annotations of different labels

### LINKED ISSUE #613: Prevent linking video annotations with different labels
It is possible to link whole frame video annotations of different labels, which causes an error on the console.
Allow only linking annotations of the same label and update the manual accordingly.

lehecht:
Resolves #613.

Disabled link-button for annotations with different label_id an updated manual accordingly.


resources/assets/js/videos/components/videoScreen/modifyInteractions.vue
```diff
@@ -30,9 +30,10 @@ export default {
                 allowedSplitShapes.indexOf(this.selectedAnnotations[0].shape) === -1;
         },
         cannotLinkAnnotations() {
-            return this.selectedAnnotations.length !== 2 
-            || this.selectedAnnotations[0].shape_id !== this.selectedAnnotations[1].shape_id
-            || this.selectedAnnotations[0].labels[0].label_id !== this.selectedAnnotations[1].labels[0].label_id;
+            return this.selectedAnnotations.length !== 2
+                || this.selectedAnnotations[0].shape_id !== this.selectedAnnotations[1].shape_id
+                || this.selectedAnnotations[0].labels.length !== this.selectedAnnotations[1].labels.length
+                || !this.getAnnotationLabelIds(0).every(labelId => this.getAnnotationLabelIds(1).includes(labelId));
```
mzur, line 36: I still have two comments about this:

1. The method is called `getAnnotationLabelIds` but it really is `getSelectedAnnotationLabelIds`.

2. It probably has no performance impact whatsoever but the repeated method call inside the `every()` callback just irks me :smile:.

What about a `this.labelsAreIdentical(...this.selectedAnnotations)` method?

```js
labelsAreIdentical(a, b) {
   let labelIdsA = a.labels.map(l => l.label_id);
   let labelIdsB = b.labels.map(l => l.label_id);

   return labelIdsA.every(id => labelIdsB.includes(id));
}
```

mzur (approved):

PR #632, opened 2023-07-28
Utf8 bug

lehecht:


tests/php/Http/Controllers/Api/Volumes/MetadataControllerTest.php
```diff
@@ -76,6 +76,19 @@ public function testStoreImageMetadata()
...
     }
 
+    public function testStoreStringMetadata()
+    {
+        $id = $this->volume()->id;
+
+        $this->beAdmin();
+
+        $this->postJson("/api/v1/volumes/{$id}/metadata", ['metadata_csv' => "metadata_string"])
+        ->assertStatus(422);
```
mzur, line 86:
```suggestion
            ->assertStatus(422);
```

mzur (changes requested):
Now I understand the problem better. It seems that there was invalid user input (a string instead of a file). So in addition to also accepting a string in the `Utf8` rule, we should prepend [`bail`](https://laravel.com/docs/9.x/validation#stopping-on-first-validation-failure) to the [validation rules](https://github.com/biigle/core/blob/master/app/Http/Requests/StoreVolumeMetadata.php#L42-L47), so they fail already if the value is not a file and skip the remaining rules.

app/Http/Requests/StoreVolumeMetadata.php
```diff
@@ -40,12 +40,12 @@ public function rules()
...
         return [
             'metadata_csv' => [
-                'required_without_all:metadata_text,ifdo_file',
+                'bail:required_without_all:metadata_text,ifdo_file',
                 'file',
                 'mimetypes:text/plain,text/csv,application/csv',
                 new Utf8,
             ],
-            'metadata_text' => 'required_without_all:metadata_csv,ifdo_file',
+            'metadata_text' => 'bail:required_without_all:metadata_csv,ifdo_file',
```
mzur, line 48: This field doesn't use `Utf8` so `bail` is not required.

```suggestion
            'metadata_text' => 'required_without_all:metadata_csv,ifdo_file',
```

app/Http/Requests/StoreVolumeMetadata.php
```diff
@@ -40,12 +40,12 @@ public function rules()
     {
         return [
             'metadata_csv' => [
-                'required_without_all:metadata_text,ifdo_file',
+                'bail:required_without_all:metadata_text,ifdo_file',
```
mzur, line 43: `bail` must be added as a separate array element of the validation rules.

```suggestion
                'bail',
                'required_without_all:metadata_text,ifdo_file',
```

tests/php/Http/Controllers/Api/Volumes/MetadataControllerTest.php
mzur, line 1: It looks like your code formatter was active here. Is this the configured coding style from `.php_cs`? General formatting like this should be done in a separate PR, as it is unrelated to this issue.

mzur (changes requested):

app/Http/Requests/StoreVolumeMetadata.php
lehecht: ah okay. Can you maybe explain what the difference is between these two? It seemed to work for me.

mzur (approved):

PR #634, opened 2023-08-07
Add checks for (non-)simple polygons

lehecht:
Shapes are checked for overlapping points. Rectangles need 4 vertices, lines need at least 2 vertices and polygons need at least 3 vertices that do not overlap. Another solution could be realized with JSTS.

Still missing: Solution for bowties, dangling lines and pinch points.

Open question: Should [JSTS](https://github.com/bjornharrtell/jsts) be used?

### DIFF
--- a/resources/assets/js/annotations/components/annotationCanvas.vue
+++ b/resources/assets/js/annotations/components/annotationCanvas.vue
@@ -514,8 +514,7 @@ export default {
             if (this.hasSelectedLabel) {
                 let geometry = e.feature.getGeometry();
                 if (geometry.getType() === 'Polygon') {
-                    let polygonValidator = new PolygonValidator(e.feature);
-                    if (polygonValidator.isInvalidPolygon()) {
+                    if (PolygonValidator.isInvalidPolygon(e.feature)) {
                         this.$emit('is-invalid-polygon');
                         this.annotationSource.once('change', () => {
                             if (this.annotationSource.hasFeature(e.feature)) {
@@ -525,7 +524,7 @@ export default {
                         return;
                     }
 
-                    e.feature = polygonValidator.getValidPolygon();
+                    e.feature = PolygonValidator.makePolygonSimple(e.feature);
                 }
 
                 e.feature.set('color', this.selectedLabel.color);
--- a/resources/assets/js/annotations/components/annotationCanvas.vue
+++ b/resources/assets/js/annotations/components/annotationCanvas.vue
@@ -531,8 +531,6 @@ export default {
                     shape: geometry.getType(),
                     points: this.getPoints(geometry),
                 }, removeCallback);
-            } else if (validDrawing) {
-                this.annotationSource.removeFeature(e.feature);
             } else {
                 let source = this.annotationSource;
                 source.on('change', (evtChange) => {
--- a/resources/assets/js/annotations/components/annotationCanvas.vue
+++ b/resources/assets/js/annotations/components/annotationCanvas.vue
@@ -412,6 +412,7 @@ export default {
                     return this.featureRevisionMap[feature.getId()] !== feature.getRevision();
                 })
                 .map((feature) => {
+                    PolygonValidator.makePolygonSimple(feature);
                     return {
                         id: feature.getId(),
                         image_id: feature.get('annotation').image_id,
--- a/resources/assets/js/annotations/components/annotationCanvas.vue
+++ b/resources/assets/js/annotations/components/annotationCanvas.vue
@@ -533,7 +533,7 @@ export default {
                 }, removeCallback);
             } else {
                 let source = this.annotationSource;
-                source.on('change', (evtChange) => {
+                source.once('change', (evtChange) => {
                     if (evtChange.target.getState() === 'ready' && source.hasFeature(e.feature)) {
                         source.removeFeature(e.feature);
                         this.$emit('has-invalid-shape', e.feature.getGeometry().getType());
--- a/resources/assets/js/annotations/components/annotationCanvas.vue
+++ b/resources/assets/js/annotations/components/annotationCanvas.vue
@@ -524,7 +524,7 @@ export default {
                         return;
                     }
 
-                    e.feature = PolygonValidator.makePolygonSimple(e.feature);
+                    PolygonValidator.makePolygonSimple(e.feature);
                 }
 
                 e.feature.set('color', this.selectedLabel.color);
--- a/resources/assets/js/annotations/components/annotationCanvas.vue
+++ b/resources/assets/js/annotations/components/annotationCanvas.vue
@@ -516,7 +516,7 @@ export default {
                 if (geometry.getType() === 'Polygon') {
                     let polygonValidator = new PolygonValidator(e.feature);
                     if (polygonValidator.isInvalidPolygon()) {
-                        this.$emit('has-invalid-polygon');
+                        this.$emit('is-invalid-polygon');
                         this.annotationSource.once('change', () => {
                             if (this.annotationSource.hasFeature(e.feature)) {
                                 this.annotationSource.removeFeature(e.feature);
--- a/resources/assets/js/annotations/components/annotationCanvas.vue
+++ b/resources/assets/js/annotations/components/annotationCanvas.vue
@@ -1,4 +1,5 @@
 <script>
+import * as PolygonValidator from '../ol/PolygonValidator';
 import AnnotationTooltip from './annotationCanvas/annotationTooltip';
 import AttachLabelInteraction from './annotationCanvas/attachLabelInteraction';
 import CanvasSource from '@biigle/ol/source/Canvas';
@@ -40,12 +41,11 @@ import ZoomifySource from '@biigle/ol/source/Zoomify';
 import ZoomLevel from './annotationCanvas/zoomLevel';
 import ZoomToExtentControl from '@biigle/ol/control/ZoomToExtent';
 import ZoomToNativeControl from '../ol/ZoomToNativeControl';
-import * as PolygonValidator from '../ol/PolygonValidator';
-import { click as clickCondition } from '@biigle/ol/events/condition';
-import { defaults as defaultInteractions } from '@biigle/ol/interaction'
-import { getCenter } from '@biigle/ol/extent';
-import { shiftKeyOnly as shiftKeyOnlyCondition } from '@biigle/ol/events/condition';
-import { singleClick as singleClickCondition } from '@biigle/ol/events/condition';
+import {click as clickCondition} from '@biigle/ol/events/condition';
+import {defaults as defaultInteractions} from '@biigle/ol/interaction'
+import {getCenter} from '@biigle/ol/extent';
+import {shiftKeyOnly as shiftKeyOnlyCondition} from '@biigle/ol/events/condition';
+import {singleClick as singleClickCondition} from '@biigle/ol/events/condition';
 
 
 /**
@@ -516,6 +516,8 @@ export default {
                 if (geometry.getType() === 'Polygon') {
                     if (PolygonValidator.isInvalidPolygon(e.feature)) {
                         this.$emit('is-invalid-polygon');
+                        // This must be done in the change event handler.
+                        // Not exactly sure why.
                         this.annotationSource.once('change', () => {
                             if (this.annotationSource.hasFeature(e.feature)) {
                                 this.annotationSource.removeFeature(e.feature);
--- a/resources/assets/js/annotations/components/annotationCanvas.vue
+++ b/resources/assets/js/annotations/components/annotationCanvas.vue
@@ -510,7 +510,9 @@ export default {
             return this.convertPointsFromOlToDb(points);
         },
         handleNewFeature(e) {
-            if (this.hasSelectedLabel) {
+            let validDrawing = this.hasValidPoints(e);
+
+            if (this.hasSelectedLabel && validDrawing) {
                 let geometry = e.feature.getGeometry();
                 e.feature.set('color', this.selectedLabel.color);
 
@@ -529,9 +531,36 @@ export default {
                     shape: geometry.getType(),
                     points: this.getPoints(geometry),
                 }, removeCallback);
-            } else {
+            } else if (validDrawing) {
                 this.annotationSource.removeFeature(e.feature);
+            } else {
+                let source = this.annotationSource;
+                source.on('change', (evtChange) => {
+                    if (evtChange.target.getState() === 'ready' && source.hasFeature(e.feature)) {
+                        source.removeFeature(e.feature);
+                        this.$emit('has-invalid-shape', e.feature.getGeometry().getType());
+                    }
+
+                });
             }
+        },
+        hasValidPoints(e) {
+            let geometry = e.feature.getGeometry();
+            let points = geometry.getCoordinates();
+
+            switch (geometry.getType()) {
+                case 'LineString':
+                    return this.getEntrySetSize(points) >= 2;
+                case 'Rectangle':
+                    return this.getEntrySetSize(points[0]) === 4;
+                case 'Polygon':
+                    return this.getEntrySetSize(points[0]) >= 3;
+                default:
+                    return true;
+            }
+        },
+        getEntrySetSize(points){
+            return new Set(points.map(xy => String(xy))).size;
         },
         deleteSelectedAnnotations() {
             if (!this.modifyInProgress && this.hasSelectedAnnotations && confirm('Are you sure you want to delete all selected annotations?')) {
--- a/resources/assets/js/annotations/components/annotationCanvas.vue
+++ b/resources/assets/js/annotations/components/annotationCanvas.vue
@@ -23,6 +23,10 @@ import MousePosition from './annotationCanvas/mousePosition';
 import MouseWheelZoom from '@biigle/ol/interaction/MouseWheelZoom';
 import Point from '@biigle/ol/geom/Point';
 import Polygon from '@biigle/ol/geom/Polygon';
+import MultiPolygon from '@biigle/ol/geom/MultiPolygon';
+import MultiPoint from '@biigle/ol/geom/MultiPoint';
+import LinearRing from '@biigle/ol/geom/LinearRing';
+import MultiLineString from '@biigle/ol/geom/MultiLineString';
 import PolygonBrushInteraction from './annotationCanvas/polygonBrushInteraction';
 import Projection from '@biigle/ol/proj/Projection';
 import Rectangle from '@biigle/ol/geom/Rectangle';
@@ -40,11 +44,18 @@ import ZoomifySource from '@biigle/ol/source/Zoomify';
 import ZoomLevel from './annotationCanvas/zoomLevel';
 import ZoomToExtentControl from '@biigle/ol/control/ZoomToExtent';
 import ZoomToNativeControl from '../ol/ZoomToNativeControl';
-import {click as clickCondition} from '@biigle/ol/events/condition';
-import {defaults as defaultInteractions} from '@biigle/ol/interaction'
-import {getCenter} from '@biigle/ol/extent';
-import {shiftKeyOnly as shiftKeyOnlyCondition} from '@biigle/ol/events/condition';
-import {singleClick as singleClickCondition} from '@biigle/ol/events/condition';
+import { click as clickCondition } from '@biigle/ol/events/condition';
+import { defaults as defaultInteractions } from '@biigle/ol/interaction'
+import { getCenter } from '@biigle/ol/extent';
+import { shiftKeyOnly as shiftKeyOnlyCondition } from '@biigle/ol/events/condition';
+import { singleClick as singleClickCondition } from '@biigle/ol/events/condition';
+import JstsPolygon from 'jsts/org/locationtech/jts/geom/Polygon';
+import JstsMultiPolygon from 'jsts/org/locationtech/jts/geom/MultiPolygon';
+import JstsLinearRing from 'jsts/org/locationtech/jts/geom/LinearRing';
+import OL3Parser from 'jsts/org/locationtech/jts/io/OL3Parser';
+import Polygonizer from 'jsts/org/locationtech/jts/operation/polygonize/Polygonizer';
+import Monkey from 'jsts/org/locationtech/jts/monkey';
+import Geometry from '@biigle/ol/geom/Geometry';
 
 
 /**
@@ -512,9 +523,8 @@ export default {
         handleNewFeature(e) {
             if (this.hasSelectedLabel) {
                 let geometry = e.feature.getGeometry();
-                let points = this.getPoints(geometry);
-
                 if (geometry.getType() === 'Polygon') {
+
                     if (this.isInvalidPolygon(e)) {
                         this.$emit('has-invalid-polygon');
                         this.annotationSource.once('change', () => {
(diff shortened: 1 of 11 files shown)


mzur (commented):
You should be able to detect "pinch points" with a check if the `coordinateSetSize` has the same size than the original coordinate array. If not, there are at least two identical points. If the annotation is still valid after superfluous points are removed, it should be created.

I think "dangling lines" could be cleaned with an additional check if 3 subsequent points have either the same x or y coordinates. If this is the case, delete the point in the middle.

"Bow ties" are still a problem and not easily detectable or fixable. Maybe we should try a solution with JSTS? But I don't want to include the whole library. Instead, only the necessary modules should be imported. Still, we should have a look at the additional bundle size. It's currently at 1.8 MB already and I really don't want it to become larger than 2 MB.

These checks should also be implemented for the video annotation tool. I think at some point we should consolidate the code of the two tools, as many things are duplicated.

Finally, the backend should get additional checks for the respective API endpoints, so "pinch points" and "dangling lines" are rejected.

resources/assets/js/annotations/components/annotationCanvas.vue
```diff
@@ -529,10 +531,37 @@ export default {
...
                 this.annotationSource.removeFeature(e.feature);
+            } else {
+                let source = this.annotationSource;
+                source.on('change', (evtChange) => {
+                    if (evtChange.target.getState() === 'ready' && source.hasFeature(e.feature)) {
+                        source.removeFeature(e.feature);
+                        this.$emit('has-invalid-shape', e.feature.getGeometry().getType());
+                    }
+
+                });
```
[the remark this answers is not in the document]
lehecht: > Also, why emit an event and handle the error message in the container instead of directly in the annotationCanvas?

Because the same was done when a[ label was not selected](https://github.com/biigle/core/blob/67c342464023a29d1c8a74b566ec30836ff0c384/resources/assets/js/annotations/components/annotationCanvas.vue#L557-L564) before annotation. I just copied that for this issue.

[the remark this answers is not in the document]
lehecht: > This will add a new event listener each time a new (invalid) annotation is created. Why the event listener instead of just calling `removeFeature()`?

Because then you get this error: `TypeError: this.featureChangeKeys_[featureKey] is undefined.`. It has something to do with asynchronous methods which is why the feature is not there yet or anymore. The problem is that the feature is still displayed on the canvas. Only after refreshing the page, the feature disappears.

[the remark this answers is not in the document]
mzur: Ok. Then please use `once()` instead of `on()`. Also, the `$emit` could be called immediately (outside of the event handler). Why is the `evtChange.target.getState() === 'ready'` required?

I think you can also remove the `else if (validDrawing)` case above. This would cause the same error and (since the error never happened) this case can/should never happen. Drawing is aborted if no label is selected and the additional check for `hasSelectedLabel` should not be required.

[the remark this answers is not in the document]
mzur: Ok. Then please use `once()` instead of `on()`. Also, the `$emit` could be called immediately (outside of the event handler). Why is the `evtChange.target.getState() === 'ready'` required?

I think you can also remove the `else if (validDrawing)` case above. This would cause the same error and (since the error never happened) this case can/should never happen. Drawing is aborted if no label is selected and the additional check for `hasSelectedLabel` should not be required.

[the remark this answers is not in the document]
lehecht: I got the solution with the event listener from stackoverflow and forgot to check if this condition was necessary. I have removed it.

lehecht:
> I think "dangling lines" could be cleaned with an additional check if 3 subsequent points have either the same x or y coordinates. If this is the case, delete the point in the middle.

It is possible to draw a horizontal or vertical line with the magic wand tool. So this check cannot be applied to this areas which makes it a bit more difficult.

mzur:
I understood that the cleaning is particularly relevant for the polygons of the magic wand tool. Can you give an example for "3 subsequent points have either the same x or y coordinates" where the point in the middle should *not* be deleted?

Please use the "re-request review" button above once you want me to look over the code again.

lehecht:
Currently if the user draws a self-intersecting polygon, the biggest part is kept and the rest is discarded. Should I also add an info-box for this? Maybe it is a bit confusing for the user when part of its drawing just disappears.

By the way, I have used the polygon area to determine the biggest part. There are cases where a polygon with more points can be smaller than the part with fewer one. If you don't like this approach, let me know.

mzur:
Note to myself: JSTS seems to increase the bundle size from 1.77 MB to 2.11 MB.

dlangenk:
> Currently if the user draws a self-intersecting polygon, the biggest part is kept and the rest is discarded. Should I also add an info-box for this? Maybe it is a bit confusing for the user when part of its drawing just disappears.
> 
> By the way, I have used the polygon area to determine the biggest part. There are cases where a polygon with more points can be smaller than the part with fewer one. If you don't like this approach, let me know.

I think it is fine this way. This check is rather for involuntary use of self-intersecting polygons and then it should work just fine.

resources/assets/js/annotations/components/annotationCanvas.vue
```diff
@@ -524,10 +540,11 @@ export default {
                         // Do nothing in this case.
                     }
                 };
+                let points = this.getPoints(geometry);
 
                 this.$emit('new', {
                     shape: geometry.getType(),
-                    points: this.getPoints(geometry),
+                    points: points,
```
mzur, line 547: This can probably be reverted.

resources/assets/js/annotations/ol/MagicWandInteraction.js
```diff
@@ -195,7 +195,8 @@ class MagicWandInteraction extends PointerInteraction {
         if (this.isShowingCross) {
             this.sketchSource.removeFeature(this.sketchFeature);
         } else {
-            this.dispatchEvent({type: 'drawend', feature: this.sketchFeature});
+            this.filterSketchCoordinates();
```
mzur, line 198: The magic wand interaction creates a polygon at the end. Is the filtering really needed here when the polygon is also checked by the PolygonValidator?

resources/assets/js/videos/videoContainer.vue
```diff
@@ -274,6 +277,24 @@ export default {
...
+                    this.removeAnnotation(tmpAnnotation);
+                    return;
+                }
+
+                // If polygon is self-intersecting, create valid polygon
+                let validPolygon = polygonValidator.getValidPolygon();
+                let twoDimCoords = validPolygon.getGeometry().getCoordinates();
+                annotation.points = [twoDimCoords[0].flat()];
+            }
+
```
mzur, line 297: I think this rather belongs into `js/videos/components/videoScreen/drawInteractions.vue`. You can add it to [`extendPendingAnnotation()`](https://github.com/biigle/core/blob/ed7d8a58e01f84d9602eee59a974442995c11e58/resources/assets/js/videos/components/videoScreen/drawInteractions.vue#L165).

resources/assets/js/videos/videoContainer.vue
```diff
@@ -285,6 +306,13 @@ export default {
                     }
                 });
         },
+        getFeatureFromAnnotation(annotation){
+            let points = annotation.points;
+            let xValues = points[0].filter((_,i) => {return i%2===0;});
+            let yValues = points[0].filter((_,i) => {return i%2===1;});
+            let coords = xValues.map((x,i) => [x,yValues[i]]);
+            return new Feature({geometry: new Polygon([coords])});
+        },
```
mzur, line 315: This function should not be necessary if the polygon is checked before it is appended to the "pending annotation" (see comment above). You should be able to use the original Feature directly there.

mzur (changes requested):
I've only had a look here and haven't tested it yet. I'll do that in my next review.

resources/assets/js/videos/videoContainer.vue
lehecht: Shall I do the same with the code in [annotationCanvas](https://github.com/biigle/core/blob/0316882b5e6a4978f3b61e6dc4f8bbec227a2b89/resources/assets/js/annotations/components/annotationCanvas.vue#L516-L529)?

mzur: I think not. The new code is added to `handleNewFeature()` which is used in several of the annotationCanvas mixins (draw, measure, magic wand, polygon brush). So the method should best stay there and the updates be applied there, too.

resources/assets/js/annotations/ol/PolygonValidator.js
```diff
@@ -26,7 +31,7 @@ class PolygonValidator {
 
     getValidPolygon() {
         // Check if polygon is self-intersecting
-        const parser = new OL3Parser();
+        let parser = new OL3Parser();
```
mzur, line 34: Why not let it be a `const`?

resources/assets/js/annotations/ol/PolygonValidator.js
```diff
@@ -37,22 +42,25 @@ class PolygonValidator {
...
-        return this.polygon;
+        // Divide non-simple polygon at cross points into several smaller polygons
+        // and translate back to ol geometry
+        let polygons = this.jstsValidate(jstsGeom)['array'].map(p => parser.write(p));
 
+        if (polygons.length > 1) {
+            // Select biggest part
+            let greatestPolygon = this.getGreatestPolygon(polygons);
+            // Only change coordinates because object references are in use
+            this.polygon.getGeometry().setCoordinates(greatestPolygon.getCoordinates());
```
mzur, line 60: This modifies the original polygon, which may be unexpected. Later, you use `setCoordinates()` again (in the annotation canvas/video screen). The validation could be implemented in a functional style instead of a class-based style. So you just implement and export the `isInvalidPolygon(feature)` and `makeValidPolygon(feature)` functions here (without putting them into a class and with all the auxiliary functions they need in this file but not exported). Then `makeValidPolygon()` can directly modify the polygon that was passed as argument.

resources/assets/js/videos/components/videoScreen/drawInteractions.vue
```diff
@@ -163,6 +164,23 @@ export default {
...
+                let polygonValidator = new PolygonValidator(e.feature);
+                if (polygonValidator.isInvalidPolygon()) {
+                    // Disallow polygons with less than three non-overlapping points
+                    this.$emit('is-invalid-polygon')
+                    this.removeFeature(e.feature)
+                    return;
+                }
+
+                // If polygon is self-intersecting, create valid polygon
+                let validPolygon = polygonValidator.getValidPolygon();
```
mzur, line 178: The method names are a little unintuitive IMO. Why does it fail if the polygon is "invalid" if there is a method that can make it "valid" later? Maybe call the second method "cleanPolygon" or something like that.

resources/assets/js/annotations/components/annotationCanvas.vue
```diff
@@ -512,6 +513,21 @@ export default {
         handleNewFeature(e) {
             if (this.hasSelectedLabel) {
                 let geometry = e.feature.getGeometry();
+                if (geometry.getType() === 'Polygon') {
+                    let polygonValidator = new PolygonValidator(e.feature);
+                    if (polygonValidator.isInvalidPolygon()) {
+                        this.$emit('is-invalid-polygon');
+                        this.annotationSource.once('change', () => {
```
mzur, line 520: Why do this in an event handler instead of immediately?

resources/assets/js/annotations/ol/MagicWandInteraction.js
```diff
@@ -375,6 +375,67 @@ class MagicWandInteraction extends PointerInteraction {
...
+                    i++;
+                }
+            } catch (err) {
+                break;
+            }
+            endNotReached = i < coordinates.length - 2;
+        }
+
+        return coordinates;
+    }
```
mzur, line 438: These are no longer needed, I think.

mzur (changes requested):

resources/assets/js/annotations/ol/PolygonValidator.js
lehecht: I changed this file a lot, where the parser was initialized in the constructor at some point. When changing everything back, I forgot that this was a const.

resources/assets/js/annotations/components/annotationCanvas.vue
lehecht: The 'requires-selected-label' issue is handled the same way, where an event is fired. So I just copied that.
But it's no problem to change that.

mzur: I see, so there is probably more into it. But I don't remember why I did it that way :smiley: If in doubt and it works like that, just leave it.

mzur (commented):
You didn't request a review yet but I'll still leave one :wink:

1. While the static methods of `PolygonValidator` essentially work the same, you don't have to wrap them with a class. You can just declare the functions and export them from the file. This is just a comment, you don't have to change this since it's working fine.

2. `makePolygonSimple` still modifies the feature that is passed as an argument internally but also returns it. I think this is risky, see my comments below.

resources/assets/js/annotations/ol/PolygonValidator.js
```diff
@@ -43,27 +46,30 @@ class PolygonValidator {
...
             // Select biggest part
-            let greatestPolygon = this.getGreatestPolygon(polygons);
+            let greatestPolygon = getGreatestPolygon(polygons);
             // Only change coordinates because object references are in use
-            this.polygon.getGeometry().setCoordinates(greatestPolygon.getCoordinates());
+            feature.getGeometry().setCoordinates(greatestPolygon.getCoordinates());
         }
 
-        return this.polygon;
+        return feature;
```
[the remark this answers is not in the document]
lehecht: If I do this, the polygon created before will not be removed because of the removeCallback.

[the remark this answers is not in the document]
mzur: Right. What I mean is that you return `greatestPolygon` and then do this outside the function:

```js
let validPolygonGeometry = PolygonValidator.makePolygonSimple(e.feature);
e.feature.getGeometry().setCoordinates(validPolygonGeometry.getCoordinates());
```

Or you do this:

```js
PolygonValidator.makePolygonSimple(e.feature);
```

without returning anything to make clear that the feature is modified internally.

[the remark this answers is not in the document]
lehecht: ah. Okay, then I had misunderstood you. Thanks for the clarification!

resources/assets/js/videos/components/videoScreen/drawInteractions.vue
```diff
@@ -163,6 +163,19 @@ export default {
             this.$emit('pending-annotation', null);
         },
         extendPendingAnnotation(e) {
+            // Check polygons
+            if (e.feature.getGeometry().getType() === 'Polygon') {
+                if (PolygonValidator.isInvalidPolygon(e.feature)) {
+                    // Disallow polygons with less than three non-overlapping points
+                    this.$emit('is-invalid-polygon')
+                    this.removeFeature(e.feature)
```
mzur, line 171: This doesn't work. The `this.removeFeature` below was executed inside an event handler of `pendingAnnotationSource`, so the `this` referred to the source there.

I'll fix this myself while I'm doing my final review.

resources/assets/js/annotations/ol/PolygonValidator.js
```diff
@@ -0,0 +1,142 @@
+import JstsLinearRing from 'jsts/org/locationtech/jts/geom/LinearRing';
+import JstsPolygon from 'jsts/org/locationtech/jts/geom/Polygon';
+import LinearRing from '@biigle/ol/geom/LinearRing';
+import LineString from '@biigle/ol/geom/LineString';
+import Monkey from 'jsts/org/locationtech/jts/monkey'; // Needed for isValid(), normalize()
```
mzur, line 5: This can be converted to a [side effect import](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import?retiredLocale=de#forms_of_import_declarations). Then JSLint won't complain anymore.

resources/assets/js/annotations/ol/PolygonValidator.js
```diff
@@ -0,0 +1,148 @@
...
+    // Translate ol geometry into jsts geometry
+    let jstsPolygon = parser.read(feature.getGeometry());
+
+    if (jstsPolygon.isSimple()) {
+        return feature;
+    }
+
+    // Divide non-simple polygon at cross points into several smaller polygons
+    // and translate back to ol geometry
+    let polygons = jstsValidate(jstsPolygon)['array'].map(p => parser.write(p));
```
mzur, line 58: As a small performance optimization, maybe the greatest polygon can be found *before* converting everything back to OL geometry? Or do you think it's not worth it?

resources/assets/js/annotations/ol/PolygonValidator.js
```diff
@@ -0,0 +1,148 @@
...
+        if (geom.isValid()) {
+            geom.normalize(); // validate does not pick up rings in the wrong order - this will fix that
+            return geom; // If the polygon is valid just return it
+        }
+        var polygonizer = new Polygonizer();
+        jstsAddPolygon(geom, polygonizer);
+
+        return polygonizer.getPolygons();
+    } else {
+        return geom;
```
mzur, line 91: Can this happen at all? The polygon is converted to a jstsPolygon just before?

mzur (changes requested):
Overall it looks good! I have only one change request. I found that you can still create a non-simple polygon by *modifying* an existing polygon. So the cleanup code has to be added to this code path as well.

[Screencast from 07.11.2023 15:07:44.webm](https://github.com/biigle/core/assets/2457311/34461c01-3270-455a-a3f5-1d5aa8ad2931)

I've also experimented with conditional loading of the JSTS dependency so the main JS file does not get that big. It is quite simple (just call `await import('module')` and webpack will automatically generate an additional file that is loaded there). However, the code needs to be updated to work with the `async/await` call which is quite tedious. I think I defer a change like this for a later time when the main JS file gets even larger (maybe 5 MB or so). BIIGLE is a "power user" tool and we don't pay for the bandwidth so a 2 MB JS file won't hurt anyone for now.

resources/assets/js/annotations/ol/PolygonValidator.js
lehecht: Could you explain why this would be an performance optimization?

mzur: Well, probably it won't have a big effect but my reasoning was this: Here you convert all JSTS Polygons back to OL Polygons (with a "parser" who "writes" which sounds expensive to me :smile:). But right after that you throw away all but one of the converted polygons. So I was thinking that you could throw away the small polygons first before converting them.

lehecht: > maybe the greatest polygon can be found _before_ converting everything back to OL geometry?

Unfortunately this is not possible any more due to the bug mentioned above. There are some cases where polygons were separated and thereby just copied. Some copies still remained in the originals polygons coordinate set which needs to be removed. During the write() process, those coordinates are separated from that original list because they are recognized as different polygons. So I could remove them easily after calling write().
I also tried to catch and handle those case before transforming the polygons but this got really messy. The jsts polygon methods like covers(), contains(), overlaps(), etc... didn't work for that kind of polygons. Additionally I couldn't simply remove these duplicated coordinates because this could still lead to intersecting polygons and cause some other problems.

resources/assets/js/annotations/ol/PolygonValidator.js
```diff
@@ -53,9 +52,9 @@ export function makePolygonSimple(feature) {
         return feature;
     }
 
-    // Divide non-simple polygon at cross points into several smaller polygons
-    // and translate back to ol geometry
-    let polygons = jstsValidate(jstsPolygon)['array'].map(p => parser.write(p));
+    // Divide non-simple polygon at cross points into several smaller polygons,
+    // translate back to ol geometry and polish them
+    let polygons = polishPolygons(jstsValidate(jstsPolygon).array.map(p => parser.write(p)));
```
mzur, line 57: It's not immediately clear what "polishing" means. I'd at least expect the comment here to explain a little but it also just says "polish".

resources/assets/js/annotations/ol/PolygonValidator.js
```diff
@@ -134,6 +130,54 @@ function jstsAddLineString(lineString, polygonizer) {
...
 }
 
+/**
+ * Removes duplicated subpolygon's coordinates which can be still included in some polygons.
+ *
+ * @param polygons List of (multi)polygons
+ * @returns List of separated polygons
+ *
+ * **/
+function polishPolygons(polygons) {
```
mzur, line 140: I don't really understand this function. The comment says that duplicated coordinates should be removed. I understand this as an operation like `[[x1, y1],[x2,y2],[x2,y2]] -> [[x1, y1],[x2,y2]]`. Also, I think that the multi-polygon here is the list of polygons that you get when a non-simple polygon is cut into simple pieces. So why are all sub-polygons compared here? Shouldn't duplicate coordinates be removed independently for each sub-polygon? And as I said before, can't you determine the biggest sub-polygon first before performing all these operations (just asking, maybe you *really* can't)?

Since I struggled to understand this, maybe more elaborate comments should be added (for future readers). An example of the special case that this fixes would also be useful (since this function wasn't there before and it already looked like it would work fine).

resources/assets/js/annotations/ol/PolygonValidator.js
```diff
@@ -106,6 +101,7 @@ function jstsAddPolygon(polygon, polygonizer) {
     jstsAddLineString(polygon.getExteriorRing(), polygonizer);
 
     for (var n = polygon.getNumInteriorRing(); n > 0; n--) {
+        n = polygon.getNumInteriorRing() === 1 ? 0 : n;
```
mzur, line 104: Maybe add a short comment why this is required.

mzur (changes requested):
You can still create a non-simple polygon by modifying an existing one in the video annotation tool.

resources/assets/js/annotations/ol/PolygonValidator.js
mzur: Ah I missed your comment above. So determining the biggest polygon is not possible before converting back to OL features. But the questions about the new function are still valid.

mzur: Maybe the "polishing" can be done only on the biggest polygon?

lehecht: > I understand this as an operation like `[[x1, y1],[x2,y2],[x2,y2]] -> [[x1, y1],[x2,y2]]`

What I meant was something like this: (original) polygon1 = [[x1,y1], [x2,y2]] and and (subpolygon) polygon2 = [[x2,y2]]. 
The polygon was separated but the coordinates were not removed from the original polygon "polygon1".
This case can occur when using the magic wand tool and modifying that polygon.

lehecht: > Shouldn't duplicate coordinates be removed independently for each sub-polygon

Unfortunately, no. Like I described above, the duplicates occur between different polygons and not within one.

lehecht: > Maybe the "polishing" can be done only on the biggest polygon?

The polish method is used to create "clean" polygons whose coordinate sets occur only once. Therefore I need to compare the polygon with the most coordinate sets with the others to check if one is contained in another.

mzur: Ok thanks for the explanation! I think I'll tinker with this a bit once you added the comments about "polishing", so I better understand what's going on :wink:

lehecht: I don't know if it helps, but I recorded the bug which I fixed with the polish method. It doesn't happen all the time when modifying a magic wand polygon, but there are cases which you can see in the video.


https://github.com/biigle/core/assets/72259845/2af9349d-3ffd-4728-95b3-a45c6c139cc1

lehecht:
> Ok thanks for the explanation! I think I'll tinker with this a bit once you added the comments about "polishing" and added the validation step to modified video annotations, so I better understand what's going on 😉

I don't know if it helps, but I recorded the bug that I fixed with the polish method. It doesn't happen all the time when modifying a magic wand polygon, but there are cases.


https://github.com/biigle/core/assets/72259845/c3e8caa1-f8dc-41d7-8a69-813c84b99160

mzur (approved):
Ok I think this is good to go now! The original issue mentioned that the backend should reject non-simple polygons but I think we just leave this check in the frontend only.

PR #657, opened 2023-09-21
Fix bug with http code 200 and server error message

### LINKED ISSUE #641: Regression with attach label interaction in video annotation tool
We had a bug report that can be reproduced as follows:

1. Create a label tree with at least two labels (or use an existing one)
2. Create a whole-frame video annotation
3. Split the annotation in the middle
4. Use the attach label interaction to add the second label to one of the parts of the annotation
5. Delete the first label from the part of the video annotation

Now an error "the server didn't respond" occurs. However, none of the requests fail and the label is correctly deleted. Find out why the error is displayed and fix it.

This seems to be a regression with a recent update because the error did not occur in a local branch of mine. In the most recent master branch it did occur. Maybe the commit introducing this error can be found with `git bisect`.

lehecht:
Whole frame video annotations don't have points which leads to an empty array. RefreshSingleAnnotation() tried to access the array which caused an exception within the then-branch of the promise.

Resolves #641

### DIFF
--- a/resources/assets/js/videos/videoContainer.vue
+++ b/resources/assets/js/videos/videoContainer.vue
@@ -403,7 +403,12 @@ export default {
         detachAnnotationLabel(annotation, annotationLabel) {
             if (annotation.labels.length > 1) {
                 annotation.detachAnnotationLabel(annotationLabel)
-                    .then(() => this.refreshSingleAnnotation(annotation))
+                    .then(() => {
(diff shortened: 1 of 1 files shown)


mzur (approved):
Thanks! Unfortunately while testing this fix, I found another bug (https://github.com/biigle/core/issues/660). Sorry :grimacing:

PR #658, opened 2023-09-22
Add annotation count feature

### LINKED ISSUE #459: Show complete count of annotations
When annotating (at least for diatoms) you should aim for 400 annotations / slide. A sum of all annotations or all annotations/Slide on the top of the annotations tab would be very helpful.

Thanks in advance.

lehecht:
Display an annotation count in the annotations tab beneath the filter. If a filter is active, return an filtered annotation count and total annotation count.

Resolves #459

### DIFF
--- a/resources/assets/js/annotations/annotatorContainer.vue
+++ b/resources/assets/js/annotations/annotatorContainer.vue
@@ -160,6 +160,9 @@ export default {
         annotationsHiddenByFilter() {
             return this.annotations.length !== this.filteredAnnotations.length;
         },
+        getAnnotationCount() {
+            return this.annotations.length;
+        }
     },
     methods: {
         getImageAndAnnotationsPromises(id) {
--- a/resources/assets/js/annotations/annotatorContainer.vue
+++ b/resources/assets/js/annotations/annotatorContainer.vue
@@ -160,7 +160,7 @@ export default {
         annotationsHiddenByFilter() {
             return this.annotations.length !== this.filteredAnnotations.length;
         },
-        getAnnotationCount() {
+        annotationCount() {
             return this.annotations.length;
         }
     },
--- a/resources/assets/js/annotations/components/annotationsTab.vue
+++ b/resources/assets/js/annotations/components/annotationsTab.vue
@@ -83,12 +83,14 @@ export default {
                     };
                 });
         },
-        getTotalAnnotationCount() {
-            return this.totalAnnotationCount;
+        annotationCount() {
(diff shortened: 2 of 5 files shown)


resources/assets/js/annotations/annotatorContainer.vue
```diff
@@ -160,6 +160,9 @@ export default {
         annotationsHiddenByFilter() {
             return this.annotations.length !== this.filteredAnnotations.length;
         },
+        getAnnotationCount() {
```
mzur, line 163: Since this is a computed property, I would not give it a name that sounds like a method. You could call it `annotationCount`.

resources/assets/js/videos/videoContainer.vue
```diff
@@ -162,6 +162,9 @@ export default {
         annotationsHiddenByFilter() {
             return this.annotations.length !== this.filteredAnnotations.length;
         },
+        getAnnotationCount() {
```
mzur, line 165: Same than the comment above.

resources/assets/js/annotations/components/annotationsTab.vue
```diff
@@ -79,6 +83,12 @@ export default {
                     };
                 });
         },
+        getTotalAnnotationCount() {
```
mzur, line 86: Same than the comment above.

resources/assets/js/annotations/components/annotationsTab.vue
```diff
@@ -79,6 +83,12 @@ export default {
...
+            return this.totalAnnotationCount;
+        },
+        getFilteredAnnotationCountString() {
```
mzur, line 89: Same than the comment above.

mzur (commented):
Thanks! What do you think about a style like this?

image

This matches the `badge` counts below. We only need to make sure the users understand what the count means, so I added "Total". The text uses the `text-muted` class so it can be distinguished from the label names.

I also thought about putting the total count at the bttom of the list which could be more intuitive. But this would hide the count if the list gets too long.

If the filtering is active, the badge could still show "yy/xx".

lehecht:
Bildschirmfoto vom 2023-09-28 10-10-36

Are these two letters above the labels also displayed to you? Somehow they appear in every commit I checked out. Even in commits when I didn't start working here. I restarted the whole docker container but that didn't help.

mzur:
That's the name of your label tree :wink:

mzur (approved):
Perfect, thanks!

PR #652, opened 2023-09-15
Apply coding style

lehecht:
.php_cs is outdated and cannot be used anymore due to [PHP CS Fixer 3 upgrade](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/master/UPGRADE-v3.md). Instead use .php-cs-fixer.php and updated methods.

PR #677, opened 2023-10-11
Add github action for coding style check

### LINKED ISSUE #676: Add github action to check the code style
Add a new github action such that PHP-CS Fixer is run after pushing new commits.

lehecht:
Resolves #676 by adding cs-php action

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -52,3 +52,29 @@ jobs:
 
     - name: Run Psalm
       run: composer lint
+
+  cs-php:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Setup PHP
+      uses: shivammathur/setup-php@v2
+      with:
+        php-version: '8.1'
+
+    - uses: actions/checkout@v1
+      with:
+        fetch-depth: 1
+
+    - name: Install Dependencies
+      run: composer install --no-ansi --no-interaction --no-scripts --no-suggest --prefer-dist --ignore-platform-reqs
+
+    - name: Copy .env
+      run: cp .env.example .env
+
+    - name: Set testing key
+      run: echo "APP_KEY=base64:STZFA4bQKDjE2mlpRPmsJ/okG0eCh4RHd9BghtZeYmQ=" >> .env
+
+    - name: Run PHP-CS Fixer
+      run: composer fix -- --dry-run


mzur:
You have to use the `php-cs-fixer` branch as base for this PR, otherwise the `composer fix` command will fail. Also, you can use the `php-cs-fixer` branch as target for this PR, so this PR will not include all the changes of #652, too.

lehecht:
> use the `php-cs-fixer` branch as target for this PR

You mean as new base branch? So php-cs-fixer branch replaces the master branch?

mzur:
I meant like this (locally):

1. You check out the `php-cs-fixer` branch
2. You create a new branch from there
3. You apply the commit (you can `git cherry-pick` the existing commit from this branch)
4. You push the branch and create a new PR with the `php-cs-fixer` branch as target

But it won't matter any more because I'll merge #652 in a few minutes. Then you just have to merge the new `master` into `php-cs-action`.

mzur:
Now it's working. You can now try to add annotations.

lehecht:
Cs2pr throws an error, which I cannot reproduce. I'm on it, trying to fix this problem.
If you know something about it, @mzur, please let me know.

mzur:
When you use the compose command you have to do it like this:

```diff
- composer fix -vvv --dry-run --format=checkstyle | cs2pr
+ composer fix -- -vvv --dry-run --format=checkstyle | cs2pr
```

Otherwise the arguments will not be forwarded to the CS Fixer command.

composer.json
```diff
@@ -45,7 +45,8 @@
         "mockery/mockery": "^1.3.1",
         "nunomaduro/collision": "^7.0",
         "phpunit/phpunit": "^10.0",
-        "psalm/plugin-laravel": "^2.0"
+        "psalm/plugin-laravel": "^2.0",
+        "staabm/annotate-pull-request-from-checkstyle": "^1.8"
```
mzur, line 49: I think you don't need to add the dependency here because it is already installed in the `setup-php` action.

lehecht:
Thank you!

mzur (approved):
It works, nice :+1:

You can put adding this action to all the modules on your agenda. When the action is added, all CS issues in the module can be resolved, too. But as I said, you don't have to do this right now so you can work on some more interesting things in the meantime.

PR #700, opened 2023-11-17
Implement alternative handling of polygons with holes

mzur:
I'd like to propose an updated solution to handling non-simple polygons that does not require the "polish" function. Here is my reasoning behind this:

The polishing function was used to remove duplicate coordinates that were sometimes added. I found that these were probably "holes", because every time a polygon had such a duplicate coordinate, there was an entry in `_holes` when the polygon was logged with `polygonizer.getGeometry()` in `jstsValidate()` (now `jstsSimplify()`). The screenshot shows the geometry with a hole at the top and the OL coordinates at the bottom:

image

So in order to remove the holes, now `getExteriorRing()` is called for each polygon. To reproduce a polygon with a hole, draw a geometry like this:

[Screencast from 17.11.2023 11:09:13.webm](https://github.com/biigle/core/assets/2457311/ce302064-6a24-40ee-b461-8e6a65d85aad)

For a similar reason I removed the "getInteriorRing" stuff from `jstsAddPolygon()`. However, holes were still present in the polygons and had to be removed as described above. Everything still worked fine without the code, though, so I left it like that.

Finally, `getGreatestPolygon()` also works if there is only one polygon so I removed the `if`.

@lehecht Could you please have a good look at this and check if this still handles all the cases that you tested? You spent much more time with this than I did so you probably know all the edge cases. It's entirely possible that my implementation does not handle some cases that your version handles correctly.

References #634

### DIFF
--- a/resources/assets/js/annotations/ol/PolygonValidator.js
+++ b/resources/assets/js/annotations/ol/PolygonValidator.js
@@ -52,16 +52,11 @@ export function simplifyPolygon(feature) {
         return feature;
     }
 
-    // Divide non-simple polygon at cross points into several smaller polygons,
-    // translate back to ol geometry and remove duplicated coordinate sets (polish)
-    let polygons = polishPolygons(jstsValidate(jstsPolygon).array.map(p => parser.write(p)));
-
-    if (polygons.length > 1) {
-        // Select biggest part
-        let greatestPolygon = getGreatestPolygon(polygons);
-        // Only change coordinates because object references are in use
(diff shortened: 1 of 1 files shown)


lehecht:
This works just perfectly. Thanks! I could not find any edge case where your solution does not work, so we can use that.

lehecht (approved):
This works just perfectly. Thanks! I could not find any edge case where your solution does not work, so we can use that.

mzur:
Thanks for the review!

PR #709, opened 2023-11-29
Create volume flow v2

### LINKED ISSUE #701: Create volume flow v2
This issue is about implementing a new flow to create volumes. In addition to what the current flow supports, it will support:

- Importing a volume from iFDO v2 and the current CSV format
- Defining additional metadata importers https://github.com/biigle/core/issues/555
- Importing annotations (from a metadata file)
- Importing image labels (from a metadata file) https://github.com/biigle/core/issues/317

The old create volume flow (v1) will still be offered as API endpoint for backwards compatibility. Only the metadata (file/text) fields might be removed from that endpoint.

Here is the concept: The v2 flow consists of multiple stages.

1. The user chooses a media type (image or video) and optionally selects a metadata file for upload. Advancing to the next step, the metadata file is uploaded and stored in a "pending volume". Relevant information from the file is extracted for the following steps (by the metadata parser https://github.com/biigle/core/issues/555).

2. In this step, the volume name is set (automatically filled if provided in the metadata but can be edited by the user). Also, the user selects a file source (remote, uploaded files, storage disk). There are different cases now (depending on the source and if a metadata file was supplied):

   - remote/no metadata: The user enters the URL and pastes a list of comma separated filenames (like in the flow v1). ~~When the filenames are pasted, a list is generated showing the filenames in separate rows. Each file can be removed from the list with an "x" button.~~

   - remote/metadata: If available, the URL is prefilled from the metadata. Otherwise the user has to enter it. The filenames are prefilled (i.e. the filename list is displayed) from the metadata. ~~The user can choose to ignore some files by removing them from the list (see above) or they can clear all files and enter a new comma-separated list of filenames.~~

   - uploaded files/no metadata: The user can select a directory or a list of files from the file browser. ~~This will show a list of filenames (with the option to remove individual files) as described above.~~

   - uploaded files/metadata: ~~The uploaded files are searched for a directory that contains all filenames from the metadata. This directory (or the subset of files) will be automatically selected. Otherwise there will be a warning that no suitable files could be found.~~ Same than uploaded files/no metadata but there will be a warning if the selected files do not match the files from the metadata.

   - storage disk/no metadata: Same than uploaded files/no metadata.

   - storage disk/metadata: Same than uploaded files/no metadata but there will be a warning if the selected files do not match the files from the metadata.

   If the metadata file includes the relevant information, the user can also select a checkbox to import annotations and/or a checkbox to import file labels in this step. Also, they can specify a handle or DOI for the volume (which is prefilled if the metadata file includes it).

   The new volume will be created after this step. If annotations or file labels should be imported, the PendingVolume is kept and associated with the new volume ("Create volume and import annotations" or "Create volume and import image labels" button). Then there are the following additional steps. Otherwise, the PendingVolume is removed ("Create volume" button).

3. This step allows the user to specify which annotations should be imported (i.e. show available labels and allow the user to import only some of them).

4. This step allows the user to specify which image labels should be imported (i.e. show available labels and allow the user to import only some of them).

5. This step checks all labels that should be imported based on the selections in the previous two steps. If all labels to import already exist in the database (matched by their UUID), this step is automatically skipped. Otherwise, the user has the option to:
   
   - Select substitute labels from an existing label tree.

   - Add imported labels to an existing label tree.

   - Create a new label tree with the imported labels.

6. This step checks if all users that belong to the annotations/file labels to import already exists in the database (matched by their UUID). If not, the user who currently creates this volume has the option to:

   - Select a user from the database for each user of the metadata that was not found

   - Select themselves as the creator of the annotations/image labels

   The new annotations and/or file labels will be created after this step. The existence of the labels chosen in step 5 has to be checked again. If some are missing, go back to that step.

When this process is complete, the new volume, files, annotations, file labels and labels are created. The "pending volume" is removed. The pending volume is also removed if this process is aborted or not finished within a week (or so).

When a pending volume exists, it is shown to the user who created it if they want to create another new volume for the same project. They can continue to create the pending volume or discard it (by creating a new one). They can not have multiple pending volumes for the same project.

mzur:
Resolves #701

Plan:

- [x] Create a new PendingVolume model
- [x] Implement a CSV metadata parser that can recognize the CSV file and return a filled VolumeMetadata object.
- [x] Create a validation rule for a VolumeMetadata object based on the existing Image/VideoMetadata rules.
- [x] Implement the CSV parser and validation rule for videos, too
- [x] Create an API endpoint to create a PendingVolume for a project (as a project admin). A new PendingVolume is associated to a media type, project and can optionally have an uploaded metadata file. Each user is allowed to create only one PendingVolume at a time for each project. When a metadata file is uploaded, the metadata file type is checked based on the configured metadata parsers. If one parser recognizes the file, it parses it and the parsed metadata object is validated.
- [x] Update the metadata handling of the legacy API endpoint to use the new mechanisms
- [x] Create an API endpoint to update a PendingVolume. The update triggers the creation of the new volume. Then, either the PendingVolume is deleted or it is actually updated (i.e. associated with the volume) so the annotation/file label import can proceed (see https://github.com/biigle/core/issues/701#issue-2000484824).
   - [x] Implement the case without annotation/file label import
   - [x] Implement the case with annotation/file label import
- [x] Update CreateNewImagesOrVideos
   - Use metadata from the file associated with the volume
   - Ignore files that are not present in the metadata, ignore metadata that is not present in the files
- [x] Update the VolumeMetadataController request to use the new metadata parsers (making no distinction between CSV and iFDO)
- [x] Update CloneImagesOrVideos
- [x] Update the clone volume feature for the new volume metadata
- [x] Update the metadata file upload UI in the edit volume view
- [x] Implement the ImportVolumeMetadata job
- [x] Implement the new create volume UI, remove the old one
   - [x] Add a way to continue or delete an existing pending volume for the project
   - [x] Implement the UI without the import features
   - [x] Implement the UI with pre-filled data from the metadata file (https://github.com/biigle/core/issues/701#issue-2000484824)
   - [x] Implement a warning if none of the selected files are included in the metadata
   - [x] Implement the UI for the import features
- [x] Ignore invalid label colors and label/user UUIDs during import
- [x] Implement a way to extend available metadata parsers through modules/packages.
    - [x] These should also provide ~~file extensions~~ MIME types for validation of the form request as well as the `accept` attribute of the HTML input elements.
- [x] Remove iFDOv1 support
   - Remove iFDOv1 files associated with volumes manually (if there are any)
- [x] Remove the `ParsesMetadata` trait. Update everything with "metadata" or "Ifdo" in it.
- [x] Implement generic metadata download button in volume overview (instead of iFDO button)
- [x] Keep algorithm to merge video metadata based on multiple files (or change it to override metadata?)
- [x] Update the manual about metadata
    - [x] Make sure to explain the new metadata merge behavior for videos (see UpdateVolumemetadata job)
    - [x] Also make metadata parser modules extend the manual
- [x] Implement a `metadata_parser` column for volumes that specifies which metadata parser class should be used for the metadata file
- [x] Remove the yaml PHP extension (used by biigle/reports)
- [x] Implement an iFDOv2 export (incl. annotations) in biigle/reports (remove v1 support) (https://github.com/biigle/reports/pull/117)
    - [x] Export the non-standard label ~~and label tree UUIDs~~ for each label (for matching during import) and the label color
    - [x] ~~Export the non-standard user UUID for each annotator (for matching during import)~~ already done with the `id` property.
- [x] Test the pre-filled volume form and the import with an iFDOv2
- [x] Test the iFDOv2 import label and user matching with an iFDO report

----

**Linked PRs**

- https://github.com/biigle/reports/pull/117
- https://github.com/biigle/metadata-ifdo/pull/3

----
**Upgrade instructions**

- Run migrations
- Delete iFDO storage disk (with contents)
- Configure `VOLUME_METADATA_STORAGE_DISK` and `VOLUME_PENDING_METADATA_STORAGE_DISK` (with config)


mzur:
Here are the tests that an iFDOv2 metadata parser should fulfil:

https://github.com/biigle/core/blob/91ac92b46d996a043cfc592b7503733b844a8f4f/tests/php/Traits/ParsesMetadataTest.php#L102-L488

PR #716, opened 2023-12-07
Adapt core templates (master) to work with the geo module

mtiessen1175:
This PR contains all changes of #713, and points at correct target-branch (see [explanation](https://github.com/biigle/core/pull/713#issuecomment-1845263565)).

Changes so far:

- [x] create fileCount-method for filterListComponent.vue
- [x] add an array-equality check in hasRule()-method of filterTab.vue

### DIFF
--- a/resources/assets/js/volumes/components/filterListComponent.vue
+++ b/resources/assets/js/volumes/components/filterListComponent.vue
@@ -18,16 +18,31 @@ export default {
             type: Object,
             required: true,
         },
+        type: {
+            type: String,
+            required: true,
+        },
     },
     data() {
         return {
             name: this.rule.id,
         };
     },
+    methods: {
+        createFileCount(rule) {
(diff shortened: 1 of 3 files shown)


resources/assets/js/volumes/components/filterListComponent.vue
```diff
@@ -2,7 +2,7 @@
     <span>
         <strong>with<span v-if="rule.negate">out</span></strong>
         <span v-text="name"></span>
-        <strong v-if="dataName" v-text="dataName"></strong>
+        <strong v-if="dataSpecs" v-text="dataSpecs"></strong>
```
mzur, line 5: IMO the image count should not be bold, too. Instead it could even be muted:
```suggestion
        <strong v-if="dataName" v-text="dataName"></strong>
        <span class="text-muted v-text="fileCount"></span>
```

mzur (changes requested):

mzur (approved):
Thanks! I'll release this in a couple of minutes.

Issue #462, opened 2022-06-20
Metadata overlay in annotation tools


mzur:
Image and video metadata could be displayed as an overlay in the annotation tool (similar to the scale or zoom factor). Video metadata could be interpolated (if possible). Put all the metadata in a single overlay but let the user decide which exact metadata fields should be shown.

One use case, particularly for video metadata, is to ignore sections of the video with a specific camera altitude.

mzur:
An additional idea for the video annotation tool:

Similar to the [video thumbnails](https://github.com/biigle/core/issues/375) that could be displayed while "scrubbing" the timeline, metadata such as the absolute timestamp or the current depth could be displayed in a tooltip as well.

mzur:
When area information is available for videos, we can add support for the scale bar in screenshots https://github.com/biigle/core/pull/1455

mzur:
We should also allow more/arbitrary metadata fields. Allowed are string or numeric or date fields. These are auto-detected. Numeric and date fields can be interpolated. There is a limit of the maximum allowed number of metadata fields (10?) and the size of each field (256 chars?).

PR #725, opened 2023-12-18
Add exiftool dependency for geo-module 

mtiessen1175:
Since the native PHP-exiftool functions do not support all file-formats (see [here](https://www.php.net/manual/en/function.exif-read-data.php)), the use of exiftool would be helpful for extracting more specific metadata, like [GeoTIFF-tags](https://exiftool.org/TagNames/GeoTiff.html).

Exiftool might also be helpful for metadata-extraction from videos #462. 
  
Tasks:
- [ ] Adding the exiftool dependency to the app & worker docker-containers.

### DIFF
--- a/.docker/worker.dockerfile
+++ b/.docker/worker.dockerfile
@@ -37,6 +37,8 @@ RUN apk add --no-cache \
         zip \
     && apk del --purge .build-deps
 
+RUN apk add --no-cache exiftool
+
 # Configure proxy if there is any. See: https://stackoverflow.com/a/2266500/1796523
 RUN [ -z "$HTTP_PROXY" ] || pear config-set http_proxy $HTTP_PROXY
 RUN apk add --no-cache yaml \
--- a/composer.json
+++ b/composer.json
@@ -32,6 +32,8 @@
         "laravel/framework": "^10.0",
         "laravel/tinker": "^2.8",
         "laravel/ui": "^4.0",
+        "league/flysystem-aws-s3-v3": "^3.12",
+        "lychee-org/php-exif": "^1.0",
         "msurguy/honeypot": "^1.0",
         "pgvector/pgvector": "^0.1.4",
         "php-ffmpeg/php-ffmpeg": "^1.0",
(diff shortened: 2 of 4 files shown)


composer.json
```diff
@@ -32,6 +32,8 @@
         "laravel/framework": "^10.0",
         "laravel/tinker": "^2.8",
         "laravel/ui": "^4.0",
+        "league/flysystem-aws-s3-v3": "^3.12",
```
mzur, line 35: This doesn't belong here.

mzur (changes requested):

mzur:
This is related to https://github.com/biigle/geo/issues/44.

mzur (dismissed):
Looks good now :+1:  This will be merged together with the biigle/geo pull request that actually implements the GeoTIFF feature. Please be sure to link this PR there, too.

mzur (commented):
Why do you need `proj4php` here? Can't you add it to the composer.json of biigle/geo?

mtiessen1175:
> Why do you need `proj4php` here? Can't you add it to the composer.json of biigle/geo?

Yes, you're right it will only be used in the geo-module. I did not think about this properly when I installed it into core previously.

mzur (approved):

.docker/worker.dockerfile
```diff
@@ -37,6 +37,8 @@ RUN apk add --no-cache \
         zip \
     && apk del --purge .build-deps
 
+RUN apk add --no-cache exiftool
```
mzur, line 40: The worker image is now based on Debian so you have to use `apt-get`.

mzur (changes requested):

.docker/worker.dockerfile
mtiessen1175: resolved 3577945dd93a1a6e008ae38b899f284479e024fa

mzur (dismissed):

mzur (changes requested):
Please remove the storage disk configuration again, as the module is not installed by default. You can open a PR with the disk configuration to the `dev-modules` branch once this is merged.

mtiessen1175:
> Please remove the storage disk configuration again, as the module is not installed by default. You can open a PR with the disk configuration to the `dev-modules` branch once this is merged.

That makes sense! I removed it.

mzur (dismissed):

PR #726, opened 2023-12-22
Fix video timeline reset bug

### LINKED ISSUE #724: Sometimes the video position is not reset
Sometimes when I move the video position to a certain time and then switch to the next or previous video in the volume, the position is not reset to 0. Here is an example (wait for it):

[Screencast from 15.12.2023 10:43:38.webm](https://github.com/biigle/core/assets/2457311/542db8df-d81c-44e0-9f94-459759878fb9)

lehecht:
Replace seeked event by timeupdate, because seeked event does not always reset time after switching videos.

Fix #724

### DIFF
--- a/resources/assets/js/videos/components/videoScreen/videoPlayback.vue
+++ b/resources/assets/js/videos/components/videoScreen/videoPlayback.vue
@@ -118,7 +118,7 @@ export default {
                 this.startRenderLoop();
             }
         },
-        setPaused(dontSeek) {
+        setPaused(dontSeek = false) {
             this.playing = false;
             this.stopRenderLoop();
             // Force render the video frame that belongs to currentTime. This is a
--- a/resources/assets/js/videos/components/videoScreen/videoPlayback.vue
+++ b/resources/assets/js/videos/components/videoScreen/videoPlayback.vue
@@ -118,7 +118,7 @@ export default {
                 this.startRenderLoop();
             }
         },
-        setPaused() {
+        setPaused(dontSeek) {
             this.playing = false;
             this.stopRenderLoop();
             // Force render the video frame that belongs to currentTime. This is a
@@ -126,7 +126,9 @@ export default {
             // currentTime (in most cases). With the workaround we can create annotations
             // at currentTime and be sure that the same frame can be reproduced later for
             // the annotations. See: https://github.com/biigle/core/issues/433
-            this.$emit('seek', this.video.currentTime, true);
+            if (!dontSeek) {
+                this.$emit('seek', this.video.currentTime, true);
+            }
         },
         togglePlaying() {
             if (this.playing) {
--- a/resources/assets/js/videos/components/videoScreen/videoPlayback.vue
+++ b/resources/assets/js/videos/components/videoScreen/videoPlayback.vue
@@ -167,7 +167,7 @@ export default {
         this.dummyCanvas.height = 1;
         this.video.addEventListener('play', this.setPlaying);
         this.video.addEventListener('pause', this.setPaused);
-        this.video.addEventListener('timeupdate', this.handleSeeked);
+        this.video.addEventListener('seeked', this.handleSeeked);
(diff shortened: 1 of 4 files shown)


resources/assets/js/videos/components/videoScreen/videoPlayback.vue
```diff
@@ -167,7 +167,7 @@ export default {
         this.dummyCanvas.height = 1;
         this.video.addEventListener('play', this.setPlaying);
         this.video.addEventListener('pause', this.setPaused);
-        this.video.addEventListener('seeked', this.handleSeeked);
+        this.video.addEventListener('timeupdate', this.handleSeeked);
```
mzur, line 170: This should still be `seeked`. The `timeupdate` event also fires while the video is playing so there will be redundand calls to `renderVideo()` here.

resources/assets/js/videos/videoContainer.vue
```diff
@@ -189,7 +189,7 @@ export default {
             }
 
             let promise = new Vue.Promise((resolve, reject) => {
-                this.video.addEventListener('seeked', resolve);
+                this.video.addEventListener('timeupdate', resolve);
```
mzur, line 192: This should also still be `seeked` because it resolves the promise returned by the `seek()` method.

resources/assets/js/videos/videoContainer.vue
```diff
@@ -676,9 +676,9 @@ export default {
                 Messages.danger('Error while loading video file.');
             }
         });
-        this.video.addEventListener('seeked', this.handleVideoSeeked);
+        this.video.addEventListener('timeupdate', this.handleVideoSeeked);
```
mzur, line 679: This should still be `seeked`.

resources/assets/js/videos/videoContainer.vue
```diff
@@ -676,9 +676,9 @@ export default {
...
         this.video.addEventListener('pause', this.updateVideoUrlParams);
-        this.video.addEventListener('seeked', this.updateVideoUrlParams);
+        this.video.addEventListener('timeupdate', this.updateVideoUrlParams);
```
mzur, line 681: This should still be seeked (because the URL is updated on pause or `seeked` but not while the video plays).

mzur (changes requested):
While the issue is apparently fixed by swapping `seeked` with `timeupdate` in the videoTimeline, we don't need to swap it in the other places. The `timeupdate` event fires continuously while the video is playing and this is not what is needed in the other places.

Also it's not quite ideal in the videoTimeline as well. The timeline starts an update loop when the video plays to update the current time of the video. When the video is seeked, the event is used to synchronize the time because the update loop may be a few ms off. However, when `seeked` is replaced with `timeupdate`, the update loop updates the time and the `timeupdate` event *also* updates the time while the video plays. This is redundant and makes the movement of the red current time indicator a little jerky.

I think when I implemented the video annotation tool, I first wanted to use only `timeupdate` to update the current time in the timeline but the event fires not frequently enough. So I disabled `timeupdate` again and implemented the update loop instead.

Maybe we have to dig a little deeper here to find out why the current time in the timeline is not reset so a better fix can be implemented.

mzur (commented):
I wondered why this change fixed the issue at all. When the next video is displayed, `reset()` of the videoContainer is called, which in turn calls `reset()` of the videoTimeline. There, the internal `currentTime` is set to 0, too, so why does setting `video.currentTime` to 0 change anything?

It turned out that the videoScreen in the culprit because its `reset()` method is called too. In there, `setPaused()` is called, which will emit a `seek` event with the current time (as a workaround for a bug, so the correct frame will be displayed). This event is sometimes handled with a delay, so the *new* video will be seeked to that time.

So a possible solution could be to add a `dontSeek` argument to `setPaused()` and if the argument is `true`, the seek event is not emitted. In `reset()` of the videoScreen, the call can then be changed to `this.setPaused(true)`.

mzur (approved):
Thanks!

PR #727, opened 2023-12-22
Update annotations by using old features

### LINKED ISSUE #660: Error after splitting video annotation with new label
I found the following error in the video annotation tool (coming from [here](https://github.com/biigle/core/blob/8f9c0864c54bb7ee57d2a910f0ce7ba1bccafbe1/resources/assets/js/videos/components/videoScreen/annotationPlayback.vue#L96)):

```
Error in event handler for "refresh": "TypeError: this.featureChangeKeys_[featureKey] is undefined"
```

Steps to reproduce:

1. Create a label tree with at least two labels (or use an existing one)
2. Create a point video annotation
3. Split the annotation in the middle
4. Use the attach label interaction to add the second label the first part of the annotation
5. Delete the first label from the part of the video annotation
6. Move the current time of the video to be inside the first part of the annotation
7. Play the video

The error occurs when the current time of the video jumps from the first part of the annotation to the second.

lehecht:
Fix #660

### DIFF
--- a/resources/assets/js/videos/components/videoScreen/annotationPlayback.vue
+++ b/resources/assets/js/videos/components/videoScreen/annotationPlayback.vue
@@ -63,7 +63,6 @@ export default {
             // a seek was performed to the exact annotation time. Hence, we  round the
             // time here to the maximum of 4 decimals, too.
             time = rtp(time);
-
             for (let i = 0, length = annotations.length; i < length; i++) {
                 // We can skip ahead and break early because of the sorting in the
                 // annotationsPreparedToRender array.
@@ -90,13 +89,13 @@ export default {
                     toCreate.push(annotation.self);
                 }
             }
-            
             if (hasRenderedFeatures) {
                 Object.values(oldRendered).forEach(function (feature) {
-                    // source.hasFeature(feature) does not work here, because it is always true
-                    if (source.getFeatures().includes(feature)) {
-                        source.removeFeature(feature);
-                    }
+                    // The annotation source uses featureChangeKeys method with featureKey as parameter.
+                    // Feature and oldFeature do not have the same key, 
+                    // which is why featureChangeKeys returns undefined in removeFeature(feature).
+                    let oldFeature = source.getFeatureById(feature.getId());
+                    source.removeFeature(oldFeature);
                     selected.remove(feature);
                 });
             } else {
@@ -110,8 +109,7 @@ export default {
                     selected.clear();
                 }
             }
-
-
+            
             let features = toCreate.map(this.createFeature);
             features.forEach(function (feature) {
                 newRendered[feature.getId()] = feature;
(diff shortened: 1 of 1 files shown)


mzur:
Can you please elaborate why this fixes the original issue (as the reported error happens somewhere else)?

lehecht:
I fixed a very similar error [here](https://github.com/biigle/core/blob/c640be10f94177962385d578d214278422ac5f12/resources/assets/js/annotations/components/annotationCanvas.vue#L523-L528).
The error occurs, because you try to delete an annotation, which does not exist any more in the annotation source. That's why you need to check for its presence. Some annotations are still present and can be deleted there, but others are already deleted. Moreover the method `hasFeature` in `source.hasFeature(feature)` cannot be used, because it still uses the old array of the source, which is why I needed to use `getFeatures().includes()`.

resources/assets/js/videos/components/videoScreen/annotationPlayback.vue
```diff
@@ -90,10 +90,13 @@ export default {
                     toCreate.push(annotation.self);
                 }
             }
-
+
```
mzur, line 93:
```suggestion

```

mzur (commented):
Ok thanks. I can confirm that the error is gone with the change. However, I see a visual bug where suddenly a second point appears on the video when the annotation label is deleted. Here is a recording (don't mind the "invisible video" bug of Firefox):

[Screencast from 04.01.2024 09:39:57.webm](https://github.com/biigle/core/assets/2457311/ae0335cc-a242-468e-b6be-8c121c586763)

The point is gone once the video played past the annotation(s). It also does not appear again if I rewind the video to the annotation.

lehecht:
A little more explanation here.
My previous solution fixed the error only if the first annotation was selected. So the bug still existed when no annotation was selected.
I found out, that there were some problems with new created features and still existing one in the annotation source. The annotation source uses the featureChangeKeys method. For features not coming from the source directly the return value is always undefined, because they do not have the same keys like their predecessor(-copies).

resources/assets/js/videos/components/videoScreen/annotationPlayback.vue
```diff
@@ -135,7 +136,13 @@ export default {
             source.addFeature(newFeature);
         },
         createFeature(annotation) {
```
mzur, line 138: In which case is `createFeature()` called with an annotation that already exists as a feature? I feel that the method should not be called in the first place in such a case.

resources/assets/js/videos/components/videoScreen/annotationPlayback.vue
```diff
@@ -128,11 +128,12 @@ export default {
...
-            let oldFeature = source.getFeatureById(annotation.id)
-            
-            source.removeFeature(oldFeature);
-            source.addFeature(newFeature);
+            let feature = source.getFeatureById(annotation.id)
+
+            source.removeFeature(feature);
+
+            feature.set('color', annotation.labels[0].label.color);
+            source.addFeature(feature);
```
mzur, line 136: Looks good! Is the call to `removeFeature()` and `addFeature()` even necessary now?

mzur (approved):
Thanks!

PR #728, opened 2023-12-22
Remove null (gap) to prevent validation errors

### LINKED ISSUE #488: Bug when deleting single-frame/multi-frame annotation with gap
An error occurs when the following steps are taken:

1. Create a multi-frame video annotation with two keyframes
2. Create a single-frame video annotation with a temporal gap to the first annotation
3. Link the multi-frame annotation to the single-frame annotation
4. Delete the keyframe of the single-frame annotation

The error occurs, the anotation is displayed wrongly and you have to repeatedly select different keyframes to get rid of the annotation. This should be fixed.

lehecht:
If single frame and clip annotations are linked and are not touching each other, then null is added. When deleting the single frame annotation, delete also the null.

Fix #488

### DIFF
--- a/app/Traits/HasPointsAttribute.php
+++ b/app/Traits/HasPointsAttribute.php
@@ -15,11 +15,6 @@ trait HasPointsAttribute
      */
     public function validatePoints(array $points)
     {
-        // Gaps are represented as empty arrays
-        if (empty($points)) {
-            return;
-        }
-
         // check if all elements are integer
         $valid = array_reduce($points, fn ($carry, $point) => $carry && (is_float($point) || is_int($point)), true);
 
--- a/app/Traits/HasPointsAttribute.php
+++ b/app/Traits/HasPointsAttribute.php
@@ -15,6 +15,11 @@ trait HasPointsAttribute
      */
     public function validatePoints(array $points)
     {
+        // Gaps are represented as empty arrays
+        if (empty($points)) {
+            return;
+        }
+
         // check if all elements are integer
         $valid = array_reduce($points, fn ($carry, $point) => $carry && (is_float($point) || is_int($point)), true);
 
--- a/resources/assets/js/videos/models/Annotation.vue
+++ b/resources/assets/js/videos/models/Annotation.vue
@@ -283,12 +283,14 @@ export default Vue.extend({
                 this.points.splice(index, 1);
 
                 // Remove null (gap filler) as first/last element to prevent validation errors
-                if (this.frames[0] === null) {
+                // Multiple consecutive 'null' can occur after removing elements of linked anntations
+                // in specific order
+                while (this.frames[0] === null) {
                     this.frames.shift();
                     this.points.shift();
                 }
 
-                if (this.frames.at(-1) === null) {
+                while (this.frames.at(-1) === null) {
                     this.frames.pop();
                     this.points.pop();
                 }
--- a/resources/assets/js/videos/models/Annotation.vue
+++ b/resources/assets/js/videos/models/Annotation.vue
@@ -282,7 +282,12 @@ export default Vue.extend({
                 this.frames.splice(index, 1);
                 this.points.splice(index, 1);
 
-                // Remove null (gap filler) as last element to prevent validation errors
+                // Remove null (gap filler) as first/last element to prevent validation errors
+                if (this.frames[0] === null) {
+                    this.frames.shift();
+                    this.points.shift();
+                }
+
                 if (this.frames.at(-1) === null) {
                     this.frames.pop();
                     this.points.pop();
--- a/resources/assets/js/videos/models/Annotation.vue
+++ b/resources/assets/js/videos/models/Annotation.vue
@@ -282,15 +282,14 @@ export default Vue.extend({
                 this.frames.splice(index, 1);
                 this.points.splice(index, 1);
 
-                // Remove null (gap filler) as first/last element to prevent validation errors
-                if (this.frames[0] === null) {
-                    this.frames.shift();
-                    this.points.shift();
-                }
-
-                if (this.frames.at(-1) === null) {
-                    this.frames.pop();
-                    this.points.pop();
+                // Remove "null" elements of adjacent gaps to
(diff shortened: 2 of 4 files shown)


mzur:
There is still an error if I put the single-frame annotation before the multi-frame annotation.

lehecht:
I found a related bug, when you delete an annotation within a multiframe annotation, which is bordered by single frame annotation on each side. You will get the error message `Invalid number of points for shape Point!`.

mzur (approved):
This works, thanks!

mzur:
> If you delete an annotation within a multiframe annotation, which is bordered by single frame annotations on each side, you will get the error message Invalid number of points for shape Point!.

Could you please also fix this in this PR?

app/Traits/HasPointsAttribute.php
```diff
@@ -15,6 +15,11 @@ trait HasPointsAttribute
      */
     public function validatePoints(array $points)
     {
+        // Gaps are represented as empty arrays
+        if (empty($points)) {
+            return;
+        }
+
```
mzur, line 22: This case is only relevant for video annotations so I would rather change `validatePoints()` of the VideoAnnotation model. It should only call `validatePoints()` of the trait with non-empty points arrays but still make sure that the gaps make sense (i.e. there must be filled points arrays before and after the gap).

resources/assets/js/videos/models/Annotation.vue
```diff
@@ -283,12 +283,14 @@ export default Vue.extend({
...
+                while (this.frames[0] === null) {
                     this.frames.shift();
                     this.points.shift();
                 }
 
-                if (this.frames.at(-1) === null) {
+                while (this.frames.at(-1) === null) {
                     this.frames.pop();
                     this.points.pop();
                 }
```
mzur, line 296: Why is this also needed? As I understand it, the new bug you found is solved by updating `validatePoints()` but why this, too? Maybe you can elaborate in the comment (what "specific order"?).

lehecht: This was another bug, where multiple consecutive nulls can occur, if you delete 
1. first/last annotation of an multiframe annotation
2. first/last annotation of an single frame annotation
After removing the first, a null is added in place of the removed element, which leads to two nulls at the array start/end.

mzur: Sorry but I still don't get it. Could you please tell me exactly what I have to do to reproduce this?

lehecht: Sorry, this is my mistake. I mixed up the order. You can see in the video, if you first delete the multi frame annotation frame by frame and then a linked single frame annotation, an error is thrown due to the fact, that there are two nulls at the start. The if() statement removes null only once, which is why the second null remains in the array.

[Bildschirmaufzeichnung vom 12.01.2024, 09:48:40.webm](https://github.com/biigle/core/assets/72259845/d01adcdb-7494-4252-9648-d21e9716963b)

lehecht: Unfortunately no. You can go on and add more single frame annotations, link and delete them such that more nulls are added.

mzur: But when each frame is deleted, the logic above would also delete one `null`, right? So the `null`'s should not accumulate.

lehecht: Here you can see, how I ended up with three nulls. 

[Bildschirmaufzeichnung vom 15.01.2024, 07:40:32.webm](https://github.com/biigle/core/assets/72259845/69ab55a2-676a-4491-b654-402ff9a86a1a)

A null is only deleted if it is the first or last element. If you accumulate them in the middle, they are not deleted, which is why I used a while loop. If you don't like while-loops, I could identify start and end index of all nulls and delete those areas.

mzur: I'm still convinced by the logic above :smile: Each time you delete a keyframe, a `null` would be deleted, too.

Your (current) while loop only trims the `null`s at the start and end, not in the middle (as demonstrated in your video). With the logic above I think it would work in all cases and no loop is needed any more.

lehecht: In my videos I use the code with the if-statements. Maybe I should have said this before :sweat_smile:

mzur: Interesting. Can you please push it? Then I'll have a look myself.

lehecht:
@mzur you can check it out now.

mzur:
I have pushed the implementation of my idea and I still think it works. Can you please test it, too?

lehecht:
It works. And now I understood what you meant. I was really confused before. Thanks!

mzur:
Sorry, I'll try to be more clear next time :wink: Do we still need the [other change](https://github.com/biigle/core/pull/728#discussion_r1442704333)? If yes, please move it to the VideoAnnotation model.

mzur (commented):
Thanks! I'm sorry to drag this on even more but could you now please add a test for the new validation behavior? The test should be written in a way that would fail before the change and pass with the change.

lehecht:
> The test should be written in a way that would fail before the change and pass with the change.

What do you mean by that?

mzur:
I would like to have a test like [this one](https://github.com/biigle/core/blob/8669b0b598361c939d97c3c3d2f687270293571a/tests/php/VideoAnnotationTest.php#L52-L61) that checks the new behavior of validating an annotation with a gap. If the test would be executed *before* the changes of this PR, it should fail. *After* the changes of this PR, the test should pass.

lehecht:
For me it sounds like, you want something like this. That is why I'm still confused.
```
$this->model->shape_id = Shape::pointId();
$this->model->points = [[]];
$this->model->validatePoints(); // does not fail
$this->expectException(Exception::class);
$this->model->points = [[]];
$this->model->validatePoints(); // should fail, but does not
```

mzur (approved):
All good now :+1:

PR #729, opened 2023-12-28
Fix flag display

### LINKED ISSUE #718: Volume filter flagging does not work
The volume filter tab can activate a "flagging" (in contrast to the actual filtering) where files that match the filter rules get a blue dot but the files not matching the rules are still shown. This no longer works somehow, as the files don't get a dot. This should be fixed.

lehecht:
Fixes #718


mzur (approved):
That's a nice fix :wink:

PR #760, opened 2024-01-26
Disable tracking button if job limit is exceeded

### LINKED ISSUE #609: Improve UI of object tracking rate limiting
If the rate limit for object tracking is reached (#602), an error message is shown. This should be improved UI-wise. The UI can track the number of running object tracking jobs, too, and disable the "track object" button (for points and circles) once the limit is reached. The button can also show the loading spinner that was [recently added](https://github.com/biigle/core/blob/8f9c0864c54bb7ee57d2a910f0ce7ba1bccafbe1/resources/assets/js/annotations/components/controlButton.vue#L58-L61) to the "control button" component. Once a tracking job is finished, the button is enabled again.

lehecht:
Tracking annotation button is disabled if there are already 10 jobs in the queue.

Closes #609

### DIFF
--- a/app/Http/Controllers/Api/VideoAnnotationController.php
+++ b/app/Http/Controllers/Api/VideoAnnotationController.php
@@ -240,6 +240,7 @@ class VideoAnnotationController extends Controller
             $queue = config('videos.track_object_queue');
             Queue::pushOn($queue, new TrackObject($annotation, $request->user()));
             Cache::increment(TrackObject::getRateLimitCacheKey($request->user()));
+            $annotation->trackedAnnotationRate = $currentJobs / (float) ($maxJobs-1);
         }
 
         $annotation->load('labels.label', 'labels.user');
--- a/app/Http/Controllers/Api/VideoAnnotationController.php
+++ b/app/Http/Controllers/Api/VideoAnnotationController.php
@@ -240,7 +240,7 @@ class VideoAnnotationController extends Controller
             $queue = config('videos.track_object_queue');
             Queue::pushOn($queue, new TrackObject($annotation, $request->user()));
             Cache::increment(TrackObject::getRateLimitCacheKey($request->user()));
-            $annotation->trackedAnnotationRate = $currentJobs / (float) ($maxJobs-1);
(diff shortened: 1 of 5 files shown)


lehecht:
The test failed because GenerateImageAnnotationPatch was not found.

app/Http/Controllers/Api/VideoAnnotationController.php
```diff
@@ -240,6 +240,7 @@ public function store(StoreVideoAnnotation $request)
             $queue = config('videos.track_object_queue');
             Queue::pushOn($queue, new TrackObject($annotation, $request->user()));
             Cache::increment(TrackObject::getRateLimitCacheKey($request->user()));
+            $annotation->trackedAnnotationRate = $currentJobs / (float) ($maxJobs-1);
```
mzur, line 243: Please update the test for this.

You could also use this to prevent possible rounding/precision errors:

```suggestion
            $annotation->trackingJobLimitReached = $currentJobs === ($maxJobs - 1);
```

resources/assets/js/videos/videoContainer.vue
```diff
@@ -281,7 +285,15 @@ export default {
...
-                .then(this.addCreatedAnnotation, handleErrorResponse)
+                .then((res) => {
+                    this.addCreatedAnnotation(res);
+                    if (tmpAnnotation.track) {
+                        this.disableJobTracking = res.body.trackingJobLimitReached;
+                    }
+                }, (res) => {
+                    handleErrorResponse(res);
+                    this.disableJobTracking = res.status === 429;
+                })
```
mzur, line 296: I noticed that the annotation was no longer flashing in the timeline (to indicate that tracking is in progress, cf. [here](https://github.com/biigle/core/blob/f0f2aeb3c193986d47e75f1390daefdf12e4c54e/resources/assets/js/videos/components/annotationClip.vue#L108)). Git bisect lead me to this change. When I revert it, the flashing works. Can you please check why it doesn't work any more with this change?

mzur (changes requested):

mzur (approved):
Thanks!

PR #761, opened 2024-01-26
Generalize TileSingleImage

mtiessen1175:
Made TileSingleImage-class more general to be used with other formats such as GeoOverlays.

### DIFF
--- a/app/Jobs/TileSingleImage.php
+++ b/app/Jobs/TileSingleImage.php
@@ -24,7 +24,7 @@ class TileSingleImage extends Job implements ShouldQueue
      *
      * @var Image
      */
-    public $image;
+    public $file;
 
     /**
      * Path to the temporary storage file for the tiles.
@@ -32,6 +32,20 @@ class TileSingleImage extends Job implements ShouldQueue
      * @var string
      */
     public $tempPath;
+
+    /**
+     * The name of the permanent storage-disk where the tiles should be stored.
+     *
+     * @var string
+     */
+    public $storage;
+
+    /**
+     * Path to the tiles within the permanent storage-disk.
+     *
+     * @var string
+     */
+    public $fragment;
 
     /**
      * Ignore this job if the image does not exist any more.
@@ -43,14 +57,17 @@ class TileSingleImage extends Job implements ShouldQueue
     /**
      * Create a new job instance.
      *
-     * @param Image $image The image to generate tiles for.
+     * @param Image $file The image to generate tiles for.
      *
      * @return void
      */
-    public function __construct(Image $image)
+    public function __construct(Image $file)
     {
-        $this->image = $image;
-        $this->tempPath = config('image.tiles.tmp_dir')."/{$image->uuid}";
+        $this->file = $file;
+        $this->tempPath = config('image.tiles.tmp_dir')."/{$file->uuid}";
+        // for uploadToStorage method
+        $this->storage = 'image.tiles.disk';
+        $this->fragment = fragment_uuid_path($file->uuid);
     }
 
     /**
@@ -61,10 +78,10 @@ class TileSingleImage extends Job implements ShouldQueue
     public function handle()
     {
         try {
-            FileCache::getOnce($this->image, [$this, 'generateTiles']);
+            FileCache::getOnce($this->file, [$this, 'generateTiles']);
             $this->uploadToStorage();
-            $this->image->tilingInProgress = false;
-            $this->image->save();
+            $this->file->tilingInProgress = false;
+            $this->file->save();
         } finally {
             File::deleteDirectory($this->tempPath);
         }
@@ -73,10 +90,10 @@ class TileSingleImage extends Job implements ShouldQueue
     /**
      * Generate tiles for the image and put them to temporary storage.
      *
-     * @param Image $image
+     * @param Image $file
      * @param string $path Path to the cached image file.
      */
-    public function generateTiles(Image $image, $path)
+    public function generateTiles($file, $path)
     {
         $this->getVipsImage($path)->dzsave($this->tempPath, [
             'layout' => 'zoomify',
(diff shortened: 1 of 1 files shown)


app/Jobs/TileSingleImage.php
```diff
@@ -43,14 +57,17 @@ class TileSingleImage extends Job implements ShouldQueue
     /**
      * Create a new job instance.
      *
-     * @param Image $image The image to generate tiles for.
+     * @param Image $file The image to generate tiles for.
      *
      * @return void
      */
-    public function __construct(Image $image)
+    public function __construct(Image $file)
```
mzur, line 64: This class is not really more generalized if you still restrict it to `Image` here. Maybe you want a `Biigle\FileCache\Contracts\File` instead?

app/Jobs/TileSingleImage.php
```diff
@@ -43,14 +57,17 @@ class TileSingleImage extends Job implements ShouldQueue
...
     {
-        $this->image = $image;
-        $this->tempPath = config('image.tiles.tmp_dir')."/{$image->uuid}";
+        $this->file = $file;
+        $this->tempPath = config('image.tiles.tmp_dir')."/{$file->uuid}";
+        // for uploadToStorage method
+        $this->storage = 'image.tiles.disk';
```
mzur, line 69: I would add this as a constructor argument. Also it's better to store the actual path, not the config key.

app/Jobs/TileSingleImage.php
```diff
@@ -43,14 +57,17 @@ class TileSingleImage extends Job implements ShouldQueue
...
+        $this->fragment = fragment_uuid_path($file->uuid);
```
mzur, line 70: This could also be a constructor argument (i.e. the "target path"). This way you can get around the fact that an `Image` or a `Biigle\FileCache\Contracts\File` is not guaranteed to have a uuid.

mzur (changes requested):

app/Jobs/TileSingleImage.php
mtiessen1175: resolved with 4f42cb7bf35ace46d1cbeacf4f6b98fc776ec3f0

app/Jobs/TileSingleImage.php
mtiessen1175: resolved 4f42cb7bf35ace46d1cbeacf4f6b98fc776ec3f0

app/Jobs/TileSingleImage.php
mtiessen1175: resolved 0d84d7e911b06aa640fd65f1e8be3b0b13508625

mzur (commented):
Looks good. There is only a minor comment left.

I should have a look why Psalm is complaining about the missing class.

app/Jobs/MigrateTiledImage.php
```diff
@@ -16,6 +16,13 @@ class MigrateTiledImage extends TileSingleImage
      */
     public $disk;
 
+    /**
+     * Path to the tiles within the permanent storage-disk.
+     *
+     * @var string
+     */
+    public $targetPath;
+
```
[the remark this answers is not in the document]
mtiessen1175: resolved da20c9d63d698f264a7e3dcf9d4d3b885fb1a8a4

app/Jobs/TileSingleImage.php
```diff
@@ -61,10 +80,10 @@ public function __construct(Image $image)
...
         try {
-            FileCache::getOnce($this->image, [$this, 'generateTiles']);
+            FileCache::getOnce($this->file, [$this, 'generateTiles']);
             $this->uploadToStorage();
-            $this->image->tilingInProgress = false;
-            $this->image->save();
+            $this->file->tilingInProgress = false;
+            $this->file->save();
         } finally {
             File::deleteDirectory($this->tempPath);
```
mzur, line 88: Now I see why the linter complains. You have removed the `File` facade to use the `File` contract from the file cache package. However, the facade is still needed for the call to `File::deleteDirectory()`. Maybe you could add a test case that actually calls the `handle()` method of the job. With such a test, the issue would have become obvious.

mzur (changes requested):

app/Jobs/TileSingleImage.php
mtiessen1175: Thank you for finding that issue!
resolved c5cda584c4a1353dc8ed9945e299ba17dd225417

mzur (approved):
Looks good. This belongs to a new feature of biigle/geo, right? I'll wait until it is ready then and the PRs can be merged together.

mzur:
@mtiessen1175 Can you please resolve the conflicts of this branch?

mtiessen1175:
@mzur, since having merged the master-branch into this one locally, I get a few strange errors (such as the one described [here](https://stackoverflow.com/questions/77148705/laravel-localhost-page-showing-the-error-illuminate-view-fileviewfinder-const)) when trying to execute tests or even do `artisan` commands. I guess this has to do with the upgrade to Laravel 11 (as done [here](https://github.com/biigle/core/commit/dfea5e6d2b252848e7f161ebac0e2754fa313479))? Could I safely upgrade my local instance (which currently is v.10.48.4)? Or do I have to be aware of anything that I should do before upgrading?

mtiessen1175:
@mzur, the php-linter still complains about accessing undefined properties, such as _Biigle\FileCache\Contracts\File::$uuid._ There are several ways to solve this type of error, as explained [here](https://phpstan.org/blog/solving-phpstan-access-to-undefined-property) (e.g. declaring the properties in the class docs with `@property string $uuid`). The linter also complains about Laravel eloquent model attributes such as _Biigle\FileCache\Contracts\File::save()_, which I find strange.

How would you approach this error? In the case at hand with the `TileSingleImage` and `TileSingleOverlay` classes, narrowing the type the property is accessed on could be one solution. I am thinking of creating an abstract class or interface `TileSingleObject` and then specify the type (Image, Overlay, etc.) in its child classes instead of using the more general _FileCache_ type.

mzur:
I'll have a look at the linter errors. If they are genuine, I'll suggest a fix and if not, I'll try to update the linter settings. The linter has become a little more strict since the Laravel 11 upgrade.

mzur:
The linter errors are genuine. We have three errors:

1. `Access to an undefined property Biigle\FileCache\Contracts\File::$uuid`: This is true, as the `File` interface does not specify a UUID. The `Biigle\Image` has a UUID but not the `Biigle\Modules\Geo\GeoOverlay`. So this would actually result in an incorrect `tempPath` for a GeoOverlay (if you wouldn't override this in `TileSingleOverlay`).

2. `Access to an undefined property Biigle\FileCache\Contracts\File::$tilingInProgress`: Also correct. A `GeoOverlay` does not have this attribute. The job fails because of this.

3. `Call to an undefined method Biigle\FileCache\Contracts\File::save()`: Also true. The `File` here is no Eloquent model and does not define this method.

To fix these errors I suggest that you implement a more abstract version of `TileSingleImage`. Then implement one special version to process `Biigle\Image` (with `uuid` for the tempPath and `tilingInProgress` and `save()`) and in biigle/geo keep `TileSingleOverlay` (with `id` for the tempPath and without `tilingInProgress` and `save()`).

mtiessen1175:
> I'll have a look at the linter errors. If they are genuine, I'll suggest a fix [...]

resolved the linter errors aa918234dc25300185a75e6bee21dd741efc6e38

mzur (approved):

PR #762, opened 2024-01-27
Fix Vips thumbnails aspect ratio and implement thumbnail preview feature on the frontend

gkourie:
- Fixes an issue with the aspect ratio of Vips thumbnails buffer by setting the `size` to `force` (see [issue-75](https://github.com/libvips/php-vips/issues/75)).
- Implements the thumbnail preview feature on the frontend.
- Adjusts the corresponding test cases to reflect these changes.


mzur (changes requested):
Fantastic! Really well done :clap:

I have a few minor comments (see below). These are the major points:

1. The preview thumbnail often briefly flashes before it repositions itself when the image is loaded or hides itself if the loading failed. Flashing with image:
   
   [Screencast from 01.02.2024 16:01:41.webm](https://github.com/biigle/core/assets/2457311/a96b19cf-d1af-420a-aa0d-11f2246c9dfc)

   Flashing without image:

   [Screencast from 01.02.2024 16:03:34.webm](https://github.com/biigle/core/assets/2457311/83c7132a-3ffd-4a04-a195-c12383356e73)

2. The thumbnail does not show when I start at the left of the timeline. Once I cross some sort of threshold, it shows. Then I can move backwards again and it keeps showing (the correct images, too). I use Firefox if that matters.

   [Screencast from 01.02.2024 16:02:40.webm](https://github.com/biigle/core/assets/2457311/b112883e-d288-4eff-8da2-dd981d83f8ac)


3. I was going to say that we have to take care of deleting the sprites when a video is deleted but luckily the existing code just [deletes the whole directory](https://github.com/biigle/core/blob/f547b9d65159b0f74a2e045775c523616b46b00f/app/Listeners/CleanupVideoThumbnails.php#L22). So nothing to do here :+1: 

4. Please [add a paragraph to the manual](https://github.com/biigle/core/blob/f547b9d65159b0f74a2e045775c523616b46b00f/resources/views/manual/tutorials/videos/sidebar.blade.php#L45-L73) that explains the new button.

That's it! After this is merged I will check the issue with `['size' => 'force']` and implement the console command to generate missing sprites.

mzur:
Also can you please sync the `dev-master` branch of your fork? Then the test-modules check should work, too.

gkourie:
> Also can you please sync the `dev-modules` branch of your fork? Then the test-modules check should work, too.

I did, but still not working.

mzur (changes requested):
Thanks! Looks even better now. One final thing came to my attention:

If the sprites do not exist, the thumbnail component sends an HTTP request basically on each mousemove event. Instead, I suggest that a) the component should try each URL only once or twice and then give up (it will try again if the mouse is moved away and then back again, creating the component anew) or b) try each URL only once each 10, 20 or 30 seconds.

mzur:
>>  Also can you please sync the dev-modules branch of your fork? Then the test-modules check should work, too.
> I did, but still not working.

Sorry, it seems like `master` had to be synced instead. I did this now and the tests pass.

mzur (approved):
Thanks :+1:

PR #764, opened 2024-01-29
fix error on console command creating new user without admin rights

ToukL:
The "new user" console command without admin rights was not handled and returned an error.


mzur (approved):
Great, thank you!

Issue #763, opened 2024-01-29
Ignore zero-sized rectangles


mzur:
We found out that some people are drawing lots of rectangles with zero size (i.e. all four points have the same coordinates. These rectangles should be ignored in the UI and not sent to the backend.

Also try to reproduce why this happens. Maybe people double-click when creating or finishing a rectangle? Then we should catch that, too.

mzur:
This depends on how these zero-rectangles are created. If they are created involuntarily (e.g. a user draws a valid rectangle, then makes a double-click and thereby creates a second zero-rectangle) then I would say this should just be silently ignored. If zero-rectangles can only be created "intentionally" by clicking three times in a row on the same spot then maybe we should show a warning.

neopaulit4n:
I am one of the users experiencing the zero-sized rectangles. I notice this happens whenever I start drawing a polygon using a straight horizontal line only. If I start drawing with a vertical or diagonal line, it never has this issue. However, starting to draw the polygon with a perfectly horizontal line then clicking the first time to set it causes the blue dashed drawing line to disappear every time. Nothing happens until you click again anywhere on the image (where you would normally want to draw the height of the polygon and complete the rectangle) and then an annotation appears in the annotations list, but the bounding box is invisible in the image itself.

lehecht:
Unfortunately, I couldn't reproduce this bug. But with #769 everything seems to work. I can draw polygons like  @neopaulit4n described without any problems or weird behaviour.

PR #769, opened 2024-02-06
Reject invalid annotation shapes

lehecht:
Reject invalid annotation shapes by not saving them on frontend side and rejecting them on the backend side. Additionally extend tests.

### DIFF
--- a/app/Http/Requests/StoreImageAnnotation.php
+++ b/app/Http/Requests/StoreImageAnnotation.php
@@ -50,7 +50,59 @@ class StoreImageAnnotation extends StoreImageAnnotationLabel
         $validator->after(function ($validator) {
             if ($this->input('shape_id') === Shape::wholeFrameId()) {
                 $validator->errors()->add('shape_id', 'Image annotations cannot have shape WholeFrame.');
+                return;
             }
+
+            if (!$this->input('points')) {
+                $validator->errors()->add('points', 'Invalid shape. No points given.');
+                return;
+            }
+
+            $fails = false;
+            $points = $this->input('points');
+
+            switch ($this->input('shape_id')) {
+                case (Shape::circleId()):
+                    // Circle needs x,y coordinate and non zero radius
+                    $fails = count($points) !== 3 || intval(($points)[2]) === 0;
+                    break;
+                case (Shape::ellipseId()):
+                case (Shape::rectangleId()):
+                    // Ellipse and rectangle need exactly 4 distinct points
+                    $fails = count($points) !== 8 || $this->countDistinctPoints($points) !== 4;
+                    break;
+                case (Shape::lineId()):
+                    // Line need at least 2 distinct points
+                    $fails = count($points) < 4 || $this->countDistinctPoints($points) < 2;
+                    break;
+                case (Shape::polygonId()):
+                    // Polygon needs at least 3 distinct points
+                    $fails = count($points) < 6 || $this->countDistinctPoints($points) < 3;
+                    break;
+                case (Shape::pointId()):
+                    // Points needs exactly one x and y coordinate
+                    $fails = count($points) !== 2;
+            }
+
+            if ($fails) {
+                $validator->errors()->add('points', 'Invalid shape.');
+            }
+
         });
+    }
+
+    /**
+     * Counts number of distinct points
+     * @param array $points containing the coordinates
+     * @return int number of distinct points
+     * **/
+    private function countDistinctPoints($points)
+    {
+        $points = collect($points);
+        // Use values to reset index
+        $x = $points->filter(fn ($x, $idx) => $idx % 2 === 0)->values();
+        $y = $points->filter(fn ($x, $idx) => $idx % 2 === 1)->values();
+        $coords = collect($x->map(fn ($x, $idx) => [$x, $y[$idx]]))->unique();
+        return count($coords);
     }
 }
--- a/app/Http/Requests/StoreImageAnnotation.php
+++ b/app/Http/Requests/StoreImageAnnotation.php
@@ -50,59 +50,7 @@ class StoreImageAnnotation extends StoreImageAnnotationLabel
         $validator->after(function ($validator) {
             if ($this->input('shape_id') === Shape::wholeFrameId()) {
                 $validator->errors()->add('shape_id', 'Image annotations cannot have shape WholeFrame.');
-                return;
             }
-
-            if (!$this->input('points')) {
-                $validator->errors()->add('points', 'Invalid shape. No points given.');
-                return;
-            }
-
-            $fails = false;
-            $points = $this->input('points');
-
-            switch ($this->input('shape_id')) {
-                case (Shape::circleId()):
-                    // Circle needs x,y coordinate and non zero radius
-                    $fails = count($points) !== 3 || intval(($points)[2]) === 0;
-                    break;
-                case (Shape::ellipseId()):
-                case (Shape::rectangleId()):
-                    // Ellipse and rectangle need exactly 4 distinct points
-                    $fails = count($points) !== 8 || $this->countDistinctPoints($points) !== 4;
-                    break;
-                case (Shape::lineId()):
-                    // Line need at least 2 distinct points
-                    $fails = count($points) < 4 || $this->countDistinctPoints($points) < 2;
-                    break;
-                case (Shape::polygonId()):
-                    // Polygon needs at least 3 distinct points
-                    $fails = count($points) < 6 || $this->countDistinctPoints($points) < 3;
-                    break;
-                case (Shape::pointId()):
-                    // Points needs exactly one x and y coordinate
-                    $fails = count($points) !== 2;
-            }
-
-            if ($fails) {
-                $validator->errors()->add('points', 'Invalid shape.');
-            }
-
         });
-    }
-
-    /**
-     * Counts number of distinct points
-     * @param array $points containing the coordinates
-     * @return int number of distinct points
-     * **/
-    private function countDistinctPoints($points)
-    {
-        $points = collect($points);
-        // Use values to reset index
-        $x = $points->filter(fn ($x, $idx) => $idx % 2 === 0)->values();
-        $y = $points->filter(fn ($x, $idx) => $idx % 2 === 1)->values();
-        $coords = collect($x->map(fn ($x, $idx) => [$x, $y[$idx]]))->unique();
-        return count($coords);
     }
 }
--- a/app/Traits/HasPointsAttribute.php
+++ b/app/Traits/HasPointsAttribute.php
@@ -83,7 +83,7 @@ trait HasPointsAttribute
         // Use values to reset index
         $x = $points->filter(fn ($x, $idx) => $idx % 2 === 0)->values();
         $y = $points->filter(fn ($x, $idx) => $idx % 2 === 1)->values();
-        $coords = collect($x->map(fn ($x, $idx) => [$x, $y[$idx]]))->unique();
+        $coords = $x->zip($y)->unique();
         return count($coords);
     }
 }
--- a/app/Traits/HasPointsAttribute.php
+++ b/app/Traits/HasPointsAttribute.php
@@ -46,7 +46,7 @@ trait HasPointsAttribute
         }
 
         if (!$valid) {
-            throw new Exception('Invalid number of points for shape '.$this->shape->name.'!');
+            throw new Exception('Invalid (number of) points for shape '.$this->shape->name.'!');
         }
 
         if ($this->shape_id === Shape::polygonId()) {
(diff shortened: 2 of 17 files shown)


mzur:
Is there any issue besides #763 associated with this PR?

app/Http/Requests/StoreImageAnnotation.php
```diff
@@ -50,7 +50,59 @@ public function withValidator($validator)
         $validator->after(function ($validator) {
             if ($this->input('shape_id') === Shape::wholeFrameId()) {
                 $validator->errors()->add('shape_id', 'Image annotations cannot have shape WholeFrame.');
+                return;
             }
+
+            if (!$this->input('points')) {
+                $validator->errors()->add('points', 'Invalid shape. No points given.');
+                return;
+            }
```
mzur, line 59: This is not needed because we already have the `required` validation rule for the attribute.

app/Http/Requests/StoreImageAnnotation.php
```diff
@@ -50,7 +50,59 @@ public function withValidator($validator)
...
+                    break;
+                case (Shape::pointId()):
+                    // Points needs exactly one x and y coordinate
+                    $fails = count($points) !== 2;
+            }
+
+            if ($fails) {
+                $validator->errors()->add('points', 'Invalid shape.');
+            }
+
```
mzur, line 90: The points validation is already done in the [HasPointsAttribute](https://github.com/biigle/core/blob/95903302705a34a4b2156c93351669352df3f5db/app/Traits/HasPointsAttribute.php#L16-L52) trait. You can probably update the validation there with the additional stuff you added here and it should work everywhere (your tests will tell you). This will also work with the "update" API endpoints that you didn't consider here.

lehecht:
I don't think so. The bug of issue #770 is also located where changes of this PR were made, but it was not caused by these changes.

resources/assets/js/annotations/annotatorContainer.vue
```diff
@@ -571,8 +571,27 @@ export default {
...
+                    count = 'at least 3';
+                    break;
+                case 'Rectangle':
+                case 'Ellipse':
+                    count = 4;
+                    break;
+                default:
+                    return;
+            }
+            Messages.danger(`Invalid shape. ${shape} needs ${count} non-overlapping vertices.`);
```
mzur, line 594: I would phrase this a little simpler for the non-technical users:

```suggestion
            Messages.danger(`Invalid shape. ${shape} needs ${count} different points.`);
```

resources/assets/js/annotations/components/annotationCanvas.vue
```diff
@@ -515,44 +515,41 @@ export default {
...
+                // This must be done in the change event handler.
+                // Not exactly sure why.
+                this.annotationSource.once('change', () => {
+                    if (this.annotationSource.hasFeature(e.feature)) {
+                        this.annotationSource.removeFeature(e.feature);
                     }
+                });
+                this.$emit('is-invalid-shape', e.feature.getGeometry().getType());
+                return;
+            }
```
mzur, line 528: There should be no "invalid shape" error if no label is selected. This can be handled separately:

```suggestion
            if (!this.hasSelectedLabel) {
                this.annotationSource.removeFeature(e.feature);
                return;
            }
            
            if (this.isInvalidShape(e.feature)) {
                // This must be done in the change event handler.
                // Not exactly sure why.
                this.annotationSource.once('change', () => {
                    if (this.annotationSource.hasFeature(e.feature)) {
                        this.annotationSource.removeFeature(e.feature);
                    }
                });
                this.$emit('is-invalid-shape', e.feature.getGeometry().getType());
                return;
            }
```

resources/assets/js/annotations/components/annotationCanvas.vue
```diff
@@ -679,6 +676,26 @@ export default {
...
+                case 'Ellipse':
+                    points = geometry.getCoordinates()[0];
+                    return (new Set(points.map(xy => String([xy])))).size !== 4;
+                case 'Polygon':
+                    points = geometry.getCoordinates()[0];
+                    return (new Set(points.map(xy => String([xy])))).size < 3;
+                default:
+                    return false;
+            }
+        },
```
mzur, line 698: You can reuse the function from `videos/utils.js`. Usually the video annotation tool reuses stuff from the image annotation tool, so maybe you can put the function into an `annotations/utils.js` and import it from there?

resources/assets/js/videos/components/videoScreen/drawInteractions.vue
```diff
@@ -163,20 +164,22 @@ export default {
...
         extendPendingAnnotation(e) {
+            if (isInvalidShape(e.feature)) {
+                // Disallow shapes with too few points.
+                this.$emit('is-invalid-shape', e.feature.getGeometry().getType());
+                // Wait for this feature to be added to the source, then clear.
+                this.pendingAnnotationSource.once('addfeature', () => {
+                    this.resetPendingAnnotation();
+                });
+                // Reset mode to enable drawing again by closing button 
+                this.resetInteractionMode();
```
mzur, line 175: This wasn't done before. Wouldn't it be easier if the button stayed active so the user can immediately try again?

resources/assets/js/videos/videoContainer.vue
```diff
@@ -629,8 +629,27 @@ export default {
...
+                    count = 'at least 3';
+                    break;
+                case 'Rectangle':
+                case 'Ellipse':
+                    count = 4;
+                    break;
+                default:
+                    return;
+            }
+            Messages.danger(`Invalid shape. ${shape} needs ${count} non-overlapping vertices.`);
```
mzur, line 652:
```suggestion
            Messages.danger(`Invalid shape. ${shape} needs ${count} different points.`);
```

tests/php/Http/Controllers/Api/ImageAnnotationControllerTest.php
mzur, line 1: Please also add a test for the update API endpoint.

tests/php/Http/Controllers/Api/VideoAnnotationControllerTest.php
mzur, line 1: Why not also test the LineString annotation as with the image annotation controller?

mzur, line 1: Please also add a test for the update API endpoint.

mzur (changes requested):

mzur:
Will #770 be fixed by this PR? Then please add it to the "development" section on the right so it will be closed with this PR.

app/Http/Requests/StoreImageAnnotation.php
mzur: It just came into my mind that the trait with the validation method was probably created before I used form requests with validation. This is why it is not located where you would expect validation logic. If you want to clean up the code, you could also implement the point validation as a custom validation rule that can be reused in the store/update requests.

lehecht:
> Will #770 be fixed by this PR? Then please add it to the "development" section on the right so it will be closed with this PR.

No, it won't.

lehecht:
@mzur 
Why can a line just contain one point in [this test](https://github.com/biigle/core/blob/27144e92cbc4f9d4f00ab38a229c0a5c0b31dd9f/tests/php/VideoAnnotationTest.php#L107)?

mzur:
> Why can a line just contain one point in [this test](https://github.com/biigle/core/blob/27144e92cbc4f9d4f00ab38a229c0a5c0b31dd9f/tests/php/VideoAnnotationTest.php#L107)?

I see no particular reason. This is probably one aspect of the validation that can be improved.

> Is it permissible for a polygon to be a line?

No, the last point of a polygon must be equal to the first point. I think this is validated somewhere, too :thinking:

lehecht:
> [](https://github.com/biigle/core/commit/cc75bb70d2d2abe53904b7219e20a205af3816cf)
> 
> I guess with the current validation, you _can_ make a line with three points (a, b, a). I'm not sure if we should disallow this.

Should I then leave it like it is?

lehecht:
I also found a [test](https://github.com/biigle/core/blob/6a46ef774cb6f9adaad249615be64c333094d236/tests/php/VideoAnnotationTest.php#L117-L120) where a polygon can be a point. Should this be kept that way, too?

mzur:
>  Should I leave it like it is?

If it's not too much effort, I think this should be rejected. Polygons with a zero size could cause all sorts of issues.

> I also found a [test](https://github.com/biigle/core/blob/6a46ef774cb6f9adaad249615be64c333094d236/tests/php/VideoAnnotationTest.php#L117-L120) where a polygon can be a point. Should this be kept that way, too?

No. Also not the "point-line" (in the test above).

app/Traits/HasPointsAttribute.php
```diff
@@ -29,18 +29,24 @@ public function validatePoints(array $points)
...
-                $valid = $size === 3;
+                $valid = $size === 3 && intval(($points)[2]) !== 0;
                 break;
             case Shape::ellipseId():
             case Shape::rectangleId():
-                $valid = $size === 8;
+                $valid = $size === 8 && $this->countDistinctCoordinates($points) === 4;
+                break;
+            case Shape::polygonId():
+                $valid = $size % 2 === 0 && count($points) >= 6 && $this->countDistinctCoordinates($points) >= 3;
```
mzur, line 39: This must be:
```suggestion
                $valid = $size % 2 === 0 && count($points) >= 8 && $this->countDistinctCoordinates($points) >= 3;
```
I think. It can't have only 3 points because the first must equal the last and we no longer allow a line here, right?

app/Traits/HasPointsAttribute.php
```diff
@@ -65,4 +71,19 @@ public function setPointsAttribute(array $points)
...
+     * @param array $points containing the coordinates
+     * @return int number of distinct points
+     * **/
+    private function countDistinctCoordinates($points)
+    {
+        $points = collect($points);
+        // Use values to reset index
+        $x = $points->filter(fn ($x, $idx) => $idx % 2 === 0)->values();
+        $y = $points->filter(fn ($x, $idx) => $idx % 2 === 1)->values();
+        $coords = collect($x->map(fn ($x, $idx) => [$x, $y[$idx]]))->unique();
```
mzur, line 86: This could also be a perfect use case for the [zip()](https://laravel.com/docs/10.x/collections#method-zip) method.
```suggestion
        $coords = $x->zip($y)->unique();
```

resources/assets/js/videos/components/videoScreen/drawInteractions.vue
```diff
@@ -169,10 +169,11 @@ export default {
                 this.$emit('is-invalid-shape', e.feature.getGeometry().getType());
                 // Wait for this feature to be added to the source, then clear.
                 this.pendingAnnotationSource.once('addfeature', () => {
-                    this.resetPendingAnnotation();
+                    // Clear annotation source only, otherwise pending annotation is invalid
+                    // and shape buttons remain disabled
+                    this.pendingAnnotationSource.clear();
```
mzur, line 174: This doesn't work correctly, yet. If I create a multi-frame annotation and, for example, create two valid polygons first and then an invalid polygon, all three pending annotations are cleared where only the last one should disappear so I can try again.

See: [Screencast from 05.03.2024 09:18:13.webm](https://github.com/biigle/core/assets/2457311/496dd253-1ec9-4b48-aa5c-2359c03247dc)

resources/assets/js/annotations/utils.js
```diff
@@ -0,0 +1,29 @@
...
+        case 'LineString':
+            points = geometry.getCoordinates();
+            return (new Set(points.map(xy => String([xy])))).size < 2;
+        case 'Rectangle':
+        case 'Ellipse':
+            points = geometry.getCoordinates()[0];
+            return (new Set(points.map(xy => String([xy])))).size !== 4;
+        case 'Polygon':
+            points = geometry.getCoordinates()[0];
+            return (new Set(points.map(xy => String([xy])))).size < 3;
```
mzur, line 23: Maybe this also has to be:
```suggestion
            return (new Set(points.map(xy => String([xy])))).size < 4;
```
(see above?)

mzur (changes requested):

resources/assets/js/annotations/utils.js
lehecht: At this point, it's right. The points array just saves the coordinates without adding the last one to the end. So a line would consist of only two points and not three.

mzur: In my test this wasn't the case. When I draw a triangle polygon, I get 4 points here.

lehecht: Ah, I see. My argumentation was wrong here. Sorry. 

If I apply this change, I cannot draw valid triangle polygons any more. 
There are 4 coordinates, where the first one is duplicated and added to the end. But the number of distinct coordinates still remains by at least 3. Your change would allow only rectangle polygons.

mzur (approved):

PR #778, opened 2024-02-13
Chunk annotations and annotation labels

### LINKED ISSUE #777: Chunk annotation inserts in clone volume job
The insert statements for annotations in the clone volume job ([here](https://github.com/biigle/core/blob/master/app/Jobs/CloneImagesOrVideos.php#L288), [here](https://github.com/biigle/core/blob/master/app/Jobs/CloneImagesOrVideos.php#L305C1-L305), [here](https://github.com/biigle/core/blob/master/app/Jobs/CloneImagesOrVideos.php#L445) and [here](https://github.com/biigle/core/blob/master/app/Jobs/CloneImagesOrVideos.php#L462)) must be chunked so they don't reach the 65k parameter limit of the database.

@lehecht Could you please prioritize this? There is currently a failed job stuck in the queue that I can resubmit once this is fixed.

lehecht:
Use chunking to prevent exceeding the database parameter limit during volume cloning.

Fixes #777.

### DIFF
--- a/app/Jobs/CloneImagesOrVideos.php
+++ b/app/Jobs/CloneImagesOrVideos.php
@@ -17,6 +17,7 @@ use Biigle\VideoAnnotation;
 use Biigle\VideoAnnotationLabel;
 use Biigle\VideoLabel;
 use Biigle\Volume;
+use Illuminate\Database\Eloquent\Collection;
 use Illuminate\Http\Request;
 use Illuminate\Queue\InteractsWithQueue;
 use Illuminate\Queue\SerializesModels;
@@ -253,6 +254,7 @@ class CloneImagesOrVideos extends Job implements ShouldQueue
         }
 
         $chunkSize = 100;
+        $databaseRecordLimit = 64999;
         $newImageIds = $copy->images()->orderBy('id')->pluck('id');
         $volume->images()
             ->with([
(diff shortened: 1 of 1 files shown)


app/Jobs/CloneImagesOrVideos.php
```diff
@@ -267,9 +269,9 @@ private function copyImageAnnotation($volume, $copy, $selectedFileIds, $selected
                 $newImageIds,
                 $chunkSize,
                 $usedAnnotationIds,
-                $selectedLabelIds
+                $databaseRecordLimit,
```
mzur, line 272: This seems wrong.

app/Jobs/CloneImagesOrVideos.php
```diff
@@ -253,6 +254,7 @@ private function copyImageAnnotation($volume, $copy, $selectedFileIds, $selected
         }
 
         $chunkSize = 100;
+        $databaseRecordLimit = 64999;
```
mzur, line 257: The limit is a *parameter* limit, i.e. `records * attributes = parameters`. This means the number must be much lower here. I think you can use 10,000 here (the annotations and annotation labels have 5–6 attributes).

app/Jobs/CloneImagesOrVideos.php
```diff
@@ -17,6 +17,7 @@
 use Biigle\VideoAnnotationLabel;
 use Biigle\VideoLabel;
 use Biigle\Volume;
+use Illuminate\Database\Eloquent\Collection;
```
mzur, line 20: I think it's rather unusual to use a database collection directly. Usually, this is only returned from a database query. Instead, if you want to use a collection, just use the `collect([])` helper. This gives you a more generic `Illuminate\Support\Collection`.

mzur (changes requested):

app/Jobs/CloneImagesOrVideos.php
mzur: Not wrong after all, `selectedLabelIds` just isn't used.

mzur (approved):
Thanks! The previous variable name "record limit" was more appropriate (65k is the parameter limit, 10k is our record limit here) but it's just a name...

PR #780, opened 2024-02-16
Video sprites

### LINKED ISSUE #375: Video preview thumbnails in annotation tool
We could expand the video preview thumbnail feature (which is displayed when the mouse hovers over a video thumbnail in the volume overview). Many popular video platforms (e.g. YouTube) show a preview thumbnail of the video when the mouse hovers over the video timeline. We can do this, too. We just have to generate more thumbnails when the video is processed (e.g. one thumbnail each 10 s). These thumbnails can then be used for the existing preview feature (subset of preview thumbnails in the volume overview) and when the user hovers over the top of the timeline in the video annotation tool (all preview thumbnails).

mzur:
Resolves #375 

Remaining tasks:

- [x] ~~Implement a console command to generate missing sprites of existing (local) videos~~
- [x] Try to make generating sprites more efficient (with a direct ffmpeg call such as `ffmpeg -i '{$path}' -vf fps={$frameRate} /tmp/frame%03d.png` which reads the video only once instead of reading it anew for each frame).
- [x] ~~Maybe simplify the config definition in the blade template.~~
- [x] Fix the issue with `['size' => 'force']`. The regular thumbnails are no longer correct with this setting.
- [x] Move the current hover time (now shown in muted color on the left) to the thumbnail element. If no thumbnail is available, only show the time there. Remove the time from the element on the left.
- [x] Fix the bug where wrong thumbnails are displayed after switching videos
- [x] Use `v-show` to display the thumbnail on the mouse event.
- [x] Update the settings switch to only control the preview thumbnail and not also the hover time display.
- [x] Preload the previous and next sprite for instant transitions
- [x] Maybe use `throttle()` for the mouse event (e.g. to limit HTTP requests if sprite images could not be found).


mzur:
I think the console command is not needed because it can just be done manually in the tinker console. Something like this:

```php
Volume::where('media_type_id', MediaType::videoid())
   ->where('url', 'not like', 'http%')
   ->eachById(fn ($volume) =>
      $volume->videos()->eachById(fn ($video) =>
         Queue::pushOn('low', new ProcessNewVideo($video)
      )
   );
```

mzur:
Also I think the `['size' => 'force']` issue may be resolved with the custom ffmpeg call. The call will generate files on disk instead of returning the frame as a buffer. This won't use the same code than the old way of generating thumbnails, so the old way can be reverted to not use `['size' => 'force']`.

Maybe we could even do it the other way around? The custom ffmpeg call generates the frame files and then a subsample of these are used as the "old" thumbnails before all of the files are combined into the sprites for the "new" thumbnail feature (if you know what I mean...).

mzur (commented):
I'm starting with the high-level comments again. These first comments all depend on each other:

- You scale the extracted video frames using two other `ffmpeg` command line processes. Calling external processes is slow and ffmpeg is not particularly well suited for image scaling. Instead, read each extracted frame file and scale it with Vips. There will be no need to store the scaled thumbnails as files again (see below).

- Storing files in a storage disk does not work like that. You get a storage path from the disk and then directly write the files to this path. This only works because your development instance exclusively uses local storage disks. Instead, you have to read a file into a variable/buffer and then use `$disk->put($path, $contents)` to upload the file. Also, you are currently storing *all* extracted frames in `generateVideoThumbnails()` but we only need the 10 that are used as video thumbnails later.

- Building on the previous two points, you could implement a single loop over the extracted video frames. Each frame is read into a variable, scaled by Vips and appended to an array that collects all thumbnails for a sprite. When the array is full, the sprite image is created with `arrayjoin()`. At even intervals during the loop, a scaled video frame is also directly uploaded as a single thumbnail. This will produce the thumbnails and sprites in the same loop.

- When you only store the 10 single thumbnails for a video, there is no need to update the code that displays the videos in the volume overview. It can work as before.

Additional unrelated comments:

- There should be no black bars in the thumbnails or sprites. If the video aspect ratio does not fit to the configured size (which will happen most of the time), the height/width of the thumbnails and sprites should be changed accordingly. Think of it like the CSS [object-fit: contain](https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit) behavior. The configured size of 360x270 is the box and the thumbnail is scaled until it is completely contained in the box. This worked already with the original code of this PR and the scaling done by Vips.

- While the thumbnails have 360x270 px, they should be displayed as 180x135 px in the UI. Currently, the preview frames in the video annotation tool are shown in the original size. The thumbnails are generated at twice the size so they look good on high-dpi devices.

lehecht:
> Also, you are currently storing _all_ extracted frames in `generateVideoThumbnails()` but we only need the 10 that are used as video thumbnails later.

But didn't I need all thumbnails for the sprites at the end? Or should I generate the first 10 images twice, one for the normal thumbnails and one for the sprites?

lehecht:
> The configured size of 360x270 is the box and the thumbnail is scaled until it is completely contained in the box.

So the image should be stretched if the width or height doesn't fit in the UI? Because the image in the CSS contain example also has black bars. Or did you mean that Vips should generate images without bars, but these bars are shown in the UI if the ratio doesn't fit?

lehecht:
Is it right that for short videos only five `($minThumbnails)` thumbnails should be generated? Because the thumbnails preview on the volume overview shows black images for the missing ones at the moment.
https://github.com/biigle/core/blob/efab8a51231edcd2dcc79096c563e66c6c9bc803/app/Jobs/ProcessNewVideo.php#L254-L255

lehecht:
> Maybe use `throttle()` for the mouse event (e.g. to limit HTTP requests if sprite images could not be found).

What exactly do you have in mind here? My idea was to check the time difference between the mouse movements on the timeline. If the difference is above a certain threshold and the sprite was found, the image will be shown.
(Remark: I work on a different branch for this feature. So the tasks won't mix up.)

mzur:
>> The configured size of 360x270 is the box and the thumbnail is scaled until it is completely contained in the box.
>
> [...] Or did you mean that Vips should generate images without bars, whereas these bars are shown in the UI if the ratio doesn't fit?

Partly, yes. The images should be generated without bars in all cases. With the (old) video thumbnails where the clickable element has a fixed size, there will be black bars in the UI but these are added dynamically with CSS and are not included in the thumbnail file. With the (new) preview thumbnails from the sprites, we can adjust the size of the element that shows the thumbnails above the video timeline. This element doesn't need a fixed size. IIRC it was already implemented like this before your changes.

> Is it right that for short videos only five ($minThumbnails) thumbnails should be generated? Because the thumbnails preview on the volume overview shows black images for the missing ones at the moment.

Good catch. Let's just take the `videos.thumbnail_count` config value as the new minimum.

>> Maybe use throttle() for the mouse event (e.g. to limit HTTP requests if sprite images could not be found).
>
> What exactly do you have in mind here? My idea was to check the time difference between the mouse movements on the timeline. If the difference is above a certain threshold and the sprite was found, the thumbnail will be shown.
(Remark: I work on a different branch for this feature. So the tasks won't mix up.)

This is about what to do if the sprite is not (yet) found. Either something went completely wrong or the video was not completely processed yet. So the current code retries fetching the sprite if the previous request turned out empty. But if you do that on mousemove, you send a huge amount of HTTP requests in a short time. The current code limits this to a fixed number of retries for each sprite but these can still happen very quickly, especially if you move back and forth across the threshold between two sprites. I was suggesting to improve this by limiting the retry requests to maybe one per second, using the `throttle()` helper function.

lehecht:
@mzur 
My code is ready to be reviewed.

lehecht:
@mzur 
On the images below, you can see how I added the hover time to the thumbnail canvas.
Can you tell me, if you like it or not and what I should change?

Bildschirmfoto vom 2024-07-12 10-49-04
Bildschirmfoto vom 2024-07-12 10-49-21

mzur:
I like this more where the time is not overlaid over the video:
image
Please center the time, give it a little padding and show it at the same size than the time of the video timeline on the left.

app/Jobs/ProcessNewVideo.php
```diff
@@ -143,17 +143,30 @@ public function handleFile($file, $path)
...
+            $tmpDir = "{$tmp}/{$fragment}";
+
+            // Directory for extracted images
+            if (!File::exists($tmpDir)) {
+                File::makeDirectory($tmpDir, 0755, true);
+            }
+
+            if (!$disk->exists($fragment)) {
+                $disk->makeDirectory($fragment);
+            }
```
mzur, line 159: This should not be required. The directories should be automatically created with `put()`.

```suggestion
```

app/Jobs/ProcessNewVideo.php
```diff
@@ -143,17 +143,30 @@ public function handleFile($file, $path)
...
+            // Extract images from video
+            $this->extractImagesfromVideo($path, $this->video->duration, $tmpDir);
+
+            // Generate thumbnails
+            $this->generateVideoThumbnails($disk, $fragment, $tmpDir);
+
+            if (File::exists($tmpDir)) {
+                $parentDir = dirname($fragment, 2);
+                File::deleteDirectory($tmp."/{$parentDir}");
             }
```
mzur, line 170: This must be done in a `finally {}` block because the tmp dir should also be deleted if there is an error.

app/Jobs/ProcessNewVideo.php
```diff
@@ -143,17 +143,30 @@ public function handleFile($file, $path)
...
+
+            // Extract images from video
+            $this->extractImagesfromVideo($path, $this->video->duration, $tmpDir);
+
+            // Generate thumbnails
+            $this->generateVideoThumbnails($disk, $fragment, $tmpDir);
+
+            if (File::exists($tmpDir)) {
+                $parentDir = dirname($fragment, 2);
+                File::deleteDirectory($tmp."/{$parentDir}");
```
mzur, line 169: Only delete `$tmpDir`, otherwise you could delete the temp dirs of other jobs with a similar UUID prefix!

```suggestion
                File::deleteDirectory($tmpDir);
```

app/Jobs/ProcessNewVideo.php
```diff
@@ -215,60 +228,84 @@ protected function getVideoDimensions($url)
...
+                $disk->put("{$fragment}/{$thumbCounter}.{$format}", $bufferedThumb);
+                $thumbCounter += 1;
+            }
 
-        // Sometimes there is one entry too few due to rounding errors.
-        if (count($range) < $count) {
-            $range[] = $end;
-        }
+            if (count($thumbnails) < $thumbnailsPerSprite) {
+                $thumbnails[] = VipsImage::thumbnail($file, $thumbnailWidth, ['height' => $thumbnailHeight, 'size' => 'force']);
```
mzur, line 296: I think the `'size' => 'force'` argument still stretches the thumbnails in an unintended way. Compare the sprite thumbnail (top) to the regular thumbnail (bottom):
0-sprite


The sprite preview should also use the correct aspect ratio like the regular thumbnail and adjust the preview element size accordingly.

app/Jobs/ProcessNewVideo.php
```diff
@@ -215,60 +228,84 @@ protected function getVideoDimensions($url)
...
-        // extract frames from a time code that is equal to 0 or $duration.
-        $step = ($duration - 1) / floatval($count - 1);
-        $start = 0.5;
-        $end = $duration - 0.5;
-        $range = range($start, $end, $step);
+        // Config for sprite thumbs
+        $thumbnailsPerSprite = config('videos.sprites_thumbnails_per_sprite');
+        $thumbnailsPerRow = sqrt($thumbnailsPerSprite);
+        $thumbnailWidth = config('videos.sprites_thumbnail_width');
+        $thumbnailHeight = config('videos.sprites_thumbnail_height');
```
mzur, line 274: I think we could also remove the extra `videos.sprites_thumbnail_width` and height and just use the `thumbnails.width` and height for that.

mzur (commented):
The job is 30% faster now :+1:

lehecht:
@mzur 
My code is ready to be reviewed again.

lehecht:
@mzur 
The code can be reviewed now.

lehecht:
> I like this more where the time is not overlaid over the video. Please center the time, give it a little padding and show it at the same size than the time of the video timeline on the left.

Do you like this?
Bildschirmfoto vom 2024-07-15 10-50-47

mzur (commented):
Looks good now. I checked the relevant todos.

There was still an issue with the size of the displayed element, which I quickly fixed.

mzur:
> Do you like this?

Please use a solid color for the background (i.e. disable transparency). Also the whole element must move a little upwards but I think my recent commit that changed the positioning should already do this automatically. Finally, is the text color white? Please use the same color than the current time on the left. Otherwise it looks good :+1: 

Does it also work if no preview thumbnails is available and how does it look then?

lehecht:
@mzur 
I fixed some bugs including the bug that I told you about some weeks ago. Can you check if everything works now?

lehecht:
@mzur 
> There was still an issue with the size of the displayed element, which I quickly fixed.

Now it can happen that the hover time doesn't fit on the thumbnail any more, because it's too long. What should I do with such cases?

mzur:
> Now it can happen that the hover time doesn't fit on the thumbnail any more, because it's too long. What should I do with such cases?

You mean if the thumbnail is too narrow? Then you can set a CSS `min-width` on the element and also center the (narrower) thumbnail in this case.

lehecht:
@mzur 
> Does it also work if no preview thumbnails is available and how does it look then?


Bildschirmfoto vom 2024-07-16 10-33-09

mzur:
Looks good :+1:

lehecht:
@mzur
> I'm pretty sure that the thumbnail dimensions can also be inferred from the size of the sprites in the JS.

The thumbnail dimensions are computed during the preview now.

mzur (commented):
Very good :+1:

I've also updated the `v-if`/`v-show` TODO above because with the current time in the preview element, it should always use `v-show`. This will require changes to the current behavior because it relies on the actions in the `created()` hook that is only called when the element is newly created (through `v-if`). The settings option will then only control the display of the preview thumbnails, the element with the current time is always shown on mousemove.

In addition, I've added a new TODO to preload sprites because I noticed a lag when a new sprite should be loaded. The preview thumbnail freezes for a bit while the sprite loads. By preloading this can be avoided.

lehecht:
@mzur 
I pushed my changes for the hover time on the thumbnail element. It's ready to be reviewed.

resources/assets/js/videos/components/thumbnailPreview.vue
```diff
@@ -81,10 +99,30 @@ export default {
...
+        },
+        xtext() {
+            return this.canvasWidth / 2;
+        },
+        ytext() {
+            // compute font size in pixel
+            return this.canvasHeight + this.fontSizeInPx + (this.hoverTimeBarHeight - this.fontSizeInPx) / 2;
+        },
+        fontSizeInPx() {
+            return this.fontSize * (72 / 96);
```
mzur, line 117: Is this dependent on the DPI value of the display?

resources/assets/js/videos/components/thumbnailPreview.vue
```diff
@@ -113,6 +151,15 @@ export default {
             // draw the current thumbnail to the canvas
             let context = this.thumbnailCanvas.getContext('2d');
             context.drawImage(this.sprite, sourceX, sourceY, this.thumbnailWidth, this.thumbnailHeight, 0, 0, this.thumbnailCanvas.width, this.thumbnailCanvas.height);
+        
+            // draw the hover time bar
+            context.clearRect(0, this.canvasHeight, this.canvasWidth, this.hoverTimeBarHeight);
+            context.fillStyle = this.hoverTimeStyle['bgColor'];
+            context.fillRect(0, this.canvasHeight, this.canvasWidth, this.hoverTimeBarHeight);
```
mzur, line 158: You could also use a transparent background. Then there is one less hardcoded color.

mzur (commented):
When I very carefully move the mouse vertically over the videoProgress element at the top of the timeline, no thumbnail appears. Maybe the `hover-time-indicator` of the scrollStrip needs a `pointer-events: none`? As soon as I move the mouse even 1 px left or right, the thumbnail appears.

Similarly with a video without sprites, I sometimes get an empty thumbnail element. The hover time is only displayed when the mouse moves left or right (i.e. `mousemove` is fired):

image

lehecht:
> When I very carefully move the mouse vertically over the videoProgress element at the top of the timeline, no thumbnail appears.

Unfortunately, I struggle to reproduce this bug even after trying for 30 minutes. Maybe I do something wrong. Could you show me a video where you produce this bug?

resources/assets/js/videos/components/thumbnailPreview.vue
lehecht: I don't know. I didn't use have another display. So it could be. I use this function just to center the time vertically.

resources/assets/js/videos/components/thumbnailPreview.vue
lehecht: I'm a little confused. What should be transparent? Because in some comment above you said that the hover time background shouldn't be transparent. :sweat_smile: 

Additionally, if I remove line 157, the background becomes greyish. If I set the opacity to zero, the canvas isn't displayed any more.

mzur: The HTML element should have the (non-transparent) background color that is configured via CSS. The canvas element can have a transparent background and only draw the time in solid color. If you remove `fillStyle` in line 157 you also have to remove `fillRect` in line 158. Otherwise it will use the `fillStyle` that is configured for the text below (in the next update) which explains what you see.

resources/assets/js/videos/components/thumbnailPreview.vue
mzur: It looked ok on my screen which is scaled 1.5x. I was just wondering about the numbers. 72 and 96 DPI are common values for screens but phones or other retina displays can have different values.

As this is a constant value, we can probably just skip the computation and take 14.5 * 72 / 96 ~= 11 as `fontSize`.

lehecht:
@mzur 
My code is ready now. Unfortunately, I couldn't reproduce the first bug. Can you tell me, if it's still there?

resources/assets/js/videos/components/currentTime.vue
```diff
@@ -33,11 +31,6 @@ export default {
         currentTimeText() {
             return Vue.filter('videoTime')(this.currentTime);
         },
-        classObject() {
-            return {
-                'current-time--seeking': this.seeking,
-            };
-        },
```
mzur: Please also remove the `--seeking` and `--hover` classes from the CSS file.

resources/assets/js/videos/components/thumbnailPreview.vue
```diff
@@ -104,25 +99,17 @@ export default {
...
-            return { 'font': `bold ${this.fontSize}px Sans-Serif`, 'color': '#cccccc', 'bgColor': '#222' };
+            return { 'font': `bold ${this.fontSize}px Sans-Serif`, 'color': '#cccccc' };
         },
         xtext() {
             return this.canvasWidth / 2;
         },
         ytext() {
-            // compute font size in pixel
             return this.canvasHeight + this.fontSizeInPx + (this.hoverTimeBarHeight - this.fontSizeInPx) / 2;
         },
```
mzur, line 109: These are never used.

resources/assets/js/videos/components/thumbnailPreview.vue
```diff
@@ -104,25 +99,17 @@ export default {
...
-        fontSizeInPx() {
-            return this.fontSize * (72 / 96);
-        }
```
mzur: I didn't see that you had to use both `fontSize` and `fontSizeInPx`, sorry. In this case the computed property is indeed a better solution than hard coding both numbers.

mzur (commented):
I don't see the bug any more. Also, I tweaked the style a little.

lehecht:
@mzur 
The code is ready to be reviewed.

lehecht:
@mzur 
The code is ready to be reviewed.

resources/assets/js/videos/components/thumbnailPreview.vue
```diff
@@ -101,12 +100,9 @@ export default {
...
             return { 'font': `bold ${this.fontSize}px Sans-Serif`, 'color': '#cccccc' };
         },
-        xtext() {
-            return this.canvasWidth / 2;
-        },
-        ytext() {
-            return this.canvasHeight + this.fontSizeInPx + (this.hoverTimeBarHeight - this.fontSizeInPx) / 2;
-        },
+        fontSizeInPx() {
+            return this.fontSize * (72 / 96);
```
mzur, line 104: At least we can save this computation :wink: 
```suggestion
            return this.fontSize * 0.75;
```

mzur (commented):
Looks good.

lehecht:
@mzur 
My code is ready to be reviewed again. I added the preload function.

resources/assets/js/videos/components/thumbnailPreview.vue
```diff
@@ -101,22 +103,96 @@ export default {
...
+                if (spriteUrl in this.preloadedSprites && this.triedUrls[spriteUrl] === 0) {
+                    let onloadFunc = this.sprite.onload;
+                    let onerrFunc = this.sprite.onerror;
+                    this.sprite = this.preloadedSprites[spriteUrl];
+                    onloadFunc.call(this.sprite, new Event('load'));
+                    this.sprite.onload = onloadFunc;
+                    this.sprite.onerror = onerrFunc;
+                } else {
+                    this.sprite.src = spriteUrl;
+                }
```
mzur, line 186: Shouldn't this rather be:
```
if spriteUrl in preloadedSprites
  -> use preloadedSprite
else if triedUrls[spriteUrl] < retryAttempts
  -> try to load sprite
else
  -> give up
```

Currently the preloaded sprite is only used if there were 0 retry attempts?

lehecht: With `this.triedUrls[spriteUrl] === 0` I tried to fix a bug. If you don't use this condition and hover really fast it can happen that the sprite index and sprite url don't belong to the same sprite any more.
I thought to only preload sprites which had no errors, because the `triedUrls` array is reset as soon as the cursor leaves the timeline. I assumed that it is not very likely for a user to keep the cursor on the timeline for a longer period without leaving it at least once.

lehecht: Regarding the structure, you are right. I forgot that the preload methods already check the `this.triedUrls[spriteUrl]` count.

lehecht: Because it was a while ago when I tried to debug this issue, I will take another look at it again and try to fix it.

mzur:
I think the error behavior works as expected now.

I also made changes so the thumbnail is only updated when necessary (instead of any hover event) and fixed the initial display of the thumbnail (without first hover event).

mzur:
Please implement the `ProcessNewVideoTest` so it does not actually process any videos or generate thumbnails. You can split the actions (video processing, thumbnail generation) into different methods and mock these methods in the `ProcessNewVideoStub` of the test. This is how it was done before. In the mocked methods you can record the times for checking and also return fake image data for storage in the test disk.

lehecht:
@mzur 
Is the task "Preload the previous and next sprite for instant transitions" finished?
And should the throttle function still be implemented?

mzur:
The preloading is finished. Throttling is no longer required.

lehecht:
@mzur 
Is it correct that the thumbnail does not disappear any more (i.e. only hover time bar is shown) if it couldn't be found?

mzur:
If no thumbnail exists then only the time should be shown, yes (if that's hat you meant).

lehecht:
Sorry, I mixed up some thoughts. What I meant was that the time is not displayed if a sprite couldn't be found. Only if you go directly to that sprite, then only the time is shown.

lehecht:
> Please implement the `ProcessNewVideoTest` so it does not actually process any videos or generate thumbnails. You can split the actions (video processing, thumbnail generation) into different methods and mock these methods in the `ProcessNewVideoStub` of the test. This is how it was done before. In the mocked methods you can record the times for checking and also return fake image data for storage in the test disk.


What exactly do you mean by mocking the thumbnail creation? Currently, fake images are created which are then converted into thumbnails and sprites. The method `generateVideoThumbnails()` does not any use other methods that could be mocked. So I would need to mock the entire `generateVideoThumbnails()` method, which would make testing useless.

I could split `generateVideoThumbnails()`, so that the thumbnail and sprite generation is moved to a new method, that could be mocked. But maybe you had something completely different in mind?

mzur:
> What I meant was that the time is not displayed if a sprite couldn't be found.

That's ok and will be different once `v-show` is used for the component.

> I could split generateVideoThumbnails(), so that the thumbnail and sprite generation is moved to a new method, that could be mocked. But maybe you had something completely different in mind?

This is more about `extractImagesfromVideo()`, which calls ffmpeg. While this method is mocked in the test, there is an explicit exception in some tests that will actually call ffmpeg. Instead, the call to ffmpeg should be mocked. If these tests are meant to test the actual ffmpeg command, then you can remove the tests. Calling out to ffmpeg is too slow. Instead, test the command manually and once everything works there, never change the command again :wink: 

We'll see if similar mocking is required for `generateVideoThumbnails()` or if the calls to Vips are reasonably fast.

lehecht:
@mzur 
> Please implement the `ProcessNewVideoTest` so it does not actually process any videos

If I use only fake videos, the `handler()` cannot be tested any more, because I had to mock it to accept the fake videos. Should I delete all the tests for the handler?

lehecht:
> That's ok and will be different once `v-show` is used for the component.

I'm a bit confused now. Can you explain how everything should look like when v-show is used?

mzur:
> If I use only fake videos, the `handler()` cannot be tested any more, because I had to mock it to accept the fake videos. Should I delete all the tests for the handler?

Which `handler()` do you mean? Only mock the methods of the class that do calls to FFmpeg or Vips. If there are methods that do something else, too, split the methods so you have methods that *only* call FFmpeg or Vips. These can be mocked in the test stub class.

> I'm a bit confused now. Can you explain how the thumbnail preview should behave when v-show is used?

It's basically [this line](https://github.com/biigle/core/blob/a54b261e2b6a243de70cccdb32eedb5e6ca5d6ea/resources/assets/js/videos/components/scrollStrip.vue#L19). It should be changed to something like:

```diff
- v-if="showThumbPreview"
+ v-if="thumbnailEnabledInSettings"
+ v-show="thumbnailVisibleOnMouseOver"
```
So the `if` is controlled with the settings switch and the `show` is controlled by the mouse over event. Then do any changes in the preview thumbnail code that are required to make it work again (as some parts rely on it being destroyed and created each time but now it will be created only once).

lehecht:
> Which `handler()` do you mean? Only mock the methods of the class that do calls to FFmpeg or Vips. If there are methods that do something else, too, split the methods so you have methods that _only_ call FFmpeg or Vips. These can be mocked in the test stub class.

I don't know how to use fake videos together with [this](https://github.com/biigle/core/blob/a54b261e2b6a243de70cccdb32eedb5e6ca5d6ea/app/Jobs/ProcessNewVideo.php#L112-L178) handler. The ProcessNewVideosTest class also contains [tests](https://github.com/biigle/core/blob/a54b261e2b6a243de70cccdb32eedb5e6ca5d6ea/tests/php/Jobs/ProcessNewVideoTest.php#L129-L248) just for this handler. But because fake videos are not actual videos, all tests for this handler fail. My questions now is what I should do.

lehecht:
> Then do any changes in the preview thumbnail code that are required to make it work again (as some parts rely on it being destroyed and created each time but now it will be created only once).

If I apply your changes, almost everything seems to work again. Except that the missing sprites are still displayed by the last sprite. And which parts of the code rely on being destroyed and created? I couldn't find any. Or maybe I didn't get, what you want  to say?

lehecht:
@mzur 
My code is ready to be reviewed.

mzur (commented):
The code regarding the preview thumbnail is mostly fine (see below). Regarding the test, you removed too much of the actual test methods for my taste, to be honest. As it is difficult to explain how the test should be implemented, I created a PR for the changes here: https://github.com/biigle/core/pull/932 Most of it reverts your commits but here are the actual changes that I meant: https://github.com/biigle/core/pull/932/commits/8f06ab5e61690dc7cab517f54e6b0b1987dfdc9f

As you can see I touched the actual test methods only when necessary, as the tests should work the same than before. I only modified the actual job class and stub to not process actual video files most of the time.

If you see no issues with the changes I will merge the PR into this one.

resources/assets/js/videos/components/scrollStrip.vue
```diff
@@ -16,7 +16,9 @@
                         :clientMouseX="clientMouseX"
                         :scrollstripTop="scrollstripTop"
                         :videoId="videoId"
-                        v-if="showThumbPreview"
+                        :showThumbnails="showThumbnailPreview"
+                        v-if="finishedInitalizingData"
```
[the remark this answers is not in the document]
lehecht: You're right that hover-time shouldn't be used in `finishedInitalizingData()`. However, removing the v-if would complicate things further. The [created](https://github.com/biigle/core/blob/b59c095e1aa74f53c98ee97fbf31a7434ff3aad5/resources/assets/js/videos/components/thumbnailPreview.vue#L297) method calls `updateThumbnailInterval()`, which relies on the duration to set certain constants. If the duration is zero, it leads to a division by zero when calculating the sprite index (NaN).

Additionally, replacing hover-time with duration in `finishedInitalizingData()` prevents the component from being destroyed after leaving the timeline.

Does this approach work for you, or do you still prefer removing the v-if?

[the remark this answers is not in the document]
lehecht: Update: After switching the thumbnail-button off and on again, the sprites are not displayed any more. So something still relies on hover-time in the thumbnailPreview component.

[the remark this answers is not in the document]
mzur: > The [created](https://github.com/biigle/core/blob/b59c095e1aa74f53c98ee97fbf31a7434ff3aad5/resources/assets/js/videos/components/thumbnailPreview.vue#L297) method calls updateThumbnailInterval(), which relies on the duration to set certain constants. If the duration is zero, it leads to a division by zero when calculating the sprite index (NaN).

In which cases can the duration be zero? If the video was not loaded yet? If the duration is zero, skip showing thumbnails in the component. Then add a watcher that updates the constants if the duration changes. We need this anyway when the user switches between two videos.

> Moreover, replacing hover-time with duration in finishedInitalizingData() prevents the component from being destroyed after leaving the timeline.

This is exactly what should happen. The component should not be destroyed, only hidden when the mouse does not hover over the timeline.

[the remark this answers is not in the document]
lehecht: So I should delete v-if completely, right?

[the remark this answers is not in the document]
mzur: Maybe keep it but just for the error/initialization state. Basically:
```js
finishedInitalizingData() {
      return !this.hasError && this.duration > 0;
},
```
The preview component can only work with a duration > 0, so if this is not loaded yet it does not have to be created. But in any other case the component should be created. Wouldn't this also solve the problem with the duration you mentioned? However, you may still need a watcher for the duration if the video changes.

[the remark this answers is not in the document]
lehecht: > However, you may still need a watcher for the duration if the video changes.

I tested this approach earlier without a watcher and it worked. Can you explain what should or could happen to the duration when the video changes?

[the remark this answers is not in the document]
mzur: The duration changes, too :smile: But maybe the duration is briefly set to 0 when a new video loads, so the preview component is destroyed and created again for the new video.

lehecht:
@mzur 
You can review the code now.

mzur (commented):
Looks good, thanks a lot! I'll release it right away and start the processing jobs.

PR #799, opened 2024-03-05
Fix polygon tool bug

lehecht:
When interaction mode is changed, the old and new computed property change, which triggers their corresponding watchers.
Fix bug by using watcher of current interaction mode only.

### DIFF
--- a/resources/assets/js/annotations/components/annotationCanvas/polygonBrushInteraction.vue
+++ b/resources/assets/js/annotations/components/annotationCanvas/polygonBrushInteraction.vue
@@ -111,18 +111,24 @@ export default {
     },
     watch: {
         isUsingPolygonBrush() {
-            this.resetCurrentInteraction();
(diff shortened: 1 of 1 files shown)


mzur (changes requested):
If I just disable one of the tools, the interaction is not reset (i.e. the tool remains active). To reproduce, activate the polygon brush and then press <kbd>Esc</kbd>. The brush is still active.

mzur (commented):
This still doesn't work in all cases, unfortunately. When I now switch from the polygon brush to the magic wand tool (or any other tool), both will be active. You can tinker a little with all the tools before you submit a new version for review, then you will quickly see issues like these yourself.

The video annotation tool does this a little differently, as it has a variable for each interaction that can be added/removed separately. There, everything works correctly.

lehecht:
I tested the code multiple times for images and videos now. I think there shouldn't be bugs any more. Sorry for my hasty approach before.

Issue #433, opened 2022-03-21
video.currentTime is not precise enough to indentify a frame


mzur:
Apparently, `video.currentTime` can be modified/rounded by the browser [for security reasons](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/currentTime#usage_notes). This means that the value returned by `currentTime` is not precise enough to identify the video frame. The following observation confirms this issue: If the video is played and paused and then `video.currentTime = video.currentTime` is set, **the displayed video frame is updated** most of the time! Try it in [this fiddle](https://jsfiddle.net/ek6wa547/).

The issue that frame-accurate handling of videos is not really possible in the browser is discussed [for a long time](https://www.w3.org/2019/Talks/TPAC/frame-accurate-sync). I don't know how we can handle this, yet, but it might become complicated.

mzur:
See [this discussion](https://github.com/cvisionai/tator/discussions/658) for how a switch to frame numbers instead of times could be accomplished.

mzur:
Also related: https://github.com/w3c/media-and-entertainment/issues/4

mzur:
PR #444 implements a workaround where the video frame is updated whenever the video is paused. The updated frame exactly shows the frame corresponding to `currentTime` so annotations placed at this time will always resolve back to the correct frame. The downside is that the video jumps one frame forward or backward whenever the video is paused. Considering the very complex alternatives, I can live with that. I'll leave this open since this issue should be fixed eventually (maybe with updated web APIs?) but I'll lower the priority, as it is no longer pressing.

mzur:
Here is an example how a video could be rendered frame by frame using the Web Codecs API: https://stackoverflow.com/a/32708998/1796523

mzur:
Here is a project that renders videos using the WebCodecs API: https://github.com/aegir-assembly/omni-clip
It uses FFMpeg (WASM) to demux the video files. This could be a solution for our support of many different containers/codecs. WHile the whole of FFMpeg is very large, maybe we can only compile a subset of it that is required for demuxing.

Another thing to consider is that with WebCodecs we can only support remote files with CORS enabled (similar to the image annotation tool).

davidsean:
In order to address some of the issues, I wrote a [community script](https://github.com/davidsean/community-resources/tree/feature/snap_annotations)  that will "snap" time annotations to an exact frame (at least with floating point accuracy).

## video @ 5 fps
This is a simple case, and easy to get a timestamps between frames.
[...]

Issue #262, opened 2020-07-10
Skip forward/back in video


mzur:
Implement keyboard shortcuts to skip forward/back x seconds in the video. The number of seconds should be configurable in the settings. Maybe also add buttons in the second level that appear while hovering over the play button.

mzur:
If we ever get to implement a more sophisticated video playback mechanism (https://github.com/biigle/core/issues/433), we could also offer a way to jump forward/backward frame by frame.

mzur:
This is being worked on here: https://github.com/biigle/core/pull/821

PR #821, opened 2024-05-02
Implement frame-to-frame video navigation

ToukL:
- Uses a hack to navigate frame-to-frame in video, based on [requestVideoFrameCallback](https://developer.mozilla.org/en-US/docs/Web/API/HTMLVideoElement/requestVideoFrameCallback) and [https://github.com/angrycoding/requestVideoFrameCallback-prev-next](https://github.com/angrycoding/requestVideoFrameCallback-prev-next)

WARNING: this works only in Chrome (not in Firefox). Not tested in other navigators.

- Adds a 'Jump step' setting to directly jump a certain amount of seconds backward and forward

Changes video navigation buttons:

- step-backward and step-forward are used to go to previous and next frames in video
- add double-backward and double-forward buttons for Jump step setting
- add chevron-left and chevron-right to go to previous/next video of volume

### DIFF
--- a/resources/assets/js/videos/components/settingsTab.vue
+++ b/resources/assets/js/videos/components/settingsTab.vue
@@ -15,6 +15,7 @@ export default {
                 'showLabelTooltip',
                 'showMousePosition',
                 'showProgressIndicator',
+                'jumpStep',
             ],
             annotationOpacity: 1,
             showMinimap: true,
@@ -22,6 +23,7 @@ export default {
             showLabelTooltip: false,
             showMousePosition: false,
             playbackRate: 1.0,
+            jumpStep: 5.0,
             showProgressIndicator: true,
         };
     },
@@ -81,6 +83,11 @@ export default {
             if (!isNaN(value)) {
                 this.$emit('update', 'playbackRate', value);
             }
+        },
+        jumpStep(value) {
+            value = parseFloat(value);
+            this.$emit('update', 'jumpStep', value);
+            Settings.set('jumpStep', value);
         },
         showProgressIndicator(show) {
             this.$emit('update', 'showProgressIndicator', show);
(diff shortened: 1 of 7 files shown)


mzur:
This is an interesting approach! I need to test this thoroughly with different videos because I had so much trouble with this in the past. There are also a couple of things that could be streamlined for the final implementation but I'll get to that once I tested it a bit.

However, this also implements the "jump forward/backward by a fixed number of seconds", right (https://github.com/biigle/core/issues/262)? Could you please split this from the "jump frame-by-frame" changes and send both in separate PRs? The jumping by seconds can be merged much sooner then.

mzur:
For the record: I mentioned the workaround in the [W3C discussion](https://github.com/w3c/media-and-entertainment/issues/4) I follow and others say that it might not produce perfect results in all cases. [This article](https://web.dev/articles/requestvideoframecallback-rvfc) recommends the [WebCodecs API](https://developer.mozilla.org/docs/Web/API/WebCodecs_API) for frame-accurate applications. [I had a look at this](https://github.com/biigle/core/issues/433#issuecomment-1482788582), too, but as the linked StackOverflow answer mentions, it only works if the media is demuxed. And I think there is no easy solution for that if we still allow many different video file formats in BIIGLE.

ToukL:
> This is an interesting approach! I need to test this thoroughly with different videos because I had so much trouble with this in the past. There are also a couple of things that could be streamlined for the final implementation but I'll get to that once I tested it a bit.
> 
> However, this also implements the "jump forward/backward by a fixed number of seconds", right (#262)? Could you please split this from the "jump frame-by-frame" changes and send both in separate PRs? The jumping by seconds can be merged much sooner then.

ToukL:
I did split the PR in two.

I know it's just a workaround that needs more testing and cleaning, if you have time and ideas for that it would be great ! I implemented it like that because we need the frame-by-frame navigation, even if not perfect fro now. I'm open to any suggestion though !

mzur:
Thanks! We'll continue in the new PRs.

PR #822, opened 2024-05-02
Implement 'Jump by seconds' forward/backward in video

ToukL:
Splitted from [#821](https://github.com/biigle/core/pull/821)

Implements a "jump by seconds" settings (settable by user) discussed here [#262](https://github.com/biigle/core/issues/262)

### DIFF
--- a/resources/assets/js/videos/components/videoScreen.vue
+++ b/resources/assets/js/videos/components/videoScreen.vue
@@ -11,12 +11,19 @@
             :position="mousePosition"
             ></label-tooltip>
         <div class="controls">
-            <div class="btn-group">
+            <div class="btn-group"
+                v-if="showPrevNext">
                  <control-button
-                    v-if="showPrevNext"
-                    icon="fa-step-backward"
-                    title="Previous video 𝗟𝗲𝗳𝘁 𝗮𝗿𝗿𝗼𝘄"
+                    icon="fa-chevron-left"
+                    title="Previous video"
                     @click="emitPrevious"
+                    ></control-button>
+            </div>     
+            <div class="btn-group">
+                <control-button
+                    icon="fa-backward"
+                    title="Rewind video by jump step"
+                    @click="jumpBackward"
                     ></control-button>
                 <control-button
                     v-if="playing"
@@ -33,9 +40,16 @@
                     @click="play"
                     ></control-button>
                 <control-button
-                    v-if="showPrevNext"
-                    icon="fa-step-forward"
-                    title="Next video 𝗥𝗶𝗴𝗵𝘁 𝗮𝗿𝗿𝗼𝘄"
+                    icon="fa-forward"
+                    title="Advance video by jump step"
+                    @click="jumpForward"
+                    ></control-button>
+            </div>
+            <div class="btn-group"
+                v-if="showPrevNext">
+                 <control-button
+                    icon="fa-chevron-right"
+                    title="Next video"
                     @click="emitNext"
                     ></control-button>
             </div>
@@ -285,6 +299,10 @@ export default {
         autoplayDraw: {
             type: Number,
             default: 0,
+        },
+        jumpStep: {
+            type: Number,
+            default: 5.0,
         },
         canAdd: {
             type: Boolean,
--- a/resources/assets/js/videos/components/videoScreen/videoPlayback.vue
+++ b/resources/assets/js/videos/components/videoScreen/videoPlayback.vue
@@ -153,6 +153,33 @@ export default {
         handleSeeked() {
             this.renderVideo(true);
         },
+        // 3 next methods are a workaround to get previous and next frames, adapted from here: https://github.com/angrycoding/requestVideoFrameCallback-prev-next/tree/main
+        frameInfoCallback() {
+            let promise = new Vue.Promise((resolve, reject) => {
+                if ("requestVideoFrameCallback" in HTMLVideoElement.prototype) {
+                    this.video.requestVideoFrameCallback((now, metadata) => {
+                        resolve(metadata);
+                    })
+                }
+                else {
+                    reject("Your browser does not support requestVideoFrameCallback().");
+                }
+            })
+            return promise;
+        },
+        // Methods to jump back and forward in video. Step is given by parameter jumpStep.
+        async jumpBackward() {
+            if (this.video.currentTime > 0 && this.jumpStep > 0) {
+                this.video.currentTime -= this.jumpStep;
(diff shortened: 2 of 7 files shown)


resources/assets/js/videos/components/videoScreen/videoPlayback.vue
```diff
@@ -153,6 +153,33 @@ export default {
...
+                this.video.currentTime -= this.jumpStep;
+                await this.frameInfoCallback();
+            }
+        },
+        async jumpForward() {
+            if (!this.video.ended && this.jumpStep > 0) {
+                this.video.currentTime += this.jumpStep;
+                await this.frameInfoCallback();
+            }
+        },
```
mzur, line 182: Why do you need the `frameInfoCallback` here? The video time should change by setting `currentTime` only.

ToukL: Yes you are right it's useless here, I removed it.

mzur (changes requested):
It works well, thank you! I have a few comments on the interface:

- The next/previous video buttons should still belong to the same `btn-group` than the "play" button.
- The next/previous buttons should still have the `fa-step-backward` or `fa-step-forward` icons because these are the icons the image annotation tool uses, too. Personally, I like the chevron icons better for next/previous, but the old icons were there forever and users are used to them so we should stick to them, too.
- Users should be able to set the jump step to 0 to disable this feature. If it is 0, the jump buttons should be hidden, too. This should be mentioned in the description and the manual (see below)
- There should be a keyboard shortcut for the jump buttons. Maybe <kbd>Ctrl</kbd>+left/right arrow?
- The feature should also be mentioned in [the manual article](https://github.com/biigle/core/blob/833d15d94a95a8865ad697c796a75b301de0bedb/resources/views/manual/tutorials/videos/sidebar.blade.php#L45).

mzur:
While looking at your other PR, I see now why you changed the icons for the next/previous video. I still think the old icons should keep their function, otherwise users will be confused.

Maybe we can use [`fa-caret-square-right`](https://fontawesome.com/v5/icons/caret-square-right?f=classic&s=solid) (or left) for the "jump by frame" buttons?

ToukL:
> It works well, thank you! I have a few comments on the interface:
> 
> * The next/previous video buttons should still belong to the same `btn-group` than the "play" button.
> * The next/previous buttons should still have the `fa-step-backward` or `fa-step-forward` icons because these are the icons the image annotation tool uses, too. Personally, I like the chevron icons better for next/previous, but the old icons were there forever and users are used to them so we should stick to them, too.
> * Users should be able to set the jump step to 0 to disable this feature. If it is 0, the jump buttons should be hidden, too. This should be mentioned in the description and the manual (see below)
> * There should be a keyboard shortcut for the jump buttons. Maybe Ctrl+left/right arrow?
> * The feature should also be mentioned in [the manual article](https://github.com/biigle/core/blob/833d15d94a95a8865ad697c796a75b301de0bedb/resources/views/manual/tutorials/videos/sidebar.blade.php#L45).

Thank you for the feedback. Yes I suspected  that the change of button would not be accepted in the official instance, I will rollback to using `fa-step-backward` and `fa-step-forward` for changing videos and use the `fa-caret-square-right` for the frame as you suggested. But do you think we can keep the left/right arrow shortcut for the prev/next frame ? I feel that if we manage to implement that feature it would be more intuitive to use keyboard arrows to change frames and maybe use another shortcut for changing videos ? Maybe `shift`+left/right arrow ? I can leave the old shortcut for this PR and do the change in the other one that deals with frame navigation.

I'll do all the other changes as suggested. Thanks again !

mzur:
> I can leave the old shortcut for this PR and do the change in the other one that deals with frame navigation.

Please do. I'll start my review of your other PR now and we can deal with this question there.

mzur:
While I reviewed you other PR, something else came to my attention. Rather than setting `currentTime` directly in videoPlayback, you should emit an event that calls the `seek()` method of the videoContainer. This will show an appropriate loading state. Also, you can then disable the jump by seconds button while `this.seeking===true` in the videoContainer. The `this.seeking` state is also available in the videoScreen component and in videoPlayback.

mzur:
And another thing when you implement the keyboard shortcuts: I always add the shortcut in bold text to the title attribute of the buttons. I use [this tool](https://qaz.wtf/u/convert.cgi?text=Right+arrow) for that and choose the "Math sans bold" text.

ToukL:
> While I reviewed you other PR, something else came to my attention. Rather than setting `currentTime` directly in videoPlayback, you should emit an event that calls the `seek()` method of the videoContainer. This will show an appropriate loading state. Also, you can then disable the jump by seconds button while `this.seeking===true` in the videoContainer. The `this.seeking` state is also available in the videoScreen component and in videoPlayback.

Sure I will do that too !

resources/assets/js/videos/components/videoScreen.vue
```diff
@@ -11,18 +11,17 @@
...
+                <control-button
+                    v-if="showPrevNext"
+                    icon="fa-step-backward"
+                    title="Previous video 𝗟𝗲𝗳𝘁 𝗮𝗿𝗿𝗼𝘄"
                     @click="emitPrevious"
                     ></control-button>
-            </div>     
-            <div class="btn-group">
                 <control-button
+                    v-if="!seeking && jumpStep!=0"
```
mzur, line 22: You should disable the button instead of hiding it while `seeking`, otherwise the button toolbar jumps around each time the feature is used:

```suggestion
                    v-if="jumpStep!=0"
                    :disabled="seeking"
```

resources/assets/js/videos/components/videoScreen.vue
```diff
@@ -40,16 +39,15 @@
                     @click="play"
                     ></control-button>
                 <control-button
+                    v-if="!seeking && jumpStep!=0"
```
mzur, line 42: Same than above:
```suggestion
                    v-if="jumpStep!=0"
                    :disabled="seeking"
```

mzur (changes requested):
Great, thank you! I only have a small comment about the button state while `seeking`.

Also I had the idea that we could maybe make the button `title` dynamic, so it reads like "Rewind video by 5 s 𝗖𝘁𝗿𝗹+𝗟𝗲𝗳𝘁 𝗮𝗿𝗿𝗼𝘄". What do you think? This could be implemented with a computed property.

mzur (approved):
Perfect, thank you!

PR #823, opened 2024-05-02
Implements 'Jump frame by frame' in video

ToukL:
Splitted from here [#821](https://github.com/biigle/core/pull/821)

Workaround to navigate frame by frame (forward and backward). Only works in Chrome for now, needs more tests.

### DIFF
--- a/resources/assets/js/videos/components/videoScreen.vue
+++ b/resources/assets/js/videos/components/videoScreen.vue
@@ -11,13 +11,19 @@
             :position="mousePosition"
             ></label-tooltip>
         <div class="controls">
-            <div class="btn-group">
+            <div v-if="showPrevNext" class="btn-group">
                 <control-button
-                    v-if="showPrevNext"
                     icon="fa-step-backward"
-                    :title="jumpByFrameEnabled ? 'Previous video 𝗦𝗵𝗶𝗳𝘁+𝗟𝗲𝗳𝘁 𝗮𝗿𝗿𝗼𝘄' : 'Previous video 𝗟𝗲𝗳𝘁 𝗮𝗿𝗿𝗼𝘄'"
+                    :title="enableJumpByFrame ? 'Previous video 𝗦𝗵𝗶𝗳𝘁+𝗟𝗲𝗳𝘁 𝗮𝗿𝗿𝗼𝘄' : 'Previous video 𝗟𝗲𝗳𝘁 𝗮𝗿𝗿𝗼𝘄'"
                     @click="emitPrevious"
                     ></control-button>
+                <control-button
+                    icon="fa-step-forward"
+                    :title="enableJumpByFrame ? 'Next video 𝗦𝗵𝗶𝗳𝘁+𝗥𝗶𝗴𝗵𝘁 𝗮𝗿𝗿𝗼𝘄' : 'Next video 𝗥𝗶𝗴𝗵𝘁 𝗮𝗿𝗿𝗼𝘄'"
+                    @click="emitNext"
+                    ></control-button>
+            </div>
+            <div class="btn-group">
                 <control-button
                     v-if="enableJumpByFrame"
                     :disabled="seeking"
@@ -59,12 +65,6 @@
                     icon="fa-forward"
                     :title="jumpForwardMessage"
                     @click="jumpForward"
-                    ></control-button>
-                <control-button
-                    v-if="showPrevNext"
-                    icon="fa-step-forward"
-                    :title="jumpByFrameEnabled ? 'Next video 𝗦𝗵𝗶𝗳𝘁+𝗥𝗶𝗴𝗵𝘁 𝗮𝗿𝗿𝗼𝘄' : 'Next video 𝗥𝗶𝗴𝗵𝘁 𝗮𝗿𝗿𝗼𝘄'"
-                    @click="emitNext"
                     ></control-button>
             </div>
             <div v-if="canAdd" class="btn-group">
--- a/resources/assets/js/videos/components/videoScreen.vue
+++ b/resources/assets/js/videos/components/videoScreen.vue
@@ -25,18 +25,18 @@
             </div>
             <div class="btn-group">
                 <control-button
-                    v-if="enableJumpByFrame"
-                    :disabled="seeking"
-                    icon="fa-caret-square-left"
-                    title="Previous frame 𝗟𝗲𝗳𝘁 𝗮𝗿𝗿𝗼𝘄"
-                    v-on:click="emitPreviousFrame"
-                    ></control-button>
-                <control-button
                     v-if="jumpStep!=0"
                     :disabled="seeking"
                     icon="fa-backward"
                     :title="jumpBackwardMessage"
                     @click="jumpBackward"
+                    ></control-button>
+                <control-button
+                    v-if="enableJumpByFrame"
+                    :disabled="seeking"
+                    icon="fa-caret-square-left"
+                    title="Previous frame 𝗟𝗲𝗳𝘁 𝗮𝗿𝗿𝗼𝘄"
+                    v-on:click="emitPreviousFrame"
                     ></control-button>
                 <control-button
                     v-if="playing"
--- a/resources/assets/js/videos/components/videoScreen.vue
+++ b/resources/assets/js/videos/components/videoScreen.vue
@@ -11,18 +11,16 @@
             :position="mousePosition"
             ></label-tooltip>
         <div class="controls">
-            <div class="btn-group"
-                v-if="showPrevNext">
-                 <control-button
-                    icon="fa-chevron-left"
-                    title="Previous video"
-                    @click="emitPrevious"
-                    ></control-button>
-            </div>     
             <div class="btn-group">
                 <control-button
-                    v-if="enableJumpByFrame"
+                    v-if="showPrevNext"
                     icon="fa-step-backward"
+                    :title="jumpByFrameEnabled ? 'Previous video 𝗦𝗵𝗶𝗳𝘁+𝗟𝗲𝗳𝘁 𝗮𝗿𝗿𝗼𝘄' : 'Previous video 𝗟𝗲𝗳𝘁 𝗮𝗿𝗿𝗼𝘄'"
+                    @click="emitPrevious"
+                    ></control-button>
+                <control-button
+                    v-if="enableJumpByFrame"
+                    icon="fa-caret-square-left"
                     title="Previous frame 𝗟𝗲𝗳𝘁 𝗮𝗿𝗿𝗼𝘄"
                     v-on:click="showPreviousFrame"
                     ></control-button>
@@ -42,16 +40,14 @@
                     ></control-button>
                 <control-button
                     v-if="enableJumpByFrame"
-                    icon="fa-step-forward"
-                    title="Next frame Right 𝗮𝗿𝗿𝗼𝘄"
+                    icon="fa-caret-square-right"
+                    title="Next frame 𝗥𝗶𝗴𝗵𝘁 𝗮𝗿𝗿𝗼𝘄"
                     v-on:click="showNextFrame"
                     ></control-button>
-            </div>
-            <div class="btn-group"
-                v-if="showPrevNext">
-                 <control-button
-                    icon="fa-chevron-right"
-                    title="Next video"
+                <control-button
+                    v-if="showPrevNext"
+                    icon="fa-step-forward"
+                    :title="jumpByFrameEnabled ? 'Next video 𝗦𝗵𝗶𝗳𝘁+𝗥𝗶𝗴𝗵𝘁 𝗮𝗿𝗿𝗼𝘄' : 'Next video 𝗥𝗶𝗴𝗵𝘁 𝗮𝗿𝗿𝗼𝘄'"
                     @click="emitNext"
                     ></control-button>
             </div>
@@ -408,6 +404,9 @@ export default {
         disableJobTracking() {
             return this.reachedTrackedAnnotationLimit;
         },
+        jumpByFrameEnabled() {
+            return this.enableJumpByFrame;
+        }
     },
     methods: {
         createMap() {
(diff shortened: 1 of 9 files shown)


mzur:
Reference to my earlier response to this: https://github.com/biigle/core/pull/821#issuecomment-2090471105

mzur (commented):
I think this still includes the "jump by seconds" code.

ToukL:
> I think this still includes the "jump by seconds" code.

Ah yes I based this branch on "jump-by-seconds" but I can rebase on master, it would be cleaner indeed

mzur (changes requested):
I see that this could be useful, although it may not be correct all the time. Since this is an "experimental" feature then, I'd rather make it opt-in in the annotation tool settings. This way, we could solve four issues at once:

1. Explain to the users that the method may not always be correct (in the manual).
2. Show non-Chrome users that the feature exists but is not available for them (by being disabled).
3. Allow users who don't need the feature to hide some of the new arrow buttons that were now introduced to clean up their user interface.
4. Set left/right arrow as keyboard shortcut to jump frames only if the feature is disabled. If the feature is disabled, the old behavior is still active (arrows switch videos). If the feature is enabled <kbd>Shift</kbd>+arrows could be used to switch videos.

So if you agree with this, here are my comments:

- Please check for the availability of `requestVideoFrameCallback` when the video annotation tool loads. If it is not available, the feature is disabled and the settings button (see below) is disabled, too, so it cannot be clicked. This way you don't have to check for the availability of the method on each frame jump any more.
- Add a settings button that enables the feature. It should be disabled by default and labeled as "experimental" with a link to the manual. If the feature is disabled, the next/previous buttons to jump frames should be hidden.
- Similar to the other PR, all next/previous buttons should belong to the same `btn-group`.
- As discussed, the jump by frame buttons should use a different icon.
- Similar to the other PR, add a section to the manual article that explains the feature, why it is "experimental" (because it may not always be correct) and that it is only available in Chrome.
- You should prevent "double clicks" on the jump by frame buttons, i.e. while the next frame is seeked, the buttons should be disabled. Ideally, you should emit an event from the videoScreen component that calls the `seek()` method of the videoContainer. This will set `this.seeking=true` which shows a little loading spinner. While `this.seeking=true`, the buttons are disabled. The this.seeking state is also available in the videoScreen component and in videoPlayback. However, this is not so easy here, as the video is "secretly" seeked to find the next frame... Maybe you have to add an extra method to the videoContainer to control `this.seeking` with events from the videoScreen component.

ToukL:
> * Please check for the availability of `requestVideoFrameCallback` when the video annotation tool loads. If it is not available, the feature is disabled and the settings button (see below) is disabled, too, so it cannot be clicked.

When you say "when the video annotation tool loads" is it in the `loadVideo()` method of the [videoContainer](https://github.com/biigle/core/blob/833d15d94a95a8865ad697c796a75b301de0bedb/resources/assets/js/videos/videoContainer.vue) or elsewhere (in the views part maybe) ?

mzur:
The appropriate location would be in the [`created()`](https://github.com/biigle/core/blob/833d15d94a95a8865ad697c796a75b301de0bedb/resources/assets/js/videos/videoContainer.vue#L680) method of the videoContainer. You can set a new attribute (e.g. `this.supportsJumpByFrame`) there and pass this on to the videoScreen and settings tab as props.

mzur:
Please re-request my review once you are finished with all your changes so I know that I should have a look :wink:

ToukL:
> Please re-request my review once you are finished with all your changes so I know that I should have a look 😉

I just did (sorry I was a bit busy these days so it took me a while...)

I am not absolutely sure for the last point (the call to seek method to prevent double clicks). I only added events in VideoScreen that call the `seek()` method if `this.seeking=false`, according to my tests it seems to catch what we want without having to add an extra method in VideoContainer but maybe I'm missing something, tell me what you think :)

resources/assets/js/videos/components/videoScreen.vue
```diff
@@ -402,6 +406,9 @@ export default {
         disableJobTracking() {
             return this.reachedTrackedAnnotationLimit;
         },
+        jumpByFrameEnabled() {
+            return this.enableJumpByFrame;
+        }
```
mzur, line 411: This seems unnecessary. Just use `enableJumpByFrame` directly.

resources/assets/js/videos/components/videoScreen.vue
```diff
@@ -513,6 +534,16 @@ export default {
...
+        emitPreviousFrame() {
+            this.$emit('seek', this.video.currentTime, true);
+            if(!this.seeking)
+                this.showPreviousFrame();
+         },
+        emitNextFrame() {
+            this.$emit('seek', this.video.currentTime, true);
+            if(!this.seeking)
+                this.showNextFrame();
+        },
```
mzur, line 546: This should better be moved to videoPlayback.vue because all the logic is there.

mzur: Also this seems to work (seeking state is active until the new frame is shown) but I don't really understand how it works. I would expect this to show the seeking state, then finish and wait a while until the new frame is found and finally show the new frame. But the seeking state finishes at the same time the new frame is shown...

If we don't find out why this works, we should at least add a comment why the seemingly unnecessary `$emit('seek')` is there.

resources/assets/js/videos/components/videoScreen.vue
```diff
@@ -540,6 +571,24 @@ export default {
...
+                Keyboard.on('ArrowRight', this.emitNextFrame, 0, this.listenerSet);
+                Keyboard.on('ArrowLeft', this.emitPreviousFrame, 0, this.listenerSet);
+            }
+            else {
+                Keyboard.off('Shift+ArrowRight', this.emitNext, 0, this.listenerSet);
+                Keyboard.off('Shift+ArrowLeft', this.emitPrevious, 0, this.listenerSet);
+                Keyboard.off('ArrowRight', this.emitNextFrame, 0, this.listenerSet);
+                Keyboard.off('ArrowLeft', this.emitPreviousFrame, 0, this.listenerSet);
+                Keyboard.on('ArrowRight', this.emitNext, 0, this.listenerSet);
+                Keyboard.on('ArrowLeft', this.emitPrevious, 0, this.listenerSet);
```
mzur, line 589:
```suggestion
                Keyboard.on('ArrowLeft', this.emitPrevious, 0, this.listenerSet);
```

resources/assets/js/videos/components/videoScreen.vue
```diff
@@ -540,6 +571,24 @@ export default {
...
+            }
+            else {
+                Keyboard.off('Shift+ArrowRight', this.emitNext, 0, this.listenerSet);
+                Keyboard.off('Shift+ArrowLeft', this.emitPrevious, 0, this.listenerSet);
+                Keyboard.off('ArrowRight', this.emitNextFrame, 0, this.listenerSet);
+                Keyboard.off('ArrowLeft', this.emitPreviousFrame, 0, this.listenerSet);
+                Keyboard.on('ArrowRight', this.emitNext, 0, this.listenerSet);
+                Keyboard.on('ArrowLeft', this.emitPrevious, 0, this.listenerSet);                
+            }
+        },
```
mzur, line 591: This only works because `jumpByFrameEnabled` is `false` initially and set to `true` later. If it was `true` initially, the keyboard shortcuts to jump by frames would not be enabled. I suggest that you move this to a separate method. Then you call the method both in the `created()` hook (delete the arrow shortcuts there) and in the `jumpByFrameEnabled` watcher.

mzur (changes requested):
It does what I wanted although I don't understand how :smile: 

Can you please merge the new `master` and resolve the conflicts? I also decided that the "previous/next video" buttons should now be moved to a separate `btn-group` at the left like this:
```html
<div v-if="showPrevNext" class="btn-group">
    <control-button
        icon="fa-step-backward"
        :title="jumpByFrameEnabled ? 'Previous video 𝗦𝗵𝗶𝗳𝘁+𝗟𝗲𝗳𝘁 𝗮𝗿𝗿𝗼𝘄' : 'Previous video 𝗟𝗲𝗳𝘁 𝗮𝗿𝗿𝗼𝘄'"
        @click="emitPrevious"
        ></control-button>
    <control-button
        icon="fa-step-forward"
        :title="jumpByFrameEnabled ? 'Next video 𝗦𝗵𝗶𝗳𝘁+𝗥𝗶𝗴𝗵𝘁 𝗮𝗿𝗿𝗼𝘄' : 'Next video 𝗥𝗶𝗴𝗵𝘁 𝗮𝗿𝗿𝗼𝘄'"
        @click="emitNext"
        ></control-button>
</div>
<div class="btn-group">
    <control-button
        v-if="enableJumpByFrame"
        :disabled="seeking"
        icon="fa-caret-square-left"
        title="Previous frame 𝗟𝗲𝗳𝘁 𝗮𝗿𝗿𝗼𝘄"
        v-on:click="emitPreviousFrame"
        ></control-button>
...
```
Then there is a separate group to switch videos and a group to control the video playback/jumping.

resources/assets/js/videos/components/videoScreen.vue
ToukL: I'm a bit lost on this one... It seems like if we trigger the previous/next frame several time quickly (by keeping the keys down for example) with emitting this seek event, at some point the timeline freeze (wheel is locked in spinning mode and we can't change time anymore). However I tried calling the seek() method directly from VideoPlayback but then the wheel doesn't appear, I suppose this is because the videoTimeline is in charge of it, catching the seek event ? 
I don't have a solution right now, what do you think is best : showing the wheel at risk of freezing timeline or don't and setting `this.seeking` by hand ?

ToukL:
Hello Martin,
I re-requested the review but there is still an issue with the `seek` event that may freeze if the user click too many times (I don't know if you saw my comment above)

resources/views/manual/tutorials/videos/sidebar.blade.php
```diff
@@ -75,5 +75,11 @@
         <p>
             The mouse position switch controls the display of an additional map overlay that shows the current position of the cursor on the video in pixels.
         </p>
+
+        <p>
+            <a name="jump-by-frame"></a> The jump by frame switch allows you to navigate frame by frame (forward and backward) in the video. Note that this is an experimental feature as it is only available in Chrome and may not always give the right frame, so please use it with caution.
+            <br>When the feature is enabled, new buttons will appear in the navigation bar and keyboard shortcuts are modified: use <button class="btn btn-default btn-xs"><i class="fa fa-arrow-left"></i></i></button> / <button class="btn btn-default btn-xs"><i class="fa fa-arrow-right"></i></i></button> to switch frames and 𝗦𝗵𝗶𝗳𝘁 + <button class="btn btn-default btn-xs"><i class="fa fa-arrow-left"></i></i></button> / <button class="btn btn-default btn-xs"><i class="fa fa-arrow-right"></i></i></button> to switch videos. When the feature is disabled, use old shortcuts to switch videos.
```
mzur, line 81: Please use `<kbd>` instead of `<button>` because the manual only uses buttons in the text when there are actual buttons to press in the UI.

mzur (changes requested):

resources/assets/js/videos/components/videoScreen/videoPlayback.vue
```diff
@@ -24,6 +24,8 @@ export default {
             // Allow a maximum of 100x magnification. More cannot be represented in the
             // URL parameters.
             minResolution: 0.01,
+            // parameter tracking seeking state specific for frame jump, needed because looking for seeking directly leads to error
+            seekingFrame: this.seeking,
```
mzur, line 28: Can you please explain why you can't use `this.seeking` directly or what error would be caused? Also I don't think you can use `this.seeking` here already.

resources/assets/js/videos/components/videoScreen/videoPlayback.vue
```diff
@@ -153,7 +155,23 @@ export default {
         handleSeeked() {
             this.renderVideo(true);
         },
-        // 3 next methods are a workaround to get previous and next frames, adapted from here: https://github.com/angrycoding/requestVideoFrameCallback-prev-next/tree/main
+        // 5 next methods are a workaround to get previous and next frames, adapted from here: https://github.com/angrycoding/requestVideoFrameCallback-prev-next/tree/main
+        async emitPreviousFrame() {
+            if(this.video.currentTime == 0 || this.seekingFrame) return;
```
mzur, line 160: I can replace `this.seekingFrame` with `this.seeking` here and see no difference.

mzur (commented):
We are getting there, thanks! I've tweaked the manual articles a little bit. Also, now I observe the following behavior: When I keep the left/right arrow key pressed, the seeking starts and stops continuously but the video display is only updated once I release the key. It seems like it actually seeks multiple frames and then renders the final result. Before, I could see the rendered frame updating on each step while keeping the key pressed.

resources/assets/js/videos/components/videoScreen/videoPlayback.vue
ToukL: Odly it seems that we don't get the same behaviour: when I hold the keys pressed I can see the frames rendering as seeking goes along... Also if I use `this.seeking` below instead of `seekingFrame` and hold the keys, the spinning wheel stays stuck at some point, as if we never leave the seeking state.

I'm running my tests on a local address on my pc that I open in Chrome v125.0.6422.144 (Windows 64 bits)
For the projet installation I followed these instructions: [https://github.com/biigle/core/blob/master/DEVELOPING.md](url)

ToukL: Yes I still get the same behavior...

I tried on chromium on WSL and I am able to see the same behavior as you, so it is indeed a linux vs windows difference.

At this point I have no better idea than warning windows users it might get stuck if using keyboard and that they may have to refresh the page. What do you think ?

mzur: Ok if this is a Windows issue and it works with your `seekingFrame` then we can keep it that way!

I also now checked if you can use `this.seeking` in `data()` like that and you can! So this conversation is resolved.

resources/assets/js/videos/videoContainer.vue
```diff
@@ -204,6 +204,28 @@ export default {
...
+            if (!this.seeking) {
+                return Vue.Promise.resolve();
+            }
+            let promise = new Vue.Promise((resolve, reject) => {
+                this.video.addEventListener('seeked', resolve);
+                this.video.addEventListener('error', reject);
+            });
+            this.seeking = false;
+            return promise;
+        },
```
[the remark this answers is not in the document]
ToukL: The `stopSeeking()` call in emitPrevious/NextFrame was actually useless as, after the frame was seeked, a `seeked` event was emitted calling `handleVideoSeeked()` anyway...

mzur (approved):
Fantastic, thanks! May I mention you and your project (called FUTURE-OBS, right?) in the BIIGLE newsletter when I announce the new features? Is there a project website that I can link?

Issue #797, opened 2024-03-04
Favourite label key icon


mzur:
Show a key icon similar to <kbd>1</kbd>, <kbd>2</kbd>, <kbd>3</kbd> etc. next to the name of a favourite label of the label trees component. This could look similar to BIIGLE.party (bottom right of each label):

![image](https://github.com/biigle/core/assets/2457311/2d96990a-05eb-4c16-9a47-c3dd27cf959e)

mzur:
This is how it could look like in the end (with suitable colors and numbers):
![Screenshot from 2024-05-24 14-25-49](https://github.com/biigle/core/assets/2457311/70806e1f-52fb-4c55-b433-17bb7e0b3061)

If the keyboard icon, number and star is too much, we could also try to squeeze the number into the star. But we still need the click action on the star to remove the favourite again.

PR #840, opened 2024-05-29
Favourite label key icon #797 _ first attempt

aneust:
Added keyboard icon and corresponding key to favorite label.

### DIFF
--- a/resources/assets/js/label-trees/components/labelTreeLabel.vue
+++ b/resources/assets/js/label-trees/components/labelTreeLabel.vue
@@ -13,11 +13,11 @@
             </span>
             <span class="label-tree-label__buttons">
                 <span v-if="showFavouriteShortcuts" class="text-muted label-tree-label_position">
-                            <span class="fa fa-keyboard" aria-hidden="true" title=""></span>
-                            <span v-text="actualPosition"></span>
+                    <span class="fa fa-keyboard" aria-hidden="true"></span>
+                    <span v-text="actualPosition"></span>
                 </span>
                 <button v-if="showFavourites" type="button" class="label-tree-label__favourite" :class="favouriteClass" @click.stop="toggleFavourite" :title="favouriteTitle">
-                    <span class="fa fa-star" aria-hidden="true" title=""></span>
+                    <span class="fa fa-star" aria-hidden="true"></span>
                 </button>
                 <span if="editable">
                     <button v-show="showEditButton" :title="editTitle" @click.stop="editThis" class="btn btn-default btn-xs"><span aria-hidden="true" class="fa fa-pencil-alt"></span></button>
--- a/resources/assets/js/label-trees/components/labelTreeLabel.vue
+++ b/resources/assets/js/label-trees/components/labelTreeLabel.vue
@@ -12,7 +12,7 @@
                 <span v-text="label.name" @click.stop="toggleSelect" @mouseenter="dontHover"></span>
             </span>
             <span class="label-tree-label__buttons">
-                <span v-if="showFavouriteShortcuts" class="text-muted" :class="label-tree-label_position">
+                <span v-if="showFavouriteShortcuts" class="text-muted label-tree-label_position">
                             <span class="fa fa-keyboard" aria-hidden="true" title=""></span>
                             <span v-text="actualPosition"></span>
                 </span>
--- a/resources/assets/js/label-trees/components/labelTreeLabel.vue
+++ b/resources/assets/js/label-trees/components/labelTreeLabel.vue
@@ -128,7 +128,7 @@ export default {
             return this.editable && this.hover && !this.editing;
         },
         actualPosition(){
-            return this.position +1
+            return this.position + 1
         }
     },
     methods: {
--- a/resources/assets/js/label-trees/components/labelTreeLabel.vue
+++ b/resources/assets/js/label-trees/components/labelTreeLabel.vue
@@ -14,7 +14,7 @@
             <span class="label-tree-label__buttons">
                 <span class="label-tree-label_position">
                     <span v-if="showFavouriteShortcuts" class="text-muted">
-                            <span class="fa fa-keyboard" aria-hidden="true" title="" ></span>
+                            <span class="fa fa-keyboard" aria-hidden="true" title=""></span>
                             <span :text="actualPosition">{{ actualPosition }}</span>
                     </span>
                 </span>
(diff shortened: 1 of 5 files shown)


aneust:
possible to remove favorite by click on key, is not usefull

aneust:
Yes

resources/assets/js/label-trees/components/labelTreeLabel.vue
```diff
@@ -13,6 +13,10 @@
             </span>
             <span class="label-tree-label__buttons">
                 <button v-if="showFavourites" type="button" class="label-tree-label__favourite" :class="favouriteClass" @click.stop="toggleFavourite" :title="favouriteTitle">
+                    <span v-if="showFavouriteShortcuts" @click.stop="!toggleFavourite" >
```
mzur, line 16: `toggleFavourite` is a method that should be called when the element is clicked. This does not make sense for the keyboard icon and number here, so please remove it.

You may have added this because a click on the keyboard should not remove the favorite label. To fix this, just move the `<span>` outside the `<button>` element.

resources/assets/js/label-trees/components/labelTreeLabel.vue
```diff
@@ -13,6 +13,10 @@
...
+                        <span class="fa fa-keyboard" aria-hidden="true" title=""></span>
+                        <span v-text="addOnePosition()"></span>
```
mzur, line 18: `addOnePosition` is a computed property. These are not called like methods. Here you would write `:text="addOnePosition"`. In the code you would write `this.addOnePosition`. Since a computed property is used like a static property, you would also name it differently (not like a method), e.g. `positionText` or `actualPosition`.

resources/assets/js/label-trees/components/labelTreeLabel.vue
mzur, line 16: Can you please add a CSS class to the `<span>` and define `pointer-events: none;` for it (in [this file](https://github.com/biigle/core/blob/master/resources/assets/sass/label-trees/components/_labelTreeLabel.scss))? With this you can click on the icon and number and the label is still de-/selected.

mzur (changes requested):
Looks great, thanks! I left some minor comments.

resources/assets/js/label-trees/components/labelTreeLabel.vue
mzur: Also you could ass the `text-muted` class to make the icon and number a little darker. Then there is a better distinction between the icon/number and the star icon.

mzur:
Ready for review again? Please re-request my review whenever you are ready.

aneust:
Not yet, wanted to check if the click on icon is like wanted.

resources/assets/js/label-trees/components/labelTreeLabel.vue
```diff
@@ -12,11 +12,13 @@
...
             </span>
             <span class="label-tree-label__buttons">
-                <button v-if="showFavourites" type="button" class="label-tree-label__favourite" :class="favouriteClass" @click.stop="toggleFavourite" :title="favouriteTitle">
-                    <span v-if="showFavouriteShortcuts" @click.stop="!toggleFavourite" >
-                        <span class="fa fa-keyboard" aria-hidden="true" title=""></span>
-                        <span v-text="addOnePosition()"></span>
+                <span class="label-tree-label_position">
+                    <span v-if="showFavouriteShortcuts" class="text-muted">
+                            <span class="fa fa-keyboard" aria-hidden="true" title=""></span>
+                            <span :text="actualPosition">{{ actualPosition }}</span>
```
mzur, line 18: My fault, it should have been `v-text` instead of `:text`. But this is also fine as it does the same thing:
```suggestion
                            <span>{{ actualPosition }}</span>
```

I avoid the `{{}}` and use `v-text` instead in the Blade templates of Laravel. However, here in the Vue template it's fine.

resources/assets/js/label-trees/components/labelTreeLabel.vue
```diff
@@ -12,6 +12,12 @@
                 <span v-text="label.name" @click.stop="toggleSelect" @mouseenter="dontHover"></span>
             </span>
             <span class="label-tree-label__buttons">
+                <span class="label-tree-label_position">
+                    <span v-if="showFavouriteShortcuts" class="text-muted">
```
mzur, line 16: No need to nest two `<span>` here. You can use only one.

mzur (changes requested):
Almost perfect. Only two nitpicky comments left :wink:

resources/assets/js/label-trees/components/labelTreeLabel.vue
```diff
@@ -12,11 +12,9 @@
                 <span v-text="label.name" @click.stop="toggleSelect" @mouseenter="dontHover"></span>
             </span>
             <span class="label-tree-label__buttons">
-                <span class="label-tree-label_position">
-                    <span v-if="showFavouriteShortcuts" class="text-muted">
+                <span v-if="showFavouriteShortcuts" class="text-muted" :class="label-tree-label_position">
```
mzur, line 15: This throws an error now. It's an easy fix that I could do myself but I'd like you to understand what happens here. The `:class` is a Vue thing that can set a class based on a variable or computed property of the Vue instance. However, the variable or computed property `label-tree-label_position` does not exist. It is just the static string of the class name. This is how you add multiple CSS classes to an element:
```suggestion
                <span v-if="showFavouriteShortcuts" class="text-muted label-tree-label_position">
```

Also please test your own UI changes with the developer console open before you request a review in the future :wink:

mzur (changes requested):

resources/assets/js/label-trees/components/labelTreeLabel.vue
aneust: Can you explain me how to test that or send me where it it described?

mzur: You can test UI stuff of BIIGLE like this only by manually interacting with the UI in the browser. In this case, run `npm run watch` to make sure the compiled JS is up to date. Then open the image annotation tool and create a favourite label in the sidebar. Now press <kbd>F12</kbd> to open the browser developer tools and select the "Console" tab. You will see the JS errors there.

aneust: Thanks!

mzur (approved):
Thanks!

PR #846, opened 2024-06-04
Add button to ensure filtering is finished

### LINKED ISSUE #692: Fix clone volume UI for filtering files
If a volume should be cloned and the user wants to filter the files, it is necessary to press <kbd>Enter</kbd> in the filename filter input field so BIIGLE loads the list of files. If <kbd>Enter</kbd> is not pressed, no filtering will be done even if the input is filled.

The UI should be improved as follows:

1. When the filter input is filled, replace the "clone" button with a "test file query" button
2. The user has to press this button first and view the preview of filtered files
3. After the filtered files are displayed, the "clone" button appears again and can be pressed
4. If the filter query changes, start again at 1.

This will ensure that the clone button is pressed only after the list of filtered files was loaded.

lehecht:
Closes #692.

Add a 'Test file query' button to enable volume cloning only after the filtering process is finished.

### DIFF
--- a/resources/assets/js/volumes/cloneForm.vue
+++ b/resources/assets/js/volumes/cloneForm.vue
@@ -38,8 +38,7 @@ export default {
             fileLabelIds: [],
             annotationLabelIds: [],
             cloneUrlTemplate: "",
-            noFilesFoundByPattern: false,
-            showCloneButton: true,
+            noFilesFoundByPattern: false
         };
     },
     computed: {
@@ -108,7 +107,6 @@ export default {
                     if (this.selectedFiles.length === 0) {
                         this.noFilesFoundByPattern = true;
                     }
-                    this.showCloneButton = true;
                 });
 
         },
@@ -176,12 +174,6 @@ export default {
         cloneFileLabels(newState) {
             if (!newState) {
                 this.filterFileLabels = false;
-            }
-        },
-        filePattern() {
-            // Show clone button only after filter process is finished.
-            if (this.showCloneButton) {
-                this.showCloneButton = false;
             }
         },
     },
--- a/resources/assets/js/volumes/cloneForm.vue
+++ b/resources/assets/js/volumes/cloneForm.vue
@@ -38,7 +38,9 @@ export default {
             fileLabelIds: [],
             annotationLabelIds: [],
             cloneUrlTemplate: "",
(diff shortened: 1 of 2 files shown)


mzur (commented):
I think the original idea has to be refined a bit, as the UX is currently not good:

- Instead of replacing the "clone" button, the "test query" button should appear somewhere next to the query input field when "filter files" is checked. Perhaps you can find a solution that does not occupy too much additional space. The button should still be `.btn-success`.

- When "filter files" is checked and the query is unchecked/dirty, the "clone" button should be disabled with a `title` saying that the query has to be checked first.

- When "filter files" is unchecked, the "clone" button should be enabled again, even if the file query is unchecked/dirty.

- When the file query is checked/ok, the "test query" button can disappear.

resources/assets/js/volumes/cloneForm.vue
```diff
@@ -172,18 +173,30 @@ export default {
...
+            if (newPattern.length === 0) {
+                this.showTestQueryBtn = false;
+                return;
             }
+
+            this.showTestQueryBtn = this.filterFiles && (oldPattern !== newPattern);
         },
+        cannotSubmit() {
+            if (this.cannotSubmit) {
+                this.cloneBtnTitle = "The query has to be checked first before the volume can be cloned.";
```
mzur, line 195:
```suggestion
                this.cloneBtnTitle = "The query has to be tested first before the volume can be cloned.";
```

mzur (changes requested):
Can you please move the "test query" button above the file pattern input field? It can still be `.pull-right` and maybe change it to `.btn-xs`. As it is now, it overflows to the checkboxes below and can sometimes not be clicked. Also, it's quite far away from the input and the relation is unclear.

Now if I initially click on "filter files", the "clone" button is still active. However, cloning with an empty file list should not be possible. Also if I enter a file pattern that turns up no files, the clone button should still be disabled.

If I change the file query but don't click on "test query", the clone button is correctly disabled. But if I now uncheck "filter files" and immediately check it again, the file pattern is still changed and does not match with the previous tested file list. Furthermore, the clone button is enabled now but it shouldn't, as there now is an untested file pattern.

mzur (approved):
Thanks! I've tweaked the button position and updated the clone button title texts one final time.

PR #849, opened 2024-06-06
Add button to mark all notifications as read

### LINKED ISSUE #397: Button to mark all notifications as read
Implement a button in the notification area to mark all new notifications as read.

lehecht:
Closes #397.

### DIFF
--- a/app/Http/Controllers/Api/NotificationController.php
+++ b/app/Http/Controllers/Api/NotificationController.php
@@ -3,6 +3,8 @@
 namespace Biigle\Http\Controllers\Api;
 
 use Illuminate\Http\Request;
+use Biigle\Http\Requests\UpdateAllNotifications;
+use Illuminate\Notifications\DatabaseNotification;
 
 class NotificationController extends Controller
 {
@@ -26,6 +28,29 @@ class NotificationController extends Controller
     {
         $notification = $request->user()->unreadNotifications()->findOrFail($id);
         $notification->markAsRead();
+    }
+
+    /**
+     * Mark all notification as read.
+     *
+     * @api {put} notifications/ Mark all notifications as read
+     * @apiGroup Notifications
+     * @apiName UpdateReadNotifications
+     * @apiPermission user
+     *
+     * @apiParamExample {String} Request example:
+     * id: "0972569c-2d3e-444d-8e7d-2054e7ab20e9"
+     *
+     * @param Request $request
+     * @return \Illuminate\Http\Response
+     */
+    public function updateAll(UpdateAllNotifications $request)
+    {
+        $userId = $request->input('user_id');
+        $notifications = DatabaseNotification::where('notifiable_id', '=', $userId)->whereNull('read_at')->get();
+        $notifications->map(function ($n) {
+            $n->markAsRead();
+        });
     }
 
     /**
--- a/app/Http/Requests/UpdateAllNotifications.php
+++ b/app/Http/Requests/UpdateAllNotifications.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace Biigle\Http\Requests;
+
+use Illuminate\Foundation\Http\FormRequest;
+
+class UpdateAllNotifications extends FormRequest
+{
+    /**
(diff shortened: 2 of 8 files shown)


app/Http/Controllers/Api/NotificationController.php
```diff
@@ -28,6 +30,29 @@ public function update(Request $request, $id)
...
+     * @param Request $request
+     * @return \Illuminate\Http\Response
+     */
+    public function updateAll(UpdateAllNotifications $request)
+    {
+        $userId = $request->input('user_id');
+        $notifications = DatabaseNotification::where('notifiable_id', '=', $userId)->whereNull('read_at')->get();
+        $notifications->map(function ($n) {
+            $n->markAsRead();
+        });
```
mzur, line 53: Why not use similar code than in the `update()` method?
```php
$request->user()->unreadNotifications()->eachById(fn ($n) => $n->markAsRead());
```

app/Http/Requests/UpdateAllNotifications.php
mzur, line 1: A user is always allowed to update their own notifications. When `$request->user()` is used (as suggested), the form request class and authorization check is not necessary.

resources/assets/js/core/notifications/list.vue
```diff
@@ -73,6 +73,24 @@ export default {
...
+                        item.read_at = new Date();
+                        if (this.removeItem) {
+                            Store.remove(this.item.id);
+                        }
+                    })
+                })
+                .catch(Messages.handleErrorResponse)
+                .finally(() => {
+                    this.isLoading = false;
+                });
```
mzur, line 91: There is no `this.isLoading` property in the notification list component, yet. Please add it and set the "mark all as read" button as disabled when loading.

resources/views/notifications/index.blade.php
```diff
@@ -19,6 +19,12 @@
             </ul>
         </div>
         <div id="notifications-list" class="col-sm-9 col-md-7 col-md-offset-1" v-cloak>
+            @if (!$all)
+                <p v-cloak v-if="hasUnreadNotifications" class="row">
+                    <button class="btn btn-default btn-s pull-right" v-on:click="markAllAsRead({{$user_id}})">Mark all as read</button>
+                </p>
+            @endif
+            <div class="row">
```
mzur, line 27: Why add `class="row"`? The elements overflow the `notifications-list` now.

routes/api.php
```diff
@@ -147,6 +147,8 @@
     'only' => ['update', 'destroy'],
 ]);
 
+$router->put('notifications', 'NotificationController@updateAll');
```
mzur, line 150: I think I would prefer `notifications/all` as URL. This way there is less chance to update all notifications by accident if the ID is forgotten (e.g. in a custom script). While it is very unlikely that people use custom scripts to update notifications, it's a more robust default.

mzur (changes requested):

resources/views/notifications/index.blade.php
lehecht: The button would not align with the grid borders if I didn't use it.

mzur: For me it does exactly that with `class="row"`:
image

If I remove the class and the new `<div>` it works. I have to add `class="clearfix"` to the new `<p>` though.

app/Http/Controllers/Api/NotificationController.php
```diff
@@ -28,6 +28,24 @@ public function update(Request $request, $id)
...
+     * Mark all notification as read.
+     *
+     * @api {put} notifications/all Mark all notifications as read
+     * @apiGroup Notifications
+     * @apiName UpdateReadNotifications
+     * @apiPermission user
+     *
+     * @apiParamExample {String} Request example:
+     * id: "0972569c-2d3e-444d-8e7d-2054e7ab20e9"
+     *
```
mzur, line 41:
```suggestion
```

mzur (approved):
All good, thanks!

PR #850, opened 2024-06-10
Fix file browser view

lehecht:
Prevent displaying file names twice on review page.

References biigle/core#835

### DIFF
--- a/resources/assets/js/core/components/fileBrowserFile.vue
+++ b/resources/assets/js/core/components/fileBrowserFile.vue
@@ -4,19 +4,21 @@
     :class="classObject"
     @click="handleClick($event)"
     >
-    <div v-if="file.url">
-        <a :href="file.url" :title="viewTitle">
-        <i class="fa fa-file"></i> {{file.name}}
-        </a>
-    </div>
-    <div v-else>
+    <div v-if="hasError">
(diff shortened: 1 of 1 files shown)


mzur (commented):
Thanks for spotting this! I'm not sure about the nesting, though. For example, if the file has a URL, no error can be shown. What about something like this?
```
if error
  ...
else
  if url
     ...
   else
     ...
```

resources/assets/js/core/components/fileBrowserFile.vue
```diff
@@ -4,22 +4,23 @@
...
-    <span v-if="hasError" :title="file.name" class="text-warning">
-        <i class="fa fa-exclamation-triangle"></i> {{file.name}}
-    </span>
-    <span v-else :title="file.name">
-        <i v-if="hasInfo" class="fa fa-info-circle"></i>
-        <i v-else class="fa fa-file"></i>
-        {{file.name}}
-    </span>
+    <div v-if="hasError">
+        <span v-if="hasError" :title="file.name" class="text-warning">
```
mzur, line 8:
```suggestion
        <span :title="file.name" class="text-warning">
```

mzur (approved):

PR #852, opened 2024-06-13
Fix clone job query

### LINKED ISSUE #848: Chunk clone volume filter query
There was this error:
```
number of parameters must be between 0 and 65535 (Connection: pgsql, SQL: select * from "image_annotations" where "image_annotations"."image_id" in (834986, ...
```
In the CloneImagesOrVideos job. I think this is because someone actually selected more then 65k images [here](https://github.com/biigle/core/blob/96b6f31459b3b8cffb06c355ee300dedfd80cb7f/app/Jobs/CloneImagesOrVideos.php#L263). Update the job so the query also chunks the filtered images.

lehecht:
Fix clone job query by chunking annotations and annotation labels.

Fixes #848.

### DIFF
--- a/app/Jobs/CloneImagesOrVideos.php
+++ b/app/Jobs/CloneImagesOrVideos.php
@@ -257,8 +257,16 @@ class CloneImagesOrVideos extends Job implements ShouldQueue
         $newImageIds = $copy->images()->orderBy('id')->pluck('id');
         $volume->images()
             ->with([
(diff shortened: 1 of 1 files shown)


mzur (commented):
Thanks for identifying the failing query! Your change does not solve the problem, unfortunately. I have pushed an updated test case that actually checks for label filtering (before it didn't). Your updated code no longer does the filtering.

As this is a little urgent, I have proposed a different solution in https://github.com/biigle/core/pull/853. Could you please review that and confirm that it should work as expected?

Issue #819, opened 2024-04-30
Option to enable video audio


merlinbest5:
Apologies if this has been discussed, but we would greatly benefit from having the audio playable in biigle videos. We often have scientists narrating various parts of the dive and the information can be very helpful for annotators (or at the very least entertaining).

mzur:
Sounds like an easy thing to do. We could implement a switch in the annotation tool options to enable the audio.

merlinbest5:
@mzur @dlangenk amazing! my team is really excited to have this function, many thanks for your efforts.

PR #856, opened 2024-06-14
Implement setting to mute/unmute video

dlangenk:
implements #819


mzur (approved):
Thanks!. I've added a missing parameter for the settings store object and extended the manual article.

PR #880, opened 2024-07-15
Exif fields

### LINKED ISSUE #370: Support EXIF fields for image area and yaw
It was suggested that the EXIF field `SubjectArea` could be imported as image area and `GPSImgDirection` as yaw.

aneust:
Support EXIF fields for image area and yaw 

resolves #370


app/Jobs/ProcessNewImage.php
```diff
@@ -206,6 +206,18 @@ protected function collectMetadata(Image $image, $path)
                     'gps_altitude' => $ref * $this->fracToFloat($exif['GPSAltitude']),
                 ]);
             }
+
+            if ($this->hasGpsImgDirInfo($exif)) {
+                $image->metadata = array_merge($image->metadata, [
+                    'yaw' => $ref * $this->fracToFloat($exif['GPSImgDirection']),
```
mzur, line 212: The `$ref` should not be used here as it is only valid for the `GPSAltitude`. Also this would fail if `GPSAltitude` is not set because then `$ref` does not exist.

app/Jobs/ProcessNewImage.php
```diff
@@ -206,6 +206,18 @@ protected function collectMetadata(Image $image, $path)
...
+                ]);
+            }
+
+            if ($this->hasSubjectAreaInfo($exif)) {
+                $image->metadata = array_merge($image->metadata, [
+                    'area' => $ref * $this->fracToFloat($exif['SubjectArea']),
```
mzur, line 218: Same as above about `$ref`.

tests/php/Jobs/ProcessNewImageTest.php
```diff
@@ -61,6 +61,25 @@ public function testHandleCollectMetadataZeroDate()
...
+        ]);
+
+        with(new ProcessNewImageMock($image))->handle();
+
+        $image = $image->fresh();
+
+        $this->assertEquals(47.75, $image->metadata['yaw']);
+        $this->assertEquals(2.6, $image->metadata['area']);
+    }
+
```
mzur, line 82: This does not test the `SubjectArea` import. You can either modify the `exif-test.jpg` file or use the `$exif` attribute of the `ProcessNewImageMock` class.

mzur (changes requested):

mzur (approved):
Thanks!

PR #888, opened 2024-07-23
Use consistent colors for users in project/volume charts

### LINKED ISSUE #497: Use consistent colors for users in project/volume charts
The colors for the users can be different in the timeline and the sankey plot. They should be consistent, if possible.

dlangenk:
This way the colors are consistent fixes #497


mzur (commented):
I'm not sure if adding the whole md5 algorithm is worth it here. Also this could randomly produce very dark colors, right? What about reusing the [random color function](https://github.com/biigle/core/blob/54c12ba8684bbc3637a74a67e9a070523728b36f/resources/assets/js/label-trees/utils.js#L56-L70) of new labels and then generating a map wit colors for each user on the fly? The only downside would be that colors of users are different between page loads...

So if we want consistent colors between page loads, md5 may be a valid option. But then I'd like to at least change it so the user ID is used for hashing instead of the name and the random label color function is used to get a bright color. The function could even be adapted to accept an optional seed.

resources/assets/js/projects/components/charts/sankeyPlot.vue
```diff
@@ -34,7 +35,7 @@ export default {
         data() {
             // returns an array of User-names and volume-names
             let volNames = this.volumeAnnotations.map(entry => {
```
mzur, line 37: The `user_id`s are available here in `this.volumeAnnotations`.

resources/assets/js/projects/components/charts/usernameToColor.js
```diff
@@ -0,0 +1,37 @@
...
+    const hash = simpleHash(name);
+
+    // Use the hash to generate a unique hue
+    const hue = hash % 360;
+    const saturation = 70; // Keep saturation high for bright colors
+    const lightness = 60;  // Keep lightness moderate for bright colors
+
+    const [r, g, b] = hslToRgb(hue, saturation, lightness);
+
+    return `#${toHex(r)}${toHex(g)}${toHex(b)}`;
```
mzur, line 36: You can make it even simpler with a CSS HSL color:

```suggestion
    const hue = hash % 360;
    
    return `hsl(${hue} 70% 60%)`;
```

mzur (commented):
The user IDs are available but it might be a little complicated to get them where you need them. While I'd still prefer the IDs over the user names, I'm ok with the names for now, as they will work in most cases. We can open a follow-up issue to use the IDs at some point.

resources/assets/js/projects/components/charts/IDToColor.js
```diff
@@ -0,0 +1,21 @@
...
+    hash += (hash << 3);
+    hash ^= (hash >> 11);
+    hash += (hash << 15);
+    id=hash & 0xFFFFFFFF; // Ensure it fits in 32 bits
+
+
+    // Use the hash to generate a unique hue
+    const hue = id % 360;
+    const saturation = 70 + (hash % 31); // Ensure saturation is between 70 and 100
+    const lightness = 50 + (hash % 41);  // Ensure lightness is between 50 and 90
```
mzur: I didn't see a difference between `id` and `hash`. Did you really mean to use both here?

dlangenk: Oops. Nope that was an error. Should be id all the way.

mzur (dismissed):

mzur:
One thing I now noticed is that if the users did not annotate in consecutive months, the x-axis is not "to scale", if you know what I mean. Do we think this as a bug or a feature? :smile:

dlangenk:
For now it's a feature but I guess it's rather a bug ;-)

mzur:
Ok then I'll create a follow-up issue.

PR #889, opened 2024-07-23
truncates long image names in the navbar/content

### LINKED ISSUE #240: Truncate long image names
Truncate long image names (e.g. image 897914) with `text-overflow: ellipsis` and the full name as `title`.

dlangenk:
Fixes #240. As I couldn't figure out how you trimmed the image filename in the annotation navbar, I used the php mb_strimwidth function. In the content the width fits exactly if you would use a resolution of 1024 x 768.

### DIFF
--- a/resources/assets/sass/volumes/main.scss
+++ b/resources/assets/sass/volumes/main.scss
@@ -14,8 +14,9 @@ $volume-image-padding: 0.5em;
 }
 
 .navbar-text.navbar-volumes-breadcrumbs {
-    &, a {
-        color: $text-color;
+
+    .annotations-breadcrumb {
+        max-width: 600px;
     }
 }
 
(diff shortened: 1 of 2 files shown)


mzur (changes requested):
Sorry but this should be done with CSS (`text-overflow: ellipsis`). This will adjust to the available space.

dlangenk:
This was far more difficult than I expected. But it works now. Not sure about if the navbar is ok like this. I tried to use the breadcrump component but failed. Probably we can have a look tomorrow how to best debug such a thing.

mzur:
If you want something easy, pick one of the "difficulty 1" [tasks](https://github.com/orgs/biigle/projects/4/views/9) :smiley:

I'll wrap this up after lunch.

mzur (approved):

PR #891, opened 2024-07-24
Prevent double click on submit of a new project

### LINKED ISSUE #335: Prevent double click on submit of a new project
Disable the button on submit of a new project that should be created.

dlangenk:
fixes #335. Uses js events. Is there a better way?

### DIFF
--- a/resources/views/projects/create.blade.php
+++ b/resources/views/projects/create.blade.php
@@ -24,20 +24,8 @@
 			</div>
 			<input type="hidden" name="_token" value="{{ csrf_token() }}">
 			<a href="{{ URL::previous() }}" class="btn btn-link">Cancel</a>
-			<input id="submit-button" type="submit" class="btn btn-success pull-right" value="Create">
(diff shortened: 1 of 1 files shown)


mzur (commented):
You can do it like this, which is essentially the same but shorter:

https://github.com/biigle/core/blob/b224653d688197a936e7f904d14e39c49edbe437/resources/views/auth/register.blade.php#L115

resources/views/projects/create.blade.php
```diff
@@ -24,8 +24,8 @@
 			</div>
 			<input type="hidden" name="_token" value="{{ csrf_token() }}">
 			<a href="{{ URL::previous() }}" class="btn btn-link">Cancel</a>
-			<input type="submit" class="btn btn-success pull-right" value="Create">
+			<input id="submit-button" type="submit" class="btn btn-success pull-right" value="Create" onclick="this.disabled = true;this.value = 'Creating project...';this.form.submit();">
```
mzur, line 27:
```suggestion
			<input type="submit" class="btn btn-success pull-right" value="Create" onclick="this.disabled = true;this.value = 'Creating project...';this.form.submit();">
```

mzur (approved):

PR #896, opened 2024-07-26
Prevent double clicks for point annotations

### LINKED ISSUE #491: Catch double-click for point annotations in UI
If people double-click by accident when they create a point annotation, the annotation is duplicated. Catch this and don't allow new annotations within a certain cooldown period (in the UI not the API). This is only relevant for point image annotations (what about whole frame video annotations?).

dlangenk:
fixes #491

### DIFF
--- a/resources/assets/js/annotations/components/annotationCanvas/drawInteractions.vue
+++ b/resources/assets/js/annotations/components/annotationCanvas/drawInteractions.vue
@@ -6,6 +6,13 @@ import { shiftKeyOnly } from '@biigle/ol/events/condition';
 import snapInteraction from '../../snapInteraction.vue';
 import { Point } from '@biigle/ol/geom';
 
+
+function computeDistance(point1, point2) {
+    let p1=point1.getCoordinates();
+    let p2=point2.getCoordinates();
+    return Math.sqrt(Math.pow(p2[0] - p1[0], 2) + Math.pow(p2[1] - p1[1], 2));
+}
+
 /**
  * Mixin for the annotationCanvas component that contains logic for the draw interactions.
  *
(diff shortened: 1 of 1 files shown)


mzur (changes requested):
I can produce incorrect features with this change (the blue ones):

[Screencast from 26.07.2024 09:58:05.webm](https://github.com/user-attachments/assets/b0bf8126-28ad-49a7-8ba2-820faa5a0137)

If I try to delete one of those, the app crashes. The features that are created within the cool down period should be immediately removed again. Also maybe we could check if the new point is "near" the previous point (there should be an OpenLayers method for that)? If not, users should be able to create points faster than the cool down period (as in the video).

dlangenk:
If i throttle my connection speed I also see that problem. Before I simply could not reproduce that.

mzur:
Strange, it should not have anything to do with the connection speed. This is purely client-side. You have to click faster than the 400 ms for this to happen.

dlangenk:
I guess now the point is removed after all, BUT the map is also dragged after a double click. I also added the distance to last annotation feature.

mzur (changes requested):
I've updated the code a bit so the point and time are only updated if drawing points. However, this still fails because there is no `distanceTo()` method in an OpenLayers geometry. LLM fail? :smile: 

This is likely also the reason why the map is "dragged" for you because there should be a JS error in your console.

dlangenk:
Pretty sure that I actually got a distance and no error, but I'll have a look at this. No LLM fail😉, but rather [StackOverflow](https://stackoverflow.com/questions/10109620/openlayers-how-to-calculate-distance-between-two-points)

dlangenk:
Don't know why the console error slipped. Now it works, without any errors ;-) Please have a look if the position of the computeDistance function is correct or if it should be somewhere else.

mzur (approved):
The distance calculation works but the feature wasn't removed for me now, too. This was because the feature is added to the source *after* the drawend event was handled. I fixed that now.

PR #902, opened 2024-07-31
Fix slider focus to enable keyboard shortcuts again

### LINKED ISSUE #635: Keyboard shortcuts not working after changing settings
After changing e.g. the opacity slider, it retains the focus and keyboard shortcuts (e.g. activate magic lasso, magic sam, change label) are not working. It would be nicer if the the shortcuts would still work. Clicking on the image should at least reenable the shortcuts, i.e. get the focus away from the slider. The only way now is to either click on an empty area in the settings or changing the sidebar tab / closing it.

lehecht:
Remove slider focus after usage so that keyboard events are working again.

Fixes #635.

### DIFF
--- a/resources/views/annotations/show/tabs/settings.blade.php
+++ b/resources/views/annotations/show/tabs/settings.blade.php
@@ -10,12 +10,12 @@
 
             <div class="sidebar-tab__section">
                 <h5 title="Set the opacity of annotations on the map">Annotation Opacity (<span v-text="annotationOpacity"></span>)</h5>
-                <input id="annotationOpacity" type="range" min="0" max="1" step="0.1" v-model="annotationOpacity" v-on:click='removeFocus("annotationOpacity")'>
+                <input type="range" min="0" max="1" step="0.1" v-model="annotationOpacity">
             </div>
 
             <div class="sidebar-tab__section">
                 <h5 title="Set the number of caches images ">Cached Images (<span v-text="cachedImagesCount"></span>)</h5>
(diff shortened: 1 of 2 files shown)


lehecht:
I only applied some changes on the frontend, so I don't think the tests failed because of them.

mzur:
The failing tests were bad luck and caused by https://github.com/biigle/core/pull/907 and https://github.com/biigle/kpis/pull/15.

resources/views/annotations/show/tabs/settings.blade.php
```diff
@@ -10,12 +10,12 @@
 
             <div class="sidebar-tab__section">
                 <h5 title="Set the opacity of annotations on the map">Annotation Opacity (<span v-text="annotationOpacity"></span>)</h5>
-                <input type="range" min="0" max="1" step="0.1" v-model="annotationOpacity">
+                <input id="annotationOpacity" type="range" min="0" max="1" step="0.1" v-model="annotationOpacity" v-on:click='removeFocus("annotationOpacity")'>
```
mzur, line 13: Good approach. It could be less code with vanilla JS:
```suggestion
                <input type="range" min="0" max="1" step="0.1" v-model="annotationOpacity" onmouseup="this.blur()">
```
What do you think?

lehecht: I really like your idea. Thanks!

mzur (approved):

PR #903, opened 2024-08-01
Add file sort for last created annotations

### LINKED ISSUE #477: Sort by "last annotated" in the volume overview
Add a sorting method by "last annotated" to the volume overview. If the sorting method is active, the image that was annotated last (by `created_at` of the annotations) is displayed first.

lehecht:
Add a new file sort to the volume overview, where files are sorted by their latest annotation.

Closes #477.


mzur:
Ready for review?

mzur (changes requested):
Please implement an API endpoint that returns the timestamps for the images instead of passing them on to the view every time. This could be a lot of data if the volume is large and it could be unnecessary data if the sorter is not used. The `getSequence()` method of the sorting component already returns a promise so taking the result of an API request here should work out of the box.

Also you should implement getting the timestamps for each file in a single database query (hint: combine `groupBy` with `max`). Currently you fetch all annotations of the entire volume and this could be literally hundreds of thousands.

app/Http/Controllers/Api/VolumeController.php
```diff
@@ -209,4 +209,36 @@ public function clone(CloneVolume $request)
...
+     */
+    public function getAnnotationTimestamps($id)
+    {
+        $volume = Volume::findOrFail($id);
+        $this->authorize('access', $volume);
+
+        $timestamps = $volume->files
+            ->flatMap(fn ($file) => $file->annotations)
+            ->groupBy('file_id')
+            ->map(fn ($annotations) => $annotations->max('created_at'));
```
mzur, line 240: It may be even better to just return the sorted file IDs from this endpoint. Then you don't need to do the sorting in the JS and transfer less data.

Also I challenge you to find a single database query that sorts the files by the newest annotation timestamp and also includes files without annotations (at the end). Hints: `groupBy` (again) and `leftJoin` :wink: This will make the API endpoint much more efficient, as it does not have to process possibly hundreds of thousands of annotations.

lehecht: Can you explain, why this approach would not process a large amount of annotations?

mzur: There are three issues:

1. `$volume->files`: This loads all files of the volume into a collection. Volumes can have many thousands of files so this may already consume a lot of memory. Often building a query with `$volume->files()` is better but this wouldn't work here, too.

2. `->flatMap(fn ($file) => $file->annotations)`: This replaces the files collection with a collection of annotations. *All* annotations in the volume. This can obviously be huge as well.

3. `->map(fn ($annotations) => $annotations->max('created_at'))`: This executes a database query for each file. With thousands of files, you have thousands of queries.

Let the database do all the work, it will always be faster and more memory efficient! What you do with 2*n*+1 queries (with *n* the number of files) can also be done with 1 query.

app/Http/Controllers/Api/VolumeController.php
```diff
@@ -211,34 +211,43 @@ public function clone(CloneVolume $request)
...
+            $ids = $volume->files()
+                ->leftJoin('video_annotations', 'videos.id', "=", "video_id")
+                ->groupBy('videos.id')
+                ->selectRaw('videos.id, max(created_at) as created_at');
+        }
 
-        return $timestamps;
+        return $ids->get()
+            ->sortByDesc('created_at')
+            ->pluck('id');
```
mzur, line 251: You can also do the sorting in the DB query. Let's find the perfect query! :smile:

app/Http/Controllers/Api/VolumeController.php
```diff
@@ -211,34 +211,43 @@ public function clone(CloneVolume $request)
...
+            $ids = $volume->files()
+                ->leftJoin('image_annotations', 'images.id', "=", "image_id")
+                ->groupBy('images.id')
+                ->selectRaw('images.id, max(created_at) as created_at');
+        } else {
+            $ids = $volume->files()
+                ->leftJoin('video_annotations', 'videos.id', "=", "video_id")
+                ->groupBy('videos.id')
+                ->selectRaw('videos.id, max(created_at) as created_at');
+        }
```
mzur, line 247: I always specify the full table and column names. While it works here, it could fail in the future if e.g. a created_at` column was added to `images` or `annotations`.

```suggestion
            $ids = $volume->files()
                ->leftJoin('image_annotations', 'images.id', "=", "image_annotations.image_id")
                ->groupBy('images.id')
                ->selectRaw('images.id, max(image_annotations.created_at) as created_at');
        } else {
            $ids = $volume->files()
                ->leftJoin('video_annotations', 'videos.id', "=", "video_annotations.video_id")
                ->groupBy('videos.id')
                ->selectRaw('videos.id, max(video_annotations.created_at) as created_at');
        }
```

tests/php/Http/Controllers/Api/VolumeControllerTest.php
```diff
@@ -351,9 +351,7 @@ public function testGetAnnotationTimestamps()
         $content = json_decode($response->getContent(), true);
 
         $this->assertCount(2, $content);
-        $this->assertNotNull($content[$img->id]);
-        $this->assertNotNull($content[$img2->id]);
-        $this->assertSame($content[$img->id], "2023-11-09T05:37:00.000000Z");
-        $this->assertSame($content[$img2->id], "2024-11-09T13:37:00.000000Z");
+        $this->assertSame($content[0], $img2->id);
+        $this->assertSame($content[1], $img->id);
```
mzur, line 355: You should also add an image without annotations to see if it is properly sorted at the end.

tests/php/Http/Controllers/Api/VolumeControllerTest.php
mzur, line 1: Please also add a test for a video volume.

mzur (changes requested):

app/Http/Controllers/Api/VolumeController.php
```diff
@@ -209,4 +209,45 @@ public function clone(CloneVolume $request)
             ->with('messageType', 'success');
 
     }
+
+    /**
+     * Return file ids which are sorted by annotations.created_at
+     *
+     * @param int $id
+     * @return object
+     * @api {get} volumes{/id}/files/annotation-timestamps Get sorted file ids of volume
```
mzur, line 218:
```suggestion
     * @api {get} volumes{/id}/files/annotation-timestamps Get file ids sorted by recently annotated
```

app/Http/Controllers/Api/VolumeController.php
```diff
@@ -209,4 +209,45 @@ public function clone(CloneVolume $request)
...
+     * @apiGroup Volumes
```
mzur, line 219:
```suggestion
     * @apiGroup Volumes
     * @apiName VolumeSortByAnnotated
```

mzur (changes requested):

mzur (approved):

PR #905, opened 2024-08-17
View recently joined users in admin area 

### LINKED ISSUE #691: Add "recently joined" view to users list in admin area
Add a "recently joined" button to the top of the users list in the admin area. When the button is pressed, the view is updated to show only the users who joined within the last 7 days. As this is similar to search/filtering, it should be put next to the search input. The "new user" button can be moved to the right.

aneust:
Introduces a new feature to view a list of users who have joined within the last 7 days. 

Added a new route (users/recent) and corresponding controller method (getRecentUsers) to fetch and display users who recently joined.
In view added a "Recently Joined" button in the admin area, which highlights the number of users who joined recently and navigates to the new view.
Created a new Blade template to display the recent users, including account creation date.

Resolves #691


mzur (commented):
Thanks!

Please put the search input field first, then the "recently joined button" and then the "new user" button with a `pull-right` class.

Also, instead of adding a new view, you can filter the list of users in the same view. To do this you can add a query parameter (e.g. `recent=1`) to the URL. The button URL would then be `href="{{route('admin-users-recent')}}?recent=1"`. In the controller you can get the parameter with `$request->get('recent')`. If the parameter is true, you can conditionally add a filter to the user query like this:

```php
->when('$request->get('recent'), fn ($query) => $query->where('created_at', '>=', now()->subWeek())
```
If the user list is filtered, you can add a `btn-info` class to the "recently joined" button, to make the state clear.

PR #910, opened 2024-08-22
 Fill empty months in timeline chart 

### LINKED ISSUE #897: Fill empty months in timeline chart
https://github.com/biigle/core/pull/893 changed the x-axis of the timeline chart from years to months. This makes it more likely that there are gaps between months where no annotations were created. These are currently ignored, so there may be non-consecutive months displayed on the x-axis. If a month is empty, fill it in the plot so the x-axis is consecutive again.

dlangenk:
#fixes #897


dlangenk:
Please don't review yet. I found a bug

dlangenk:
> Please don't review yet. I found a bug

Bug is fixed

mzur (approved):
Thanks

PR #927, opened 2024-09-20
Use scrollable typeahead to attach volumes/labeltrees

### LINKED ISSUE #876: Improve attach volume input
Currently the attach volume input of the project overview loads a complete list of available label trees on the `focus` event and then a typeahead is used to search and select the label tree. This has two problems:

1. The input loses focus again after the label trees were loaded. The user has to focus the element again afterwards.

2. There may be many label trees with similar names. If the user searches for generic "fauna", for example, the label tree may not show up in the top 5 of the typeahead results.

To fix these issues, implement a fuzzy search API endpoint for label trees. The endpoint is queried after the user has entered at least 3 characters and stopped typing for 500 ms (?). As long as the user is not typing, the full list of returned label trees is shown in a dropdown.

lehecht:
Add new scrollable typeahead component for project view.

Closes #876.

TODO:

- [x] Fix CSS


lehecht:
> You return some unnecessary information for volumes and label trees from the API endpoints. 

I reused the old controller methods that were used before to fetch all the volumes or trees. Can you explain why you needed the additional data before? I thought that the data was used somewhere in the container components.

mzur:
I see, you are right! The additional information is used to immediately show the volume thumbnails etc. when the volume is attached.

app/Http/Controllers/Api/ProjectLabelTreeController.php
```diff
@@ -86,19 +86,22 @@ public function index($id)
...
      */
-    public function available($id)
+    public function available($id, $name)
     {
         $project = Project::findOrFail($id);
         $this->authorize('access', $project);
 
         $public = LabelTree::publicTrees()
             ->select('id', 'name', 'description', 'version_id')
+            ->whereRaw('UPPER(name) LIKE UPPER(?)', ["%{$name}%"])
```
mzur, line 99: You could also use:
```php
->where('name', 'ilike', "%{$name}%")
```
This is basically the same but there is less chance to accidentally allow a SQL injection with a raw query. While `ilike` is not supported in all DBMS, BIIGLE exclusively uses Postgres so it is ok.

app/Http/Controllers/Api/ProjectLabelTreeController.php
```diff
@@ -59,13 +59,13 @@ public function index($id)
     }
 
     /**
-     * Display all label trees that can be used by the specified project.
+     * Display label trees that match the given name and can be used by the specified project.
      *
      * @api {get} projects/:id/label-trees/available Get all available label trees
```
mzur, line 64: You should add the name to the route docs.

app/Http/Controllers/Api/ProjectsAttachableVolumesController.php
```diff
@@ -10,14 +10,14 @@
 class ProjectsAttachableVolumesController extends Controller
 {
     /**
-     * Shows all volumes that can be attached to the project by the requesting user.
+     * Shows volumes that match the given volume name and can be attached to the project by the requesting user.
      *
      * @api {get} projects/:id/attachable-volumes Get attachable volumes
```
mzur, line 15: Same as above, please update the route.

resources/assets/js/core/components/scrollableTypeaheadItem.vue
mzur, line 1: Instead of adding a new component, you could also update the default `typeaheadItem` and add a new prop (which is `false` by default so the new CSS class is added only in the scrollable typeahead). This is less code and less files.

In fact, you could also check if the default typeahead component could be extended with a `scrollable=true` (default `false`) prop. If this is easy, it would be even less new code and files.

tests/php/Http/Controllers/Api/ProjectLabelTreeControllerTest.php
mzur, line 1: Please add a test case with the fuzzy search (e.g. tree name "my tree" and request param "tree").

tests/php/Http/Controllers/Api/ProjectsAttachableVolumesControllerTest.php
mzur, line 1: Same as above please add a test case for the new fuzzy search.

mzur (changes requested):
Remaining high-level comments:
- Please already send the request on `length >= 3` characters.
- Can you implement the typeahead so that it automatically scrolls when I move down the focused item using the arrow keys? Now the focused item becomes invisible in this case.

The rest is mostly minor comments.

app/Http/Controllers/Api/ProjectLabelTreeController.php
lehecht: Ah, okay. Could you explain why my approach is still vulnerable to SQL injections? When I was writing the code, this method was recommended as a secure way to prevent SQL injections.

mzur: It is safe as it is (because you use `?` instead of adding `$name` directly to the string). However, I would only use `whereRaw()` if there is no other way. It's just a matter of risk avoidance.

lehecht: Thanks!

resources/assets/js/core/components/scrollableTypeaheadItem.vue
lehecht: > In fact, you could also check if the default typeahead component could be extended with a `scrollable=true` (default `false`) prop. If this is easy, it would be even less new code and files.

Extending the functionality is easy, but not the css part. I don't know if it's just me struggling with the css, but now I need to check every typeahead. Fixing one typeahead always breaks another one.
Do you want to look through my code or should I just use the new component?

mzur: I can have a look at the CSS if you want. If it makes it easier, you could also move the CSS into the single file component and use `<style scoped>` so the CSS is only valid for the component and nothing else. At some point I want to include all CSS into the single file components anyway.

lehecht: > I can have a look at the CSS if you want.

That would be really helpful. I have already spent several hours just trying to fix the css.

lehecht:
@mzur 

Here are two examples of weird looking typeaheads.

Merge labetrees typeahead
Bildschirmfoto vom 2024-10-22 13-05-15

Clone file labels typeahead (clone volume)
Bildschirmfoto vom 2024-10-22 12-53-40

app/Http/Controllers/Api/ProjectLabelTreeController.php
```diff
@@ -61,7 +61,7 @@ public function index($id)
     /**
      * Display label trees that match the given name and can be used by the specified project.
      *
-     * @api {get} projects/:id/label-trees/available Get all available label trees
+     * @api {get} projects/:id/label-trees/available Get matching label trees
```
mzur, line 64: The name is still missing:

```suggestion
     * @api {get} projects/:id/label-trees/available/:name Get matching label trees
```

app/Http/Controllers/Api/ProjectsAttachableVolumesController.php
```diff
@@ -12,7 +12,7 @@ class ProjectsAttachableVolumesController extends Controller
     /**
      * Shows volumes that match the given volume name and can be attached to the project by the requesting user.
      *
-     * @api {get} projects/:id/attachable-volumes Get attachable volumes
+     * @api {get} projects/:id/attachable-volumes Get matching attachable volumes
```
mzur, line 15:
```suggestion
     * @api {get} projects/:id/attachable-volumes/:name Get matching attachable volumes
```

resources/assets/js/core/components/typeahead.vue
```diff
@@ -81,17 +92,39 @@ export default {
...
             default: () => TypeaheadItem,
         },
+        scrollable: {
+            type: Boolean,
+            default: false,
+        },
+        isLabelTree: {
+            type: Boolean,
+            default: false,
+        }
```
mzur, line 102: Why do wen need `isLabelTree` here? The only difference I can see is that the typeahead item has height 44px then instead of 40px but why does it need another height?

resources/assets/js/core/components/typeahead.vue
```diff
@@ -108,11 +141,52 @@ export default {
...
+
+            this.timerTask = setTimeout(() => {
+                let added = v.trim().includes(this.oldInput.trim());
+                let useTypeaheadFilter = this.oldInput.length > 3 && added;
+                if (v.length >= 3 && !useTypeaheadFilter) {
+                    this.$emit('fetch', v);
+                }
+                this.isTyping = false;
+                this.oldInput = v
+            }, 500);
```
mzur, line 182: You can also use the [debounce](https://github.com/biigle/core/blob/1365b19497f1b57e3a2ba24374a78dfc7f606638/resources/assets/js/core/utils.js#L17-L22) helper method for this:

```js
debounce(() => {...}, 500, 'typeahead-fetch');
```

mzur (changes requested):
Thanks for the examples. I've got a couple of comments left before I dive into the CSS.

The fest cases for fuzzy search are still missing. Right now you match only the exact volume/label tree name in the tests.

The automatic scrolling works great except in one case. If I press Arrow Down until I got to the bottom of the list and then keep pressing Down a few times, the scrolling does not work correctly when I begin pressing Up afterwards:

[Screencast from 23.10.2024 10:19:48.webm](https://github.com/user-attachments/assets/2b2bfd28-a376-4fd2-a8a3-c2ca24f5f9a3)

resources/assets/js/core/components/typeahead.vue
lehecht: It has to do with the scrolling. If I don't set it to 44px for all items, the highlighted item can be hidden.

lehecht:
> The test cases for fuzzy search are still missing. Right now you match only the exact volume/label tree name in the tests.

I will add a test for the volume fuzzy search, but a label tree fuzzy search test is already [there](https://github.com/biigle/core/blob/587d734fa4956730651056594d07435ae1103ce1/tests/php/Http/Controllers/Api/ProjectLabelTreeControllerTest.php#L52).

resources/assets/js/core/components/typeahead.vue
mzur: You mean the scrolling by arrow keys? Can you maybe use the [clientHeight](https://developer.mozilla.org/en-US/docs/Web/API/Element/clientHeight) of the elements instead? It should be independent of the CSS if possible.

mzur (commented):
Looks good so far (except the CSS issue I'll look into). Let's see if we can get rid of the `isLabelTree` prop first (from last review).

resources/assets/js/core/components/typeahead.vue
lehecht: If I use the clientHeight of the current highlighted element, the scrolling does not work properly any more because of different heights. I already tried different things, but this was the only solution that worked.

lehecht:
@mzur 
Everything works. Thank you!
I've just noticed that label names can be very long now. In this case, scrolling is enabled in x direction.
Bildschirmfoto vom 2024-11-05 07-48-17

mzur (approved):
Thanks. The item name is now also clipped.

Issue #636, opened 2023-08-08
Magic SAM/Lasso sketchFeature opacity


dlangenk:
If opacity is set to e.g. 0.5 and you create a polygon the intermediate result has opacity 1. But for Magic SAM/Lasso the opacity is according to the settings.  I would prefer the opacity=1 solution.

PR #933, opened 2024-09-27
Fix magic wand sketch opacity

lehecht:
Delete sketch feature after adding it to annotation source.

References biigle/core#636.

### DIFF
--- a/resources/assets/js/annotations/ol/MagicWandInteraction.js
+++ b/resources/assets/js/annotations/ol/MagicWandInteraction.js
@@ -183,14 +183,14 @@ class MagicWandInteraction extends PointerInteraction {
     handleUpEvent() {
         this.currentThreshold = this.colorThreshold;
 
-        if (this.isShowingCross) {
-            this.sketchSource.removeFeature(this.sketchFeature);
-        } else {
-            this.dispatchEvent({type: 'drawend', feature: this.sketchFeature});
+        if (!this.isShowingCross) {
+            this.sketchFeature.setStyle(null);
+            this.dispatchEvent({ type: 'drawend', feature: this.sketchFeature });
             // Add feature to annotation source to prevent flickering feature
             this.source.addFeature(this.sketchFeature);
-            this.sketchSource.removeFeature(this.sketchFeature);
         }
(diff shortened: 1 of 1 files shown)


resources/assets/js/annotations/ol/MagicWandInteraction.js
```diff
@@ -188,6 +187,9 @@ class MagicWandInteraction extends PointerInteraction {
             this.sketchSource.removeFeature(this.sketchFeature);
         } else {
             this.dispatchEvent({type: 'drawend', feature: this.sketchFeature});
+            // Add feature to annotation source to prevent flickering feature
+            this.source.addFeature(this.sketchFeature);
+            this.sketchSource.removeFeature(this.sketchFeature);
```
mzur, line 192: The `removeFeature()` can be modes out of the `if` now.

resources/assets/js/annotations/ol/MagicWandInteraction.js
```diff
@@ -188,6 +187,9 @@ class MagicWandInteraction extends PointerInteraction {
             this.sketchSource.removeFeature(this.sketchFeature);
         } else {
             this.dispatchEvent({type: 'drawend', feature: this.sketchFeature});
```
mzur, line 189: I was wondering why there was a delay between the new feature in it's "editing" state (i.e. dashed line) and the "created" state (i.e. solid line) after finishing with the magic wand. The reason is that the interaction sets the editing style directly on the feature and this is moved over to the annotation source, so it keeps the style after `drawend`. Only when the HTTP request returned, the feature is recreated with the proper style.

To fix this, please add `this.sketchFeature.setStyle(null)` above `dspatchEvent()`, so the feature is passed to the event without the editing style.

mzur (changes requested):
Only two comments, the rest is fine! While this still requires the `source` option for the interaction, it solves the opacity issue. Also, the regular draw interaction uses a `source` option as well, so it's fine I guess.

mzur:
Ready for review?

mzur (approved):
Thanks!

Issue #475, opened 2022-08-17
Shortcut to convert line measurement to annotation


mzur:
Add a shortcut for the action to convert a line measurement (<kbd>Shift</kbd>+<kbd>f</kbd>) to an annotation. Suggestion: <kbd>Enter</kbd> (only works if a line measurement is active). Would this conflict with the annotation modes?

mzur:
The measure interaction is defined [here](https://github.com/biigle/core/blob/master/resources/assets/js/annotations/components/annotationCanvas/measureInteraction.vue). And [here](https://github.com/biigle/core/pull/865) is a PR that could serve as example for how to add another keyboard shortcut.

PR #935, opened 2024-10-02
Annotation shortcut

dbrembilla:
This PR resolves issue #475 ; adds a shortcut with _Enter_ when using the ruler / measurement tool to add an annotation.

To test this, try and use the line measurement tool and create an annotation with _Enter_ after having closed the interested area. Please check that this addition does not interfere with other parts of the project (e.g. conflict with other keyboard shortcuts. I checked, and I should have fixed the only problem I found, which was when pressing enter with other instruments)

### DIFF
--- a/resources/assets/js/annotations/components/annotationCanvas/measureInteraction.vue
+++ b/resources/assets/js/annotations/components/annotationCanvas/measureInteraction.vue
@@ -69,6 +69,12 @@ export default {
             this.setMeasureFeature(undefined);
             measureLayer.getSource().clear();
             this.cantConvertMeasureFeature = true;
+        },
+
+        toggleCreateAnnotationFromMeasurement() {
+            if (this.isMeasuring) {
+                this.resetInteractionMode();
+            }
         }
     },
     watch: {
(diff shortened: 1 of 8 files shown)


mzur (changes requested):
Mostly fine, thanks! Only three comments:

- Instead of showing a notification that the line should be finished, I would simply disable/ignore the keyboard shortcut until the measurement line was finished (similar to the "convert" button being disabled then).
- Please add the new shortcut to the manual ([here](https://github.com/biigle/core/blob/master/resources/views/manual/tutorials/annotations/shortcuts.blade.php), maybe add a new section at the bottom).
- The video annotation tool doesn't have a measure tool at all, so no changes are required here.

resources/views/manual/tutorials/annotations/shortcuts.blade.php
```diff
@@ -145,6 +145,25 @@
...
+            <thead>
+                <tr>
+                    <th>Key</th>
+                    <th>Function</th>
+                </tr>
+            </thead>
+            <tbody>
+                <tr>
+                    <td><kbd>Enter</kbd></td>
+                    <td>Annotate selection with label</td>
```
mzur, line 162:
```suggestion
                    <td>Convert measurement to a line string annotation</td>
```

resources/assets/js/annotations/components/annotationCanvas/measureInteraction.vue
```diff
@@ -57,13 +57,14 @@ export default {
...
-                this.clearMeasureFeature();
-            }
+            if (this.isMeasuring && !this.cantConvertMeasureFeature) {
+                if (!this.hasSelectedLabel) {
+                    this.requireSelectedLabel(false);
+                } else {
+                    this.annotationSource.addFeature(this.measureFeature);
+                    this.handleNewFeature({feature: this.measureFeature});
+                    this.clearMeasureFeature();
+                }}
```
mzur, line 67:
```suggestion
                }
            }
```

mzur (approved):
Thanks!

PR #939, opened 2024-10-07
Tile colourspace

mtiessen1175:
Initially convert geoTIFF colourspace in the tiling-process to 0 - 255 colour-band.


mtiessen1175:
used wrong branch for merging.

PR #948, opened 2024-10-11
Fix overflowing long labeltree description

### LINKED ISSUE #925: LabelTrees with long descriptions overflow in Project Labtree view
When label trees contain very long descriptions, the text overflows the box in the project labtree view.

lehecht:
Truncates long labeltree descriptions to prevent overflowing content.

Closes #925.

### DIFF
--- a/resources/assets/js/projects/components/labelTreeListItem.vue
+++ b/resources/assets/js/projects/components/labelTreeListItem.vue
@@ -8,7 +8,7 @@
     </li>
 </template>
 
-<style>
+<style scoped>
 .limit-text {
     word-wrap: break-word;
 }
--- a/resources/assets/js/projects/components/labelTreeListItem.vue
+++ b/resources/assets/js/projects/components/labelTreeListItem.vue
@@ -8,12 +8,6 @@
     </li>
 </template>
 
-<style scoped>
-.limit-text {
-    word-wrap: break-word;
-}
-</style>
-
 <script>
 export default {
     props: {
--- a/resources/assets/js/projects/components/labelTreeListItem.vue
+++ b/resources/assets/js/projects/components/labelTreeListItem.vue
@@ -4,9 +4,17 @@
             <button v-if="editable" v-show="editing" type="button" class="btn btn-default btn-sm pull-right" title="Detach this label tree" @click="emitRemove"><i class="fa fa-trash"></i></button>
             <a :href="url" v-text="tree.name"></a>
         </h4>
-        <p v-if="tree.description" class="list-group-item-text" v-text="tree.description"></p>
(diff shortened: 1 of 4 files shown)


mzur (changes requested):
Thanks! So this is only a problem if there are no spaces in the description and the browser can't break the text, right? This will probably be more of an edge case.

I think the description should be displayed in full in the project list. This is what the description is for, after all. So instead of truncating it all the time (even if it has spaces and can be broken nicely), maybe we should rather use `word-wrap: break-word` instead. This will break long words without spaces so they don't overflow the element.

While you are at it, you could do the same with the label tree title. And also on the label tree overview page:

image

resources/assets/sass/bootstrap/_labels.scss
```diff
@@ -55,4 +55,5 @@ h3, .h3 {
   .small.label {
     font-size: 65%;
   }
+  word-wrap: break-word;
```
mzur, line 58: This affects all `h1`, `h2`, `h3` etc. in BIIGLE. Better use a custom class.

resources/assets/js/projects/components/labelTreeListItem.vue
```diff
@@ -8,6 +8,12 @@
     </li>
 </template>
 
+<style>
```
mzur, line 11:
```suggestion
<style scoped>
```

mzur (changes requested):

resources/assets/js/projects/components/labelTreeListItem.vue
```diff
@@ -8,6 +8,12 @@
     </li>
 </template>
 
+<style scoped>
+.limit-text {
+    word-wrap: break-word;
+}
+</style>
+
```
mzur, line 16: Since you now have a global `limit-text` class (again), this is no longer necessary.

mzur (commented):
I've merged https://github.com/biigle/core/pull/953 into this branch to fix the JS action errors.

resources/assets/js/projects/components/labelTreeListItem.vue
lehecht: Just out of curiosity, do you always know which scss class corresponds to which vue or blade.php file? Or do you use a specific tool to help you identify this? Or are all scss classes accessible in every vue or blade.php file?

mzur: I don't use any tool. I could just see it from the diff of this PR. The SASS files are globally valid, so if you specify `.limit-text` there without additional constraint, it's valid everywhere. In a Vue single-file component, you should usually use `<style scoped>` for exactly the reason that you do not know where a particular class comes from otherwise. If all styles are scoped (except some global ones), you can always look for the class in the same file.

mzur (approved):

PR #950, opened 2024-10-14
Add keyboard shortcut for last annotation

### LINKED ISSUE #499: Select last annotation keyboard shortcut
Implement a keyboard shortcut to select the annotation that was created last (the one with the highest ID). This can be implemented in the image- and video annotation tool.

lehecht:
- [x] Add shortcut to videos
- [x]  Add shortcut to images
- [x] Add shortcut to the manual

Closes #499.

### DIFF
--- a/resources/assets/js/annotations/annotatorContainer.vue
+++ b/resources/assets/js/annotations/annotatorContainer.vue
@@ -19,7 +19,7 @@ import ShapeFilter from './models/ShapeAnnotationFilter';
 import Sidebar from '../core/components/sidebar';
 import SidebarTab from '../core/components/sidebarTab';
 import UserFilter from './models/UserAnnotationFilter';
-import VolumeImageAreaApi from './api/volumes';
+import VolumeImageApi from './api/volumes';
 import {CrossOriginError} from './stores/images';
 import {debounce} from './../core/utils';
 import {handleErrorResponse} from '../core/messages/store';
@@ -551,7 +551,7 @@ export default {
         fetchImagesArea() {
             if (!this.imagesArea) {
                 this.imagesArea = {};
-                VolumeImageAreaApi.get({id: this.volumeId})
+                VolumeImageApi.getArea({id: this.volumeId})
                     .then(this.setImagesArea, handleErrorResponse);
             }
         },
@@ -592,6 +592,19 @@ export default {
             }
             Messages.danger(`Invalid shape. ${shape} needs ${count} different points.`);
         },
+        fetchLastAnnotationId() {
+            VolumeImageApi.getLastAnnotationId({ id: this.image.id })
+                .then((res) => {
+                    let id = parseInt(res.body);
+                    this.annotations.map((a) => {
+                        if (a.id === id) {
+                            a.selected = true;
+                        } else {
+                            a.selected = false;
+                        }
+                    })
+                });
+        }
     },
     watch: {
         async imageId(id) {
--- a/resources/assets/js/annotations/annotatorContainer.vue
+++ b/resources/assets/js/annotations/annotatorContainer.vue
@@ -748,7 +748,7 @@ export default {
             }
         }
 
-        Keyboard.on('Control', this.selectLastAnnotation, 0, this.listenerSet);
+        Keyboard.on('C', this.selectLastAnnotation, 0, this.listenerSet);
     },
     mounted() {
         Events.$emit('annotations.map.init', this.$refs.canvas.map);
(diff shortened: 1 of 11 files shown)


mzur:
I saw that you added a new controller for this. The original idea was to just take the annotation [with the highest ID](https://github.com/biigle/core/issues/499#issue-1435697244) and select it. This does not require a controller etc.

lehecht:
> I saw that you added a new controller for this. The original idea was to just take the annotation [with the highest ID](https://github.com/biigle/core/issues/499#issue-1435697244) and select it. This does not require a controller etc.

You are right. I thought that the annotation list could not be up-to-date at some point, but I was wrong. I reverted all my changes.

resources/assets/js/annotations/annotatorContainer.vue
```diff
@@ -592,6 +592,12 @@ export default {
             }
             Messages.danger(`Invalid shape. ${shape} needs ${count} different points.`);
         },
+        selectLastAnnotation() {
+            let lastAnnotation = this.annotations.reduce((lastAnnotated, a) => a.id > lastAnnotated.id ? a : lastAnnotated, { id: 0 });
+            let remainingAnnotations = this.annotations.filter(a => a.id != lastAnnotation.id);
+            lastAnnotation.selected = true;
+            remainingAnnotations.map(a => a.selected = false);
```
mzur, line 599: Please use `handleSelectAnnotation()` so there are not too many different places that handle selecting of annotations.

```suggestion
            this.handleSelectAnnotation(lastAnnotation);
```

resources/assets/js/videos/videoContainer.vue
```diff
@@ -673,6 +673,12 @@ export default {
             }
             Messages.danger(`Invalid shape. ${shape} needs ${count} different points.`);
         },
+        selectLastAnnotation() {
+            let lastAnnotation = this.annotations.reduce((lastAnnotated, a) => a.id > lastAnnotated.id ? a : lastAnnotated, { id: 0 });
+            let remainingAnnotations = this.annotations.filter(a => a.id != lastAnnotation.id);
+            lastAnnotation.selected = true;
+            remainingAnnotations.map(a => a.selected = false);
```
mzur, line 680: Have you tried this and it works? When an annotation is selected (e.g. in the sidebar) the tool should seek the video to the start of the annotation. In the videoContainer `annotation.selected` is then set to this time. Same as above, it should work fine if you reuse `selectAnnotations()`:

```suggestion
            this.selectAnnotations([lastAnnotation], [], lastAnnotation.startFrame);
```

mzur (changes requested):
Sorry to be this nitpicky about such a small change again but, well, that's my job :grin: 

Is there a specific reason why the shortcut is handled by the annotationCanvas/videoScreen? It could also be handled directly in the annotatorContainer/videoContainer. Then there is no need to pass and handle an extra event.

Also I'm ambivalent about `Ctrl+l`. The `l` shortcut is already associated with the attach/swap tool. While this action is also about "selecting something", it's not directly related to the tool. Maybe we find an unused key and can add the shortcut without modifier (`Ctrl`)?

Finally, the new shortcut must be added to the manual.

lehecht:
> Is there a specific reason why the shortcut is handled by the annotationCanvas/videoScreen?

Because all keyboard events are handled in these components and transfer the event to the parent components. So I just copied what was already there, but it is no problem to change that.

resources/assets/js/videos/videoContainer.vue
lehecht: Thanks! When I tested it, I created multiple annotations in the same frame, which is why I didn't notice that.

lehecht: Did you leave the second parameter empty on purpose? Because this causes the last annotation only to be selected but not to seek the video, if another annotation was already selected on the timeline.

mzur: No. I just wanted to avoid `selectAnnotation()` because this can have unexpected side effects with attaching or swapping labels. You should use:

```js
this.selectAnnotations([lastAnnotation], this.selectedAnnotations, lastAnnotation.startFrame);
```

mzur:
Please change it. We should prefer the solution with less code.

mzur (changes requested):
Now you set <kbd>Ctrl</kbd> as shortcut. This is not what I meant and it messes with other shortcuts, too. I meant that maybe we can find a key other than <kbd>L</kbd> that can be used as shortcut so we don't need the <kbd>Ctrl</kbd> modifier. This would enable users to just press a single key to select the newest annotation. Take a look at the shortcut manual articles, I'm sure there are still unused keys.

mzur (approved):

PR #952, opened 2024-10-16
Improve clone files job

### LINKED ISSUE #944: Make CloneImagesOrVideos more efficient
The CloneImagesOrVideos job calls ProcessNewVolumeFiles to generate thumbnails etc. for the new files of a volume. This can be problematic if the volume consists of many large videos or tiled images. If all of these images/videos are processed at the same time, worker hosts can be overwhelmed.

But the clone job can be made more efficient, as all the thumbnail etc data already exists. Instead of calling ProcessNewVolumeFiles, we should implement new jobs "CloneImageThumbnails" and "CloneVideoThumbnails" (or similar). Each of these jobs gets one old image/video and the respective new (cloned) image/video. The CloneImageThumbnails job then copies the image thumbnail of the old image to the location of the thumbnail of the new image. Also, if the old image is tiled, it copies all the tile data to the location of the new image. If either the thumbnail or tile data does not exist, the job calls ProcessNewImage. Similarly, CloneVideoThumbnails copies all the video thumbnails and the new video preview sprites. If any of these does not exist, it calls ProcessNewVideo (on the appropriate queue, see ProcessNewVolumeFiles).

lehecht:
- [x] Implement CloneImageThumbnails
- [x] Add test for CloneImageThumbnails
- [x] Use CloneImageThumbnails if thumbnails and tiled images exist
- [x] Add tests for tiled images with thumbnails in ProcessNewVolumeFilesTest
- [x] Implement CloneVideoThumbnails
- [x] Add test for CloneVideoThumbnails
- [x] Use CloneVideoThumbnails if thumbnails and sprites exist
- [x] Add tests for videos with thumbnails in ProcessNewVolumeFilesTest

Closes #944.

### DIFF
--- a/app/Jobs/CloneImageThumbnails.php
+++ b/app/Jobs/CloneImageThumbnails.php
@@ -39,30 +39,32 @@ class CloneImageThumbnails extends Job implements ShouldQueue
 
     public function handle()
     {
-        if (!$this->hasThumbnail() || !$this->hasTiledImages()) {
+        $diskThumb = Storage::disk(config('thumbnails.storage_disk'));
+        $diskTiles = Storage::disk(config('image.tiles.disk'));
+
+        if (!$this->hasThumbnail($diskThumb) || !$this->hasTiledImages($diskTiles)) {
             ProcessNewImage::dispatch($this->image);
             return;
         }
 
         $format = config('thumbnails.format');
-        Storage::disk(config('thumbnails.storage_disk'))->copy($this->prefix.".{$format}", $this->copyPrefix.".{$format}");
+        $diskThumb->copy($this->prefix.".{$format}", $this->copyPrefix.".{$format}");
 
-        $disk = Storage::disk(config('image.tiles.disk'));
-        $files = $disk->allFiles($this->prefix);
+        $files = $diskTiles->allFiles($this->prefix);
         foreach ($files as $file) {
             $fileName = str_replace("{$this->prefix}/", "", $file);
-            $disk->copy($file, "{$this->copyPrefix}/{$fileName}");
+            $diskTiles->copy($file, "{$this->copyPrefix}/{$fileName}");
         }
     }
 
-    private function hasThumbnail()
+    private function hasThumbnail($disk)
     {
         $format = config('thumbnails.format');
-        return Storage::disk(config('thumbnails.storage_disk'))->exists("{$this->prefix}.{$format}");
+        return $disk->exists("{$this->prefix}.{$format}");
     }
 
-    private function hasTiledImages()
+    private function hasTiledImages($disk)
     {
-        return Storage::disk(config('image.tiles.disk'))->exists("{$this->prefix}/ImageProperties.xml");
+        return $disk->exists("{$this->prefix}/ImageProperties.xml");
     }
 }
--- a/app/Jobs/CloneImageThumbnails.php
+++ b/app/Jobs/CloneImageThumbnails.php
@@ -2,10 +2,10 @@
 
 namespace Biigle\Jobs;
 
-use File;
 use Illuminate\Contracts\Queue\ShouldQueue;
 use Illuminate\Queue\InteractsWithQueue;
 use Illuminate\Queue\SerializesModels;
+use Illuminate\Support\Facades\File;
 use Storage;
 
 class CloneImageThumbnails extends Job implements ShouldQueue
@@ -24,12 +24,10 @@ class CloneImageThumbnails extends Job implements ShouldQueue
 
     public function handle()
     {
-        $this->copyFiles(Storage::disk(config('thumbnails.storage_disk')));
-        $this->copyFiles(Storage::disk(config('image.tiles.disk')));
-    }
+        $format = config('thumbnails.format');
+        Storage::disk(config('thumbnails.storage_disk'))->copy($this->prefix.".{$format}", $this->copyPrefix.".{$format}");
 
-    private function copyFiles($disk)
-    {
+        $disk = Storage::disk(config('image.tiles.disk'));
         File::copyDirectory($disk->path($this->prefix), $disk->path($this->copyPrefix));
     }
 }
--- a/app/Jobs/CloneImageThumbnails.php
+++ b/app/Jobs/CloneImageThumbnails.php
@@ -41,19 +41,21 @@ class CloneImageThumbnails extends Job implements ShouldQueue
     {
         $diskThumb = Storage::disk(config('thumbnails.storage_disk'));
         $diskTiles = Storage::disk(config('image.tiles.disk'));
+        $hasTiledImages = $this->hasTiledImages($diskTiles);
 
-        if (!$this->hasThumbnail($diskThumb) || !$this->hasTiledImages($diskTiles)) {
+        if (!$this->hasThumbnail($diskThumb) || $this->shouldBeTiled() && !$hasTiledImages) {
             ProcessNewImage::dispatch($this->image);
             return;
         }
-
         $format = config('thumbnails.format');
         $diskThumb->copy($this->prefix.".{$format}", $this->copyPrefix.".{$format}");
 
-        $files = $diskTiles->allFiles($this->prefix);
-        foreach ($files as $file) {
-            $fileName = str_replace("{$this->prefix}/", "", $file);
-            $diskTiles->copy($file, "{$this->copyPrefix}/{$fileName}");
+        if ($hasTiledImages) {
+            $files = $diskTiles->allFiles($this->prefix);
+            foreach ($files as $file) {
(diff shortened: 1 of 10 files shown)


app/Jobs/CloneImageThumbnails.php
```diff
@@ -0,0 +1,41 @@
...
+        $this->copyPrefix = $copyPrefix;
+    }
+
+    public function handle()
+    {
+        $format = config('thumbnails.format');
+        Storage::disk(config('thumbnails.storage_disk'))->copy($this->prefix.".{$format}", $this->copyPrefix.".{$format}");
+
+        $disk = Storage::disk(config('image.tiles.disk'));
+        File::copyDirectory($disk->path($this->prefix), $disk->path($this->copyPrefix));
```
mzur, line 39: This only works for local files. You have to work exclusively with `$disk`.

app/Jobs/CloneVideoThumbnails.php
```diff
@@ -0,0 +1,38 @@
...
+    public function __construct(String $prefix, String $copyPrefix)
+    {
+        $this->prefix = $prefix;
+        $this->copyPrefix = $copyPrefix;
+    }
+
+    public function handle()
+    {
+        $disk = Storage::disk(config('videos.thumbnail_storage_disk'));
+        File::copyDirectory($disk->path($this->prefix), $disk->path($this->copyPrefix));
```
mzur, line 36: Same as above.

app/Jobs/CloneImagesOrVideos.php
```diff
@@ -134,7 +141,7 @@ public function handle()
                 }
             }
             if ($copy->files()->exists()) {
-                ProcessNewVolumeFiles::dispatch($copy);
+                ProcessNewVolumeFiles::dispatch($copy, [], $this->uuidMap);
```
mzur, line 144: You can also use named arguments in cases like these:

```suggestion
                ProcessNewVolumeFiles::dispatch($copy, uuidMap: $this->uuidMap);
```

app/Jobs/ProcessNewVolumeFiles.php
mzur, line 1: I think it would be appropriate here to implement a new job ("ProcessClonedVolumeFiles") for this case and leave ProcessNewVolumeFiles as it is. Considering the comment below, ProcessClonedVolumeFiles will essentially look like ProcessNewVolumeFiles only that it submits CloneImage/VideoThumbnails instead of ProcessNewImage/Video.

app/Jobs/ProcessNewVolumeFiles.php
```diff
@@ -60,10 +72,52 @@ public function handle()
...
+
+                $format = config('thumbnails.format');
+                $hasThumbnail = Storage::disk(config('thumbnails.storage_disk'))->exists($prefix.".{$format}");
+                $hasTiledData = count(Storage::disk(config('image.tiles.disk'))->files($prefix)) > 0;
+                
+                if ($hasThumbnail && $hasTiledData) {
+                    CloneImageThumbnails::dispatch($prefix, $copyPrefix);
+                } else {
+                    ProcessNewImage::dispatch($img);
+                }
```
mzur, line 93: Checking for file existence for all files in a single job will take a very long time. It's better to keep the runtime of individual jobs low and submit more jobs instead. So here you should check for file existence in CloneImageThumbnails/CloneVideoThumbnails. If the files do not exist, call ProcessNewImage/ProcessNewVideo from there.

For this you probably have to pass the full oldImg and newImg models to the CloneThumbnails jobs (instead of just the UUIDs) but that's ok.

app/Jobs/ProcessNewVolumeFiles.php
```diff
@@ -60,10 +72,52 @@ public function handle()
...
+                    ProcessNewImage::dispatch($img);
+                    return;
+                }
+
+                $prefix = fragment_uuid_path($this->uuidMap[$img->uuid]);
+                $copyPrefix = fragment_uuid_path($img->uuid);
+
+                $format = config('thumbnails.format');
+                $hasThumbnail = Storage::disk(config('thumbnails.storage_disk'))->exists($prefix.".{$format}");
+                $hasTiledData = count(Storage::disk(config('image.tiles.disk'))->files($prefix)) > 0;
```
mzur, line 87: Each tiled image directory has an XML file with a fixed name. Just check for this file, then there is no need to enumerate possibly tens of thousands of files.

app/Jobs/ProcessNewVolumeFiles.php
```diff
@@ -60,10 +72,52 @@ public function handle()
...
+                    if (!$this->uuidMap) {
+                        ProcessNewVideo::dispatch($video)->onQueue($queue);
+                        return;
+                    }
+
+                    $prefix = fragment_uuid_path($this->uuidMap[$video->uuid]);
+                    $copyPrefix = fragment_uuid_path($video->uuid);
+
+                    $nbrFiles = count(Storage::disk(config('videos.thumbnail_storage_disk'))->files($prefix));
+                    $nbrSprites = count(array_filter(Storage::disk(config('videos.thumbnail_storage_disk'))
```
mzur, line 109: Just check for the first sprite file.

app/Jobs/ProcessNewVolumeFiles.php
```diff
@@ -60,10 +72,52 @@ public function handle()
...
+                    // Process new video, if volume is just created
+                    if (!$this->uuidMap) {
+                        ProcessNewVideo::dispatch($video)->onQueue($queue);
+                        return;
+                    }
+
+                    $prefix = fragment_uuid_path($this->uuidMap[$video->uuid]);
+                    $copyPrefix = fragment_uuid_path($video->uuid);
+
+                    $nbrFiles = count(Storage::disk(config('videos.thumbnail_storage_disk'))->files($prefix));
```
mzur, line 108: Just check for the first thumbnail.

mzur (changes requested):

app/Jobs/CloneImagesOrVideos.php
lehecht: @mzur 
When this change is applied I get this error: Biigle\Jobs\ProcessCloneVolumeFiles::__construct(): Argument #2 ($only) not passed

mzur: Curious, I thought it would use the default value of `$only` then. You probably have to use the names of all arguments but this is even more verbose than before so forget it.

app/Jobs/CloneImageThumbnails.php
```diff
@@ -24,18 +24,47 @@ class CloneImageThumbnails extends Job implements ShouldQueue
...
 
     public function handle()
+    {
+        $diskThumb = Storage::disk(config('thumbnails.storage_disk'));
+        $diskTiles = Storage::disk(config('image.tiles.disk'));
+
+        if (!$this->hasThumbnail($diskThumb) || !$this->hasTiledImages($diskTiles)) {
+            ProcessNewImage::dispatch($this->image);
+            return;
+        }
```
mzur, line 48: Not all images require tiling. Have a look at `ProcessNewImage::shouldBeTiled()`.

app/Jobs/CloneVideoThumbnails.php
```diff
@@ -24,15 +24,44 @@ class CloneVideoThumbnails extends Job implements ShouldQueue
...
+        $this->copyPrefix = fragment_uuid_path($video->uuid);
     }
 
     public function handle()
     {
         $disk = Storage::disk(config('videos.thumbnail_storage_disk'));
-        File::copyDirectory($disk->path($this->prefix), $disk->path($this->copyPrefix));
+
+        if (!$this->hasThumbnails($disk) || !$this->hasSprites($disk)) {
+            ProcessNewVideo::dispatch($this->video);
```
mzur, line 45:
```suggestion
            ProcessNewVideo::dispatch($this->video);
            return;
```

app/Jobs/ProcessNewVolumeFiles.php
```diff
@@ -2,6 +2,7 @@
 
 namespace Biigle\Jobs;
 
+use Biigle\Image;
```
mzur, line 5:
```suggestion
```

app/Jobs/ProcessNewVolumeFiles.php
```diff
@@ -64,6 +65,7 @@ public function handle()
         } else {
             $queue = config('videos.process_new_video_queue');
             $query->eachById(fn (Video $v) => ProcessNewVideo::dispatch($v)->onQueue($queue));
+
```
mzur, line 68:
```suggestion
```

mzur (changes requested):
Almost done!

mzur (approved):
Thanks!

PR #958, opened 2024-10-22
Enlarge thumbnail image

lehecht:
Increase size to display thumbnail image in full width and height.

Closes biigle/largo#121.

### DIFF
--- a/resources/assets/js/volumes/components/imageGrid.vue
+++ b/resources/assets/js/volumes/components/imageGrid.vue
@@ -118,12 +118,12 @@ export default {
     computed: {
         columns() {
             // This might be 0 if the clientWidth is not yet initialized, so force 1.
-            // The double margin preserves grid functionality by accommodating larger thumbnail sizes.
+            // The double "margin" accounts for the CSS margin, padding and border.
             return Math.max(1, Math.floor(this.clientWidth / (this.width + 2 * this.margin)));
         },
         rows() {
             // This might be 0 if the clientHeight is not yet initialized, so force 1.
-            // The double margin preserves grid functionality by accommodating larger thumbnail sizes.
+            // The double "margin" accounts for the CSS margin, padding and border.
             return Math.max(1, Math.floor(this.clientHeight / (this.height + 2 * this.margin)));
         },
         imagesOffsetEnd() {
--- a/resources/assets/js/volumes/components/imageGrid.vue
+++ b/resources/assets/js/volumes/components/imageGrid.vue
@@ -118,11 +118,13 @@ export default {
     computed: {
         columns() {
             // This might be 0 if the clientWidth is not yet initialized, so force 1.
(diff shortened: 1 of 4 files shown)


mzur (changes requested):
If you add the margin inside the imageGrid component, then there should not be a need to update all instances where the imageGrid is used, right? So essentially leave the props at width: 135 and height: 180 but then in the component do something like this:
```js
computed: {
  actualWidth: () => this.width + this.margin
  actualHeight: () => this.height + this.margin
}
```

I mention this because the imageGrid is not only used in biigle/largo but also in biigle/maia. If implemented as described above we should not need to change anything in these modules.

resources/assets/js/volumes/components/imageGrid.vue
```diff
@@ -118,11 +118,11 @@ export default {
     computed: {
         columns() {
             // This might be 0 if the clientWidth is not yet initialized, so force 1.
-            return Math.max(1, Math.floor(this.clientWidth / (this.width + this.margin)));
+            return Math.max(1, Math.floor(this.clientWidth / (this.actualWidth + this.margin)));
         },
         rows() {
             // This might be 0 if the clientHeight is not yet initialized, so force 1.
-            return Math.max(1, Math.floor(this.clientHeight / (this.height + this.margin)));
+            return Math.max(1, Math.floor(this.clientHeight / (this.actualHeight + this.margin)));
```
mzur, line 125: Since this already uses the margin, can't you do: `this.width + 2 * this.margin`? Then there is no need for a new computed property. Also, please add a comment why the margin is used twice here (because the CSS margin is 8 px and the border+padding is another 8 px).

resources/assets/js/volumes/components/imageGrid.vue
```diff
@@ -118,11 +118,13 @@ export default {
     computed: {
         columns() {
             // This might be 0 if the clientWidth is not yet initialized, so force 1.
-            return Math.max(1, Math.floor(this.clientWidth / (this.actualWidth + this.margin)));
+            // The double margin preserves grid functionality by accommodating larger thumbnail sizes.
```
mzur, line 121:
```suggestion
            // The double "margin" accounts for the CSS margin, padding and border.
```

resources/assets/js/volumes/components/imageGrid.vue
```diff
@@ -118,11 +118,13 @@ export default {
...
+            return Math.max(1, Math.floor(this.clientWidth / (this.width + 2 * this.margin)));
         },
         rows() {
             // This might be 0 if the clientHeight is not yet initialized, so force 1.
-            return Math.max(1, Math.floor(this.clientHeight / (this.actualHeight + this.margin)));
+            // The double margin preserves grid functionality by accommodating larger thumbnail sizes.
```
mzur, line 126:
```suggestion
            // The double "margin" accounts for the CSS margin, padding and border.
```

PR #1004, opened 2024-12-06
Add screenshot button to video annotator

### LINKED ISSUE #956: Screenshot button for videos
Similar to the screenshot button in the image annotation tool, there should be a button in the video annotation tool as well (also with keyboard shortcut). It can probably be implemented in exactly the same way. Only if there is a non-CORS enabled video, the button must be disabled.

lehecht:
Closes #956.


lehecht:
@mzur 
How can I check if my button is disabled for non-CORS enabled videos? Do you know any websites that I could use to test this? Or did you use a different approach for testing?

mzur:
On your local instance you could use the [BIIGLE landing page video](https://s3.bi.denbi.de/swift/v1/BiigleAssets/BiigleLanding.mp4).

resources/assets/js/videos/videoContainer.vue
```diff
@@ -560,7 +580,19 @@ export default {
...
+                    res.blob().then(function (blob) {
+                        let urlCreator = window.URL || window.webkitURL;
+                        self.video.src = urlCreator.createObjectURL(blob);
+                    });
+                })
+                .catch((e) => {
+                    // Access on non-CORS enabled file causes TypeError
+                    this.hasCrossOriginError = e instanceof TypeError;
+                    self.video.src = self.videoFileUri.replace(':id', video.id);
+                });
```
mzur, line 595: This tries to download the whole video. Usually the video is fetched dynamically through a series of HTTP range requests, which is crucial here.

You need another method to detect CORS. Maybe create a small canvas element and try to read some data from the video into it?

resources/assets/js/annotations/annotatorContainer.vue
```diff
@@ -89,6 +89,25 @@ export default {
...
         };
     },
+    provide() {
+        const appData = {}
+
+        let imageInfo = {};
+        imageInfo['currentId'] = biigle.$require('annotations.imageId');
+        imageInfo['ids'] = biigle.$require('annotations.imagesIds');
+        imageInfo['filenames'] = biigle.$require('annotations.imagesFilenames');
+        imageInfo['type'] = 'image';
```
mzur, line 99: Where do you use `type`?

mzur (changes requested):
I think I prefer not to use provide/inject as it hides where the data is coming from. Maybe this is a chance to refactor the screenshot button a bit so it gets all data it needs through props. E.g. if it gets the image/video ID as a prop, it does not even need the image/video changed event any more.

The `map.init` event must stay, though, as the map object must not be passed as a prop. It should never be made reactive.

lehecht:
@mzur 
> Maybe this is a chance to refactor the screenshot button a bit so it gets all data it needs through props

I also thought about it, but then I need to refactor the video and image `settingsTab`, too. The settingsTabs import the `screenshotButton`, which would then need the same properties as the `screenshotbutton`. I am not sure, if they should get props that they will not use except to pass them to the `screenshotbutton`.

mzur:
If it's not too much, I think it is fine to pass the props through the settings tab.

Alternatively you could `provide` the individual properties (e.g. current ID, current filename) instead of a single "files" object. This is at least more descriptive and you can find the location where these are coming from easier than with the generic name. This would still be better than the solution with the event.

resources/assets/js/annotations/annotatorContainer.vue
```diff
@@ -751,7 +754,9 @@ export default {
         Keyboard.on('C', this.selectLastAnnotation, 0, this.listenerSet);
     },
     mounted() {
+        // Event is used in laserpointer module
         Events.$emit('annotations.map.init', this.$refs.canvas.map);
+        this.map = this.$refs.canvas.map;
```
mzur, line 759: The map must not become a reactive property (see previous review). While you can replace the other use of `Event` with props, the `map.init` event listener must stay.

resources/assets/js/annotations/components/screenshotButton.vue
```diff
@@ -9,16 +8,24 @@ import Keyboard from '../../core/keyboard';
...
-    data() {
-        return {
-            filenames: {},
-            currentId: null,
-        };
+    props: {
+        filesObj: {
+            type: Object,
+            default: () => {}
+        },
```
mzur, line 15: The previous name `filenames` was more descriptive.

resources/views/annotations/show.blade.php
```diff
@@ -13,6 +13,7 @@
     biigle.$declare('annotations.shapes', {!! $shapes !!});
     biigle.$declare('annotations.imagesIds', {!! $images->keys() !!});
     biigle.$declare('annotations.imagesFilenames', {!! $images->values() !!});
+    biigle.$declare('annotations.imagesObj', {!! $images !!});
```
mzur, line 16: Since you only need the filenames, you can use `imagesFilenames` instead.

resources/views/videos/show.blade.php
```diff
@@ -67,6 +67,7 @@ class="sidebar-container__content"
     biigle.$declare('videos.isEditor', @can('add-annotation', $video) true @else false @endcan);
     biigle.$declare('videos.videoIds', {!! $videos->keys() !!});
     biigle.$declare('videos.videoFilenames', {!! $videos->values() !!});
+    biigle.$declare('videos.videosObj', {!! $videos !!});
```
mzur, line 70: Since you only need the filenames, you can use `videoFilenames` instead.

resources/assets/js/videos/videoContainer.vue
```diff
@@ -551,7 +554,10 @@ export default {
         },
         fetchVideoContent(video) {
             let videoPromise = new Vue.Promise((resolve) => {
-                this.video.addEventListener('canplay', resolve);
+                this.video.addEventListener('canplay', () => {
+                    this.checkCORSProperty();
+                    resolve();
+                });
             });
```
mzur, line 561: This would be more "promise-y":

```js
let videoPromise = new Vue.Promise((resolve) => {
    this.video.addEventListener('canplay', resolve);
});
videoPromise.then(this.checkCORSProperty);
```

mzur (changes requested):

resources/assets/js/annotations/annotatorContainer.vue
lehecht: Just out of curiosity, why should the map not be made a reactive property?

resources/views/annotations/show.blade.php
lehecht: I need the object, because it maps the video ids to the file names. This ensures that the screenshots get the right file name.

resources/assets/js/annotations/annotatorContainer.vue
mzur: Vue will recursively walk through the whole object (which is very large) and make everything reactive that it can find. It will hook up the reactivity watchers/getters/setters to everything. This will make the whole OpenLayers system slow. With performance-critical paths/objects like this, you have to be very careful with Vue.

resources/views/annotations/show.blade.php
mzur: I still find it redundant with `imagesIds` and `imagesFilenames` just above this line. This will inflate the generated HTML of this template with duplicate information. The image annotation tool produced a screenshot with the correct filename before, too, so there should be a way to do it with the existing information (correct me if I'm wrong).

mzur (approved):
Thanks!

davidsean:
not sure if the disabled for non-CORS is working.
(video hosted on your s3.bi.denbi.de storage)
image
https://biigle.de/videos/33784/annotations

mzur:
@davidsean Thanks, I can confirm that. I haven't checked the cause, yet, but we'll look at it again.

Issue #457, opened 2022-06-02
Highlight settings tab button if image annotation opacity is 0


mzur:
The keyboard shortcut <kbd>o</kbd> sets the annotation opacity to 0 in the image annotation tool. Highlight the settings tab with the `color-info` in this case, so it is clear that there are hidden annotations.

mzur:
The annotation tool settings tab is [here](https://github.com/biigle/core/blob/master/resources/views/annotations/show/tabs/settings.blade.php). You can control the highlight state with the `:highlight` attribute like with the [annotation modes tab](https://github.com/biigle/core/blob/master/resources/views/annotations/show/tabs/annotationModes.blade.php#L1).

The settings state is available in the [annotatorContainer](https://github.com/biigle/core/blob/7cbaa18feed37b8f50bbb9efc27bfc05bbd59f09/resources/assets/js/annotations/annotatorContainer.vue#L66). You could add a new computed property that becomes `true` if the opacity is 0 and then control the `:highlight` attribute with it.

Here is the section of the [developing instructions](https://github.com/biigle/core/blob/master/DEVELOPING.md#3-build-javascript-and-assets) on how to build the JS assets.

PR #1005, opened 2024-12-06
Highlight settings tab when annotation opacity is set to 0

mzur:
Just found this branch.

Resolves https://github.com/biigle/core/issues/457


mzur:
Continued here: https://github.com/biigle/core/pull/1009

Issue #988, opened 2024-11-27
LabelBOT


mzur:
This is about the implementation of LabelBOT. The details have already been discussed. Below is my mockup of the UI. Further discussions can happen here.

![labelbot-ui-1](https://github.com/user-attachments/assets/9669c92c-690b-482c-a842-4b2dd22fef17)

mzur:
Maybe this will finally require a merge of biigle/largo into biigle/core (https://github.com/biigle/core/issues/362). otherwise the required feature vector tables will not be available. Largo is installed in all default BIIGLE instances anyway.

mzur:
Here are some thoughts I just found in my notes:

- When the random/regular sampling mode is active, LabelBOT should immediately suggest a label for the currently active point. When the label is confirmed with <kbd>Enter</kbd> or a click, the annotation mode advances to the next point. The LabelBOT timeout is disabled here, i.e. users always have to confirm the chosen labels. The suggested label for the next point could even be prefetched while the user is still looking at the current point.

- There could be an integrated random/regular sampling and classification workflow at some point. This is started like the laser point detection. It computes random/regular point locations and for each runs LabelBOT to get a label. When the processing is finished they are notified. The user then can review these in the annotation tool using the volume label review mode.

- There is a limit of up to 5 pending LabelBOT requests per user. This limit can be tracked in a cache variable like the video tracking. Requests above this limit are rejected but the UI should handle this gracefully and tell users about this instead of sending actual requests that produce the rejection.

- If computing the feature vector client-side takes a significant amount of time, we could try to compute vectors at fixed intervals already while the user draws the annotation. If the final annotation is reasonably contained in the vector bounding box that was previously computed, immediately send the vector to the backend instead of computing it only after the annotation was finished.

PR #1012, opened 2024-12-12
LabelBOT

gkourie:
# TODO v3 ([#988](https://github.com/biigle/core/issues/988))

## Backend
- [x] Handle the case when LabelBOT returns no result because there are no annotations are associated with any labels of the project.
- [x] A migration file for the HNSW index with [progress bar](https://gist.github.com/djaiss/6685a6033b62e7eb3430b80a3bf258cc).
- [x] There is a limit of up to 5 pending LabelBOT requests per user. This limit can be tracked in a cache variable like the video tracking.
## Frontend
- [x] Fix drawing start point of line feature of the popup for line annotation shapes.
- [x] The LabelBOT button has to be disabled if a remote image without proper CORS configuration is loaded.
- [x] Draggable popup with dashed line connected to the annotation for better visibility when multiple popups are close to each other.
- [ ] (later) ~~If computing the feature vector client-side takes a significant amount of time, we could try to compute vectors at fixed intervals already while the user draws the annotation. If the final annotation is reasonably contained in the vector bounding box that was previously computed, immediately send the vector to the backend instead of computing it only after the annotation was finished.~~
- [x] (later) ~~Implement the ONNX inference as a web worker so the UI does not block while inference is running~~
- [ ] (later) ~~When the random/regular sampling mode is active, LabelBOT should immediately suggest a label for the currently active point. When the label is confirmed with Enter or a click, the annotation mode advances to the next point. The LabelBOT timeout is disabled here, i.e. users always have to confirm the chosen labels. The suggested label for the next point could even be prefetched while the user is still looking at the current point.~~
- [ ] (later) ~~Point to shape conversion with SAM or PTP.~~

### DIFF
--- a/resources/assets/js/annotations/stores/styles.js
+++ b/resources/assets/js/annotations/stores/styles.js
@@ -153,7 +153,9 @@ export default {
         }),
     ],
     editing(feature) {
-        let color = feature.get('color');
+        // Allow this to be called without feature too, as some modules use the default
+        // editing style.
+        let color = feature?.get('color');
         color = color ? ('#' + color) : colors.blue;
 
         if (!styleCache.editing.hasOwnProperty(color)) {


mzur (changes requested):
Please don't implement all the logic as helper methods. Since the logic is used only in the single controller (for now), implement it as (protected) controller methods there.

Also please us camelCase for variable names.

Finally, please fix the failing checks (except the test-current check which fails because biigle/largo is not there). The lint checks can be run locally with `composer lint` and `composer fix`.

app/Http/Controllers/Api/ImageAnnotationController.php
```diff
@@ -236,6 +257,11 @@ public function store(StoreImageAnnotation $request)
 
         $annotation->load('labels.label', 'labels.user');
 
+        // Attach the other two labels if they exist.
+        for ($i = 1; $i < count($topNLabels); $i++) {
+            $annotation->labelBOTLabels[] = $topNLabels[$i];
```
mzur, line 262: The UI will need the full label models (with name and color), not only the IDs.

app/Http/Controllers/Api/ImageAnnotationController.php
```diff
@@ -331,4 +357,108 @@ public function destroy($id)
...
+     * Perform ANN search (HNSW + Post-Subquery-Filtering).
+     *
+     * @param vector $featureVector
+     * @param int[] $trees
+     *
+     * @return array
+     */
+    protected function performAnnSearch($featureVector, $trees)
+    {
+        $featureVector = new Vector($featureVector);
```
mzur, line 408: I think you should also be able to pass a simple array here. No need ti instantiate a `Vector`.

app/Http/Controllers/Api/ImageAnnotationController.php
```diff
@@ -331,4 +357,108 @@ public function destroy($id)
...
+
+        return DB::table(DB::raw("({$subquery->toSql()}) as subquery"))
+            ->setBindings([$featureVector])
+            ->whereIn('label_tree_id', $trees)
+            ->select('label_id')
+            ->groupBy('label_id')
+            ->orderByRaw('MIN(distance)')
+            ->limit(config('labelbot.N')) // N = 3
+            ->pluck('label_id')
+            ->toArray();
```
mzur, line 424: Does this produce the same SQL query than you developed in your thesis? You can check by placing this at the beginning of the method:
```php 
DB::listen(fn ($q) => dd($q->sql));
```
This will dump the SQL when you run a test.

app/Http/Controllers/Api/ImageAnnotationController.php
```diff
@@ -331,4 +357,108 @@ public function destroy($id)
...
+        $subquery = ImageAnnotationLabelFeatureVector::select('label_id', 'label_tree_id')
+            ->selectRaw('(vector <=> ?) AS distance', [$featureVector])
+            ->orderBy('distance')
+            // K = 100
+            ->limit(config('labelbot.K'));
+
+        return DB::table(DB::raw("({$subquery->toSql()}) as subquery"))
+            ->setBindings([$featureVector])
+            ->whereIn('label_tree_id', $trees)
+            ->select('label_id')
```
mzur, line 419: No need for `select()` if you use `pluck()`.

app/Http/Controllers/Api/ImageAnnotationController.php
```diff
@@ -331,4 +357,108 @@ public function destroy($id)
...
+        $featureVector = new Vector($featureVector);
+
+        $subquery = ImageAnnotationLabelFeatureVector::select('label_id', 'label_tree_id')
+            ->selectRaw('(vector <=> ?) AS distance', [$featureVector])
+            ->orderBy('distance')
+            // K = 100
+            ->limit(config('labelbot.K'));
+
+        return DB::table(DB::raw("({$subquery->toSql()}) as subquery"))
+            ->setBindings([$featureVector])
```
mzur, line 417: Is this required again although you already added the binding to the subquery? Maybe there is a better way than to add the subquery as a string.

app/Http/Controllers/Api/ImageAnnotationController.php
```diff
@@ -331,4 +357,108 @@ public function destroy($id)
...
+
+    /**
+     * Perform KNN search (B-Tree + Post-Filtering).
+     *
+     * @param Vector $featureVector
+     * @param int[] $trees
+     *
+     * @return array
+     */
+    protected function performKnnSearch($featureVector, $trees)
```
mzur, line 435: Basically the same comments than above.

app/Http/Controllers/Api/ImageAnnotationController.php
```diff
@@ -331,4 +357,108 @@ public function destroy($id)
...
+                    ->whereIn('project_id', $projectIds);
+            })
+            ->pluck('id')
+            ->toArray();
+
+        return $trees;
+    }
+
+    /**
+     * Perform ANN search (HNSW + Post-Subquery-Filtering).
```
mzur, line 399: Please expand the comment with a brief explanation why we need both ANN and KNN search here.

app/Http/Requests/StoreImageAnnotation.php
```diff
@@ -5,7 +5,7 @@
 use Biigle\Image;
 use Biigle\Shape;
 
-class StoreImageAnnotation extends StoreImageAnnotationLabel
+class StoreImageAnnotation extends StoreImageAnnotationLabelFeatureVector
```
mzur, line 8: No need for inheritance here, just implement the stuff of `StoreImageAnnotationLabelFeatureVector` directly in this class.

app/Http/Requests/StoreImageAnnotationLabelFeatureVector.php
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * Determine if the user is authorized to make this request.
+     *
+     * @return bool
+     */
+    public function authorize()
+    {
+        $this->annotation = ImageAnnotation::findOrFail($this->route('id'));
+        $this->label = Label::findOrFail($this->input('label_id'));
```
mzur, line 33: Won't this fail if no `label_id` is provided?

config/labelbot.php
```diff
@@ -0,0 +1,14 @@
...
+
+    /*
+    | K for KNN
+    */
+    'K' => 100,
+
+    /*
+    | N for top N labels
+    */
+    'N' => 3,
```
mzur, line 13: Please elaborate these comments so you still understand what the variables mean if you come back in 5 years :wink:

mzur (commented):
Just minor comments left now. We have to make sure that we get the exact queries here. Maybe you can dump the two queries as I explained below and post them here for comparison?

app/Http/Controllers/Api/ImageAnnotationController.php
gkourie: Yes it does. 

ANN:

"select "label_id" 
from (
    select "label_id", "label_tree_id", (vector <=> ?) AS distance 
    from "image_annotation_label_feature_vectors" 
    order by "distance" asc 
    limit 100
    ) as "subquery" 
where "label_tree_id" in (?) 
group by "label_id" 
order by MIN(distance) 
limit 3"


KNN:

"select "label_id" 
from (
    select "label_id", "label_tree_id", (vector <=> ?) AS distance 
    from "image_annotation_label_feature_vectors" 
    where "label_tree_id" in (?) 
    order by "distance" asc 
    limit 100
    ) as "subquery" 
group by "label_id" 
order by MIN(distance) 
limit 3"

app/Http/Controllers/Api/ImageAnnotationController.php
mzur: Correct, thanks. I was confused because I thought this worked with a plain array in Largo but it is automatically cast to the correct type there.

app/ImageAnnotation.php
```diff
@@ -19,6 +19,13 @@ class ImageAnnotation extends Annotation
         'points' => 'array',
     ];
 
+    /**
+     * The attributes that should be included in the JSON response.
+     *
+     * @var array<int, string>
+     */
+    protected $appends = ['labelBOTLabels'];
+
```
mzur, line 28: Maybe this should only be appended in the single response from the `store()` controller method. Otherwise it will be included anywhere annotations are used even if it has nothing to do with LabelBOT. You can use `$annotation->append('labelBOTLabels')` at runtime to do this.

mzur (commented):
Looks fine so far (ignoring the missing TODOs). Please go ahead :wink:

gkourie:
@mzur what do you think about the last three points of TODO?

mzur (commented):
The query *must* always be limited to the label trees of the project. Otherwise users might see labels of other users which they are not allowed to see. If no labels exist, the result is empty and that's ok. The UI will show an appropriate message if no label exists (i.e. LabelBOT can't be activated).

If the label tree contains only one label then the existing search will automatically return this label, right? No need to implement a special case for this.

gkourie:
okay, thanks.

mzur (commented):
Where do I get the model file, i.e. which did you use? This should be mentioned in the config file (see below).

I will review the UI once it is implemented according to the mockup in https://github.com/biigle/core/issues/988. If you want feedback before that, we should meet and you can show me what you currently have.

We also need a migration file for the HNSW index. Maybe this migration could also implement a [progress bar](https://gist.github.com/djaiss/6685a6033b62e7eb3430b80a3bf258cc) that shows the progress of the index creation? This also has to be mentioned in the upgrade instructions of the release. We should prepare a text for that. See [here](https://github.com/biigle/biigle/releases/tag/v1.3.0) for an example of upgrade instructions.

mzur (commented):
Thanks! High-level comments first:

- Only initialize the ONNX runtime and load the ONNX model when LabelBOT is first activated. Loading all this is not necessary when users still want to annotate manually.

- There is no need for the ONNX model API controller (which also causes a failing test right now). Just load the model directly from a public URL. This can be the URL that is already configured in the config. While we want to release the ONNX file as a release asset, we can't directly use GitHub as URL to download the model (because of CORS). So we have to add instructions for instance admins where to get the model file, where to put it for their BIIGLE instance and how to configure the public URL.

- I wondered if we can use the same code to load the DINO model than in biigle/largo:

   ```python
   dinov2_vits14 = torch.hub.hub_load('facebookresearch/dinov2:main', 'dinov2_vits14')
   ```

   I tried to export this and it worked but with warnings. The file was successfully initialized in JS, too, but the model inference just never returned anything. So you have any idea why this doesn't work? This is important IMO since the backend and frontend models need to match exactly.

- Add a readme and `requirements.txt` for the ONNX export script.

- The onnxruntime-web WASM is put to the public directory on build. It should be put to public/assets.

- In the labelbot.vue mixin, is the input image normalized? Compare this with how the feature vector is generated in [biigle/largo](https://github.com/biigle/largo/blob/efab001365ea4bad0471f3be974beead6e655863/src/resources/scripts/ExtractFeatures.py#L32-L36).

- The timer/progress bar in the LabelBOT popup should also stop if the user just hovers over any of the label candidates.

- We also need some keyboard shortcuts for the popup. For example, <kbd>Esc</kbd> closes the "focused" popup (where the mouse recent popup is focused, when this is closed, the second most recent popup is focused etc.). <kbd>Enter</kbd> could confirm the top 1 label. <kbd>Return</kbd> already deletes the entire annotation as it should. Arrow keys combined with <kbd>Enter</kbd> should allow the selection of any of the top 3 labels and maybe even put the focus on the typeahead at the bottom.

- Is it correct that there is no HNSW index at all, yet, because the migration is missing?

- Labelbot can be activated even if model loading fails. The dot is yellow but I can still create annotations, causing lots of JS errors.

- The popup should be 300px wide, similar to the sidebar. This way label names are not broken to a new line so often.

mzur:
I this ready for review or are you still working on it?

gkourie:
> I this ready for review or are you still working on it?

Oh sorry, i made just some small changes. please keep going..

mzur:
Did you push the `dev-modules` branch instead of the `labelbot` branch? Suddenly all sorts of commits from `dev-modules` are included as well.

gkourie:
> Did you push the `dev-modules` branch instead of the `labelbot` branch? Suddenly all sorts of commits from `dev-modules` are included as well.

oops, now the history is clean

mzur (commented):
- When I first enabled LabelBOT, the ONNX inference was extremely slow. At some point after a refresh it was fast again and stayed fast. Maybe we should implement a "warmup" of the model as it is done with Magic SAM? This is just an inference pass with empty data.

- I also thought about implementing the ONNX inference as a web worker so the UI does not block while inference is running. But we could leave this as an idea for later.

- The WASM file now seems to be put into the build/assets/ directory as it should. Did you fix this?

- The <kbd>Enter</kbd> shortcut does not work with an open LabelBOT popup unless I use the arrow keys first.

- When I use the arrow keys in the popup and arrive at the typeahead at the bottom, I can no longer move up again or use <kbd>Esc</kbd>.

- The LabelBOT button has to be disabled if a remote image without proper CORS configuration is loaded.

- When I run `npm install` the `package-lock.json` is updated. Isn't it up to date in your branch?

- I decided to tweak the LabelBOT indicator style a bit and use the proper colors of the theme. What do you think?

gkourie:
> * The WASM file now seems to be put into the build/assets/ directory as it should. Did you fix this?

No, I didn't fix it, maybe after upgrading to vue 3?

> * When I use the arrow keys in the popup and arrive at the typeahead at the bottom, I can no longer move up again or use Esc.

The `input` tag is ignored by keyboard events [keyboard.js](https://github.com/biigle/core/blob/33ec2388e6ad7c7983a03bc61c9a37a33173ae69/resources/assets/js/core/keyboard.js#L29C1-L34C3). You can leave the `input`  by pressing `Tab`.

> * When I run npm install the package-lock.json is updated. Isn't it up to date in your branch?

Another merge oversight 😅

> * I decided to tweak the LabelBOT indicator style a bit and use the proper colors of the theme. What do you think?

It also looks legit. I only changed the padding to centralize the circle.

vite.config.js
```diff
@@ -59,4 +59,7 @@ export default defineConfig({
             external: ['vue'],
         },
     },
+    optimizeDeps: {
+        exclude: ["onnxruntime-web"],
+    },
```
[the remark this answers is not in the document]
gkourie: To load the onnx model, otherwise we get the error:

`GET http://127.0.0.1:5173/node_modules/.vite/deps/ort-wasm-simd-threaded.jsep.mjs
NS_ERROR_CORRUPTED_CONTENT

Loading module from “http://127.0.0.1:5173/node_modules/.vite/deps/ort-wasm-simd-threaded.jsep.mjs” was blocked because of a disallowed MIME type (“”).

Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at http://127.0.0.1:5173/node_modules/.vite/deps/ort-wasm-simd-threaded.jsep.mjs. (Reason: CORS request did not succeed). Status code: (null).`

resources/assets/js/annotations/annotatorContainer.vue
```diff
@@ -440,35 +427,43 @@ export default {
...
+                                    this.showLabelbotPopup(createdAnnotation);
+                                }
+                                return createdAnnotation;
+                            })
+                            .then(this.setLastCreatedAnnotation);
+                    })
+                    .catch((e) => {
+                        if (this.isLabelbotOn) {
+                            this.updateLabelbotState(LABELBOT_STATES.OFF);
+                            Messages.danger("There are no annotations associated with any labels in this project!");
```
mzur, line 461: This might show a misleading message if something else went wrong while LabelBOT was active. You should only show this if LabelBOT did not return any suggestions. If no annotations are present in the project, users should not be able to activate LabelBOT in the first place.

resources/assets/js/annotations/mixins/labelbot.vue
```diff
@@ -43,29 +63,40 @@ export default {
...
+                        .then((model) => {
+                            this.labelbotModel = model;
+                            this.warmUpLabelbotModel();
+                            this.labelbotState = LABELBOT_STATES.READY;
+                        })
+                        .catch((error) => {
+                            this.labelbotState = LABELBOT_STATES.OFF;
+                            handleErrorResponse(error);
+                        })
                 })
```
mzur, line 99: Maybe put this in a cleaner promise chain like this:

```js
InferenceSession.create('webgpu')
   .catch(() => InferenceSession.create('wasm'))
   .then(model => this.labelbotModel = model)
   .then(this.warmUpLabelbotModel)
   .then(() => this.labelBotState = READY)
   .catch(...);

File: resources/assets/js/core/components/powerToggle.vue [hunk 15]
```diff
@@ -61,9 +65,11 @@ export default {
         currentTitle() {
             if (this.active) {
                 return this.titleOn || this.title;
+            } else if (this.disabled) {
+                return this.titleDisabled || this.title;
+            } else {
+                return this.titleOff || this.title;
```
mzur, line 71: I wouldn't add an extra prop just for one use case here. I don't think the separate props for on/off titles were a good idea in the first place. We can just modify the title prop dynamically. You can either do this here, too, or use two conditional components like this:

```html
<power-toggle v-if="labelBotEnabled" title="My enabled title"></power-toggle>
<power-toggle v-else :disabled="true" title="My disabled title"></power-toggle>
```

gkourie:
> I don't like the UX of that. Maybe let the user switch only between the top 3 labels with arrow keys and add Tab to jump in/out of the typeahead?

To do so we need to:

1.  modify the [toggleLastOpenedTab](https://github.com/biigle/core/blob/69f21c5c0d02e885c9cb78cc3b7829980848f877/resources/assets/js/core/components/sidebar.vue#L109-L120) function in the sidebar to check if any LabelBOT popup is focused. Otherwise the two `keydown tab` events will interfere with each other, and also we need to add the `keydown.esc`/ `keydown.tab` event to the [typeahead](https://github.com/biigle/core/blob/69f21c5c0d02e885c9cb78cc3b7829980848f877/resources/assets/js/core/components/typeahead.vue) to allow exiting the input and highlighting the last label,
2. Modify the `Keyboard` class to include input elements in the events or
3. Let user select/focus the `typeahead` only via mouse click

resources/assets/js/annotations/components/annotationCanvas/drawInteractions.vue
```diff
@@ -48,12 +49,15 @@ export default {
...
             return this.interactionMode === 'drawEllipse';
         },
+        isLabelbotOn() {
+            return this.labelbotState !== LABELBOT_STATES.OFF && this.labelbotState !== LABELBOT_STATES.DISABLED;
+        },
     },
     methods: {
         draw(name) {
-            if (this['isDrawing' + name] || (this.labelbotIsOn && this.labelbotState === 'busy')) {
+            if (this['isDrawing' + name] || this.labelbotState === LABELBOT_STATES.BUSY) { // When LabelBOT is busy (max number of requests is reached) no drawing is allowed until it is ready again or turned off.
```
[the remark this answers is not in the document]
gkourie: > What happens when I draw an annotation with LabelBOT and then disable LabelBOT while the result is still pending? Will I still see the popup and is the annotation updated?

No, the popup will not appear but the annotation will be saved with Top-1 label.  But this behavior can be changed.

gkourie:
> We also need a migration file for the HNSW index. Maybe this migration could also implement a [progress bar](https://gist.github.com/djaiss/6685a6033b62e7eb3430b80a3bf258cc) that shows the progress of the index creation? This also has to be mentioned in the upgrade instructions of the release. We should prepare a text for that. See [here](https://github.com/biigle/biigle/releases/tag/v1.3.0) for an example of upgrade instructions.

I don't think we can show the progress bar directly in the migration step because the `CREATE INDEX` Statement needs to be finished first. I made separate Command to show the progress bar.

mzur (commented):
I've now started to tweak some UI stuff. What I did:

- Tweak the highlight state of the popups
- Added title attributes to label items and typeahead
- Made arrow shortcut keys cyclic

What I still want to do:

- Tweak the LabelBOT button
- Maybe work on the line placement during popup dragging
- Add a DB model and API endpoint to track some statistics on LabelBOT performance (i.e. how often was top1, top2, top3 chosen, how often a different label, how often was popup dismissed etc.)

Please have a look at [my changes](https://github.com/biigle/core/pull/1012/files/32b0726a6fdc9e0dc856d68d40109426cd940d15..f42e80d99cd457750294a03d71df02b9da75b599) for any errors.

Now to your changes:

- The moving of popups is really neat and the keyboard shortcuts work great now!
- Maybe we should disable the timeoout/progress bar if the popup is moved.
- <kbd>Esc</kbd> should delete the annotation and close the popup, too.
- <kbd>Backspace</kbd> should delete the annotation, not <kbd>Del</kbd>. This is more consistent with the normal annotation workflow.
- Sometimes the progress bar is immediately at 100%. Not sure if this is because of some of my changes or if you observed this before, too.

app/Http/Controllers/Views/Annotations/AnnotationToolController.php
```diff
@@ -86,7 +84,7 @@ public function show(Request $request, $id)
             'shapes' => $shapes,
             'annotationSessions' => $annotationSessions,
             'tilesUriTemplate' => $tilesUriTemplate,
-            'annotationCount' => $annotationCount,
+            'annotationsExist' => $annotationsExist,
```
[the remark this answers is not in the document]
gkourie: It's permanently disabled until a page reload.

gkourie:
> Please have a look at [my changes](https://github.com/biigle/core/pull/1012/files/32b0726a6fdc9e0dc856d68d40109426cd940d15..f42e80d99cd457750294a03d71df02b9da75b599) for any errors.
> 
Thanks. I just removed the overflow property to list the label items in the input tag.


> Esc should delete the annotation and close the popup, too. Backspace should delete the annotation, not Del. This is more consistent with the normal annotation workflow.

Why both keys `Esc` and `Backspace` should have the same function?
I thought `Esc` should only close the popup keeping the annotation with Top-1 label, which make more sense for me.

resources/assets/js/annotations/annotatorContainer.vue
```diff
@@ -425,16 +425,25 @@ export default {
...
                     this.updateLabelbotState(LABELBOT_STATES.COMPUTING);
 
+                    availableOverlayKey = this.getAvailableLabelbotOverlay();
+
+                    if (availableOverlayKey === -1) {
+                        this.$nextTick(removeCallback);
+                        Messages.danger(`You already have ${this.labelbotOverlays.length} pending LabelBOT requests. Please wait for one to complete before submitting a new one.`)
+                        return;
+                    }
+
```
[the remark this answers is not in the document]
mzur: I'd say if users want to keep as many popups open, it's their problem :stuck_out_tongue_winking_eye: But a limit of 20 popups sounds reasonable, too.

resources/assets/js/annotations/mixins/labelbot.vue
```diff
@@ -0,0 +1,400 @@
...
+            const annotationDataArray = new Float32Array(size * 3);
+
+            // Image normalization
+            const mean = [0.485, 0.456, 0.406];
+            const std = [0.229, 0.224, 0.225];
+
+            for (let i = 0; i < size; i++) {
+                annotationDataArray[i] = ((annotationData[i * 4] / 255.0) - mean[0]) / std[0]; // R
+                annotationDataArray[size + i] = ((annotationData[i * 4 + 1] / 255.0) - mean[1]) / std[1];  // G
+                annotationDataArray[2 * size + i] = ((annotationData[i * 4 + 2] / 255.0) - mean[2]) / std[2];  // B
```
mzur, line 161: @gkourie Could you explain why the layout of `annotationDataArray` is `rrrgggbbb` instead of `rgbrgbrgb`? Your code seems to be correct but I'm confused by the unusual layout.

gkourie: It's the standard dimension layout for ONNX inference tensors, as shown in the [ONNX image classification example](https://onnxruntime.ai/docs/tutorials/web/classify-images-nextjs-github-template.html). The planar layout (rrrgggbbb) is supposed to be more efficient when executing ONNX Conv operators [conv](https://onnx.ai/onnx/operators/onnx__Conv.html).

mzur: Great thanks! I was wondering why I never came across this although I already worked with ORT-web in Magic SAM. But Magic SAM uses the image embedding as input, not the raw image data.

mzur:
Collecting TODOs:

- [ ] Implement LabelBOT for tiled images.
   - As a first step it could be disabled for tiled images.
   - LabelBOT could extract a screenshot of the OpenLayers canvas for inference if a tiled image is shown (regular images keep using the original). This could be done in the annotationCanvas, i.e. if LabelBOT is active the annotationCanvas passes on either the image patch or the whole feature vector in the event to the annotatorContainer along with the OpenLayers feature. LabelBOT for videos can work the same way using the original video.
- [ ] Implement LabelBOT for videos.
- [x] ~~Implement support for LabelBOT in the random and regular sampling modes.~~ Already supported as-is!

mzur (approved):

Issue #70, opened 2017-01-25
Use Imagick/Gmagick


mzur:
For possible future support of TIFF and other formats and maybe to address #69 we can think about using Imagick or Gmagick instead of php-gd. InterventionImage (our PHP graphics processing library) supports [Imagick](https://www.imagemagick.org/script/index.php) out of the box. The people of [Gmagick](http://www.graphicsmagick.org/) claim their library would be better, though.

First, check if we want to use Gmagick over Imagick. Then check if we need InterventionImage or if it can be configured to run with Gmagick, too.

We use InterventionImage to have automatic image caching for remote volumes (among other things) so it might be nice to be able to keep it.

PR #1014, opened 2024-12-16
Add video info view

ToukL:
As stated in [Footprint / laser point detection for videos #70](https://github.com/biigle/laserpoints/issues/70), we would like to add a new info view for videos, as for images, with metadata and, later, computed laser points info.

### DIFF
--- a/resources/assets/js/volumes/components/metadataModal.vue
+++ b/resources/assets/js/volumes/components/metadataModal.vue
@@ -1,0 +1,76 @@
+<template>
+    <modal
+        id="modal-show-metadata"
+        ref="modal"
+        v-model="show"
+        size="lg"
+        :title="name"
+        :footer="false"
+        >
+        <div class="panel panel-default table-responsive">
+            <table class="table">
+                <thead>
+                    <tr>
+                        <th v-for="time in times">{{ time }}</th>
+                    </tr>
+                </thead>
+                <tbody>
+                    <tr v-if="items">
+                        <td v-for="item in items">{{ item }}</td>
+                    </tr>
+                </tbody>
+            </table>
+        </div>
+    </modal>
+</template>
+
+<script>
+import { Modal } from 'uiv';
+
+export default {
+    components: {
+        modal: Modal,
+    },
+    props: {
+        showModal: {
+            required: true,
+            type: Boolean,
+        },
+        times: {
+            required: true,
+            type: Array,
+        },
+        items: {
+            required: true,
+            type: Array,
+        },
+        name: {
+            required: true,
+            type: String,
+        },
+    },
+    data() {
+        return {
+            show: false,
+        };
+    },
+    watch: {
+        // if volume-metadata-button pressed, trigger modal
+        showModal: function () {
+            if (this.showModal) {
+                this.show = true;
+            }
+        },
+        // if modal is closed, trigger the close-modal-event, which sets 'showModal' in parent container to false again
+        show: function() {
+            if (this.show === false) {
+                this.$emit('close-modal');
+            }
+        }
+    },
+    created() {
+        this.show = false;
+        this.$emit('load-modal');
+    }
+}
+</script>
--- a/resources/assets/js/volumes/components/metadataModal.vue
+++ b/resources/assets/js/volumes/components/metadataModal.vue
@@ -1,22 +1,28 @@
 <template>
     <modal
-        id="modal-show-metadata"
         ref="modal"
         v-model="show"
-        size="lg"
+        size="sm"
         :title="name"
         :footer="false"
         >
-        <div class="panel panel-default table-responsive">
+        <div class="panel panel-default">
             <table class="table">
                 <thead>
                     <tr>
-                        <th v-for="time in times">{{ time }}</th>
+                        <th class="text-center" v-if="times">{{ "Time" }}</th>
+                        <th class="text-center" v-if="name !== 'Times'">{{ name }}</th>
                     </tr>
                 </thead>
-                <tbody>
-                    <tr v-if="items">
-                        <td v-for="item in items">{{ item }}</td>
+                <tbody v-if="name === 'Times'">
+                    <tr v-for="time in times">
+                        <td class="text-center">{{ time }}</td>
+                    </tr>
+                </tbody>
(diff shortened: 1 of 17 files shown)


routes/web.php
```diff
@@ -335,7 +340,7 @@
         ]);
 
         // Legacy support.
-        $router->redirect('videos/{id}', '/videos/{id}/annotations');
+        $router->redirect('video-annotate/{id}', '/videos/{id}/annotations');
```
mzur, line 343: This redirect doesn't make sense any more. I can be removed.

resources/views/volumes/videos/index/meta.blade.php
```diff
@@ -0,0 +1,51 @@
...
+                    <th>{{ $metadataMap[$key] }}</th>
+                    @if (is_array($field))
+                        @foreach ($field as $value)
+                            <td>{{ $value }}</td>
+                        @endforeach
+                    @else
+                        <td>{{ $value }}</td>
+                    @endif
+                </tr>
+            @endforeach
```
mzur, line 48: Video metadata may be an array, too. This will produce an "array to string conversion" error here. How could we show these values here in a meaningful way? Maybe show the min/max values of each field only?

resources/views/volumes/videos/index/meta.blade.php
```diff
@@ -0,0 +1,51 @@
+<div class="panel panel-default">
+    <div class="panel-heading">
+        <h3 class="panel-title">Video information</h3>
+    </div>
+    <div class="panel-body table-responsive">
```
mzur, line 5: No need to wrap this into a `.panel-body`, this will only add an unnecessary padding around the table (see images/index/meta.blade.php).

mzur (changes requested):
Thanks, almost perfect!

While reviewing your code I found https://github.com/biigle/core/issues/1016 so if you feel like it, you could fix that here, too (but you don't have to 😉).

resources/views/volumes/videos/index/meta.blade.php
ToukL: That is why I added the `@if(is_array)` loop for `taken_at` and each metadata field. I tested with attached metadata file as array and it works. The result is quite heavy so yes, maybe we could print only the min/max or first/last values, what do you think ?

Capture

[biigle-volume-6-metadata.csv](https://github.com/user-attachments/files/18276307/biigle-volume-6-metadata.csv)

ToukL: Yes a modal seems a good idea to me, I can take example on the popup showing statistics on a project ?

mzur (commented):
No complaints! Two comments of the previous review are still open.

And don't mind the failing lint check, that's my fault :wink:

resources/assets/js/volumes/components/metadataModal.vue
```diff
@@ -0,0 +1,76 @@
+<template>
+    <modal
+        id="modal-show-metadata"
```
mzur, line 3: Is this ID required for anything?

resources/views/volumes/videos/index/meta.blade.php
```diff
@@ -22,28 +22,38 @@
...
+                    <td>
+                        <button class="btn btn-default" type="button" title="Show full timestamps" v-on:click.prevent="showTimes()">Show values</button>
+                    </td>
+                </tr>
+                @foreach ($metadata as $field => $value)
+                    <tr>
+                        <th>{{ $metadataMap[$field] }}</th>
+                        @if (is_array($value))
+                            <td>
+                                <button class="btn btn-default" type="button" title="Show full metadata array" v-on:click.prevent="getMetadata({{ json_encode($metadataMap[$field]) }}, {{ collect($value) }})">Show values</button>
```
mzur, line 38: I find the buttons a bit bulky. Maybe make the field name (e.g. "Created") clickable as a link element? Or just make "show values" a link instead of a button?

mzur (commented):
This works great! In addition to the points below I have two comments which are more like my opinion rather than requirements, so please feel free to argue against :wink: 

First, you show the timestamps in the first row and the respective metadata values in the second row. With lots of values, users have to scroll horizontally. However, users are more used to scrolling vertically. So I would instead show the timestamps in the first column and the values in the second column. This also gives you the chance to add column headers (e.g. "time", "area") in the first row.

Second, you add the metadata values as attributes to the HTML elements. This produces "messy" HTML with lots of escape characters. Example:

```
v-on:load-modal="getTimes([&quot;2016-12-19T11:27:00.000000Z&quot;,&quot;2016-12-19T11:28:00.000000Z&quot;,&quot;2016-12-19T11:29:00.000000Z&quot;,&quot;2016-12-19T11:30:00.000000Z&quot;,&quot;2...
```

In cases like these where I want to pass JS values in the template rather than loading them with an extra API request, I usually do it like this:

```blade
@push('scripts')
   <script type="text/javascript">
        biigle.$declare('videos.metadata', {!! collect($video->metadata) !!});
   </script>
@endpush
```

Then you can use the whole metadata object in `volumeMetadata.vue` (or better call it `videoMetadata.vue`?) like this:

```js
created() {
   this.metadata = biigle.$require('videos.metadata');
}
```

resources/views/volumes/videos/index/meta.blade.php
ToukL: No problems, I agree with all your comments :)

For the buttons I tried using field name as link elements, but it gives weird results with table's borders (at least on chrome, see below). So maybe the second option ? The thing is : if I use `class="btn btn-link"` the padding is not the same as for the rest of the table. So I tried with `class="btn-link"` and it works but is it a thing ?

```suggestion
                                <a class="btn-link" v-on:click.prevent="showMetadata({{ json_encode($field) }})">Show values</a>
```

Capture

mzur: Your screenshot looks like you removed the `<td>` elements completely. It should work if you change the text in the `<th>` to be a link and keep an empty `<td>` as next element.

mzur (approved):
I've added some final tweaks:

- Increased size of modal to `md`
- Fixed flashing of empty table in closing modal
- Updated modal open links to look like actual links
- Removed legacy route that no longer made sense

PR #1019, opened 2024-12-19
Video single frame annotation

### LINKED ISSUE #786: Settings option to create single-frame video annotations by default
Add a settings option to the video annotation tool to automatically finish new annotations after the first frame. If the option is active, the additional control buttons to finish or track an annotation should be hidden (or disabled?) from the UI.

aneust:
Add a settings option to the video annotation tool to automatically finish new annotations after the first frame

- Add button to settings
- Add description in manual
- Disabled "Finish" and "Track Annotation" control buttons, when option is activated
- Prevented double-click behavior for point annotations

closes #786


mzur:
@aneust Was this closed on purpose?

aneust:
@mzur Closed on purpose, Leane found some strange behavior I first could not reproduce. Think now it was caused by the settings.

mzur:
Ok, thanks. You could also keep this as a Draft PR while you keep working on the issue.

lehecht (changes requested):

lehecht (changes requested):

lehecht (commented):
Looks good to me now. Please request a review from @mzur.

mzur:
I saw the mention and self-requested the review :wink: @lehecht You can also request the review from me when you are done.

resources/assets/js/videos/components/videoScreen/drawInteractions.vue
```diff
@@ -199,13 +213,19 @@ export default {
             let lastFrame = this.pendingAnnotation.frames[this.pendingAnnotation.frames.length - 1];
 
             if (lastFrame === undefined || lastFrame < this.video.currentTime) {
-                this.pendingAnnotation.frames.push(this.video.currentTime);
-                this.pendingAnnotation.points.push(this.getPointsFromGeometry(e.feature.getGeometry()));
+                if (this.singleAnnotation && this.isPointDoubleClick(e)) { 
+                    this.pendingAnnotationSource.once('addfeature', function (e) {
```
mzur, line 217: Please add the explanation:

```
// The feature is added to the source only after this event
// is handled, so remove has to happen after the addfeature
// event.
```

resources/assets/js/videos/components/videoScreen/drawInteractions.vue
```diff
@@ -199,13 +213,19 @@ export default {
...
+                        this.removeFeature(e.feature);
+                    });
```
mzur, line 219: Shouldn't you return here and skip the rest of the function?

resources/assets/js/videos/components/videoScreen/drawInteractions.vue
```diff
@@ -7,21 +7,35 @@ import VectorLayer from '@biigle/ol/layer/Vector';
...
 
 /**
  * Mixin for the videoScreen component that contains logic for the draw interactions.
  *
  * @type {Object}
  */
+
+const POINT_CLICK_COOLDOWN = 400;
+const POINT_CLICK_DISTANCE = 5;
+
```
mzur, line 21: Since `computeDistance` and the constants are used both in the image annotation tool and the video annotation tool, please create a new file for these and reuse the same function/constants in both tools.

resources/assets/js/videos/components/videoScreen/drawInteractions.vue
```diff
@@ -199,13 +213,19 @@ export default {
             let lastFrame = this.pendingAnnotation.frames[this.pendingAnnotation.frames.length - 1];
 
             if (lastFrame === undefined || lastFrame < this.video.currentTime) {
-                this.pendingAnnotation.frames.push(this.video.currentTime);
-                this.pendingAnnotation.points.push(this.getPointsFromGeometry(e.feature.getGeometry()));
+                if (this.singleAnnotation && this.isPointDoubleClick(e)) {
```
mzur, line 216: Don't you have to check `this.isDrawingPoint` here, too?

resources/assets/js/videos/components/videoScreen/drawInteractions.vue
```diff
@@ -217,6 +237,22 @@ export default {
...
+                if (this.isDrawingPoint) {
+                    if (this.isPointDoubleClick(e)) {
+                        this.resetPendingAnnotation(this.pendingAnnotation.shape);
+                        return;
+                    }
+                    this.lastDrawnPointTime = new Date().getTime();
+                    this.lastDrawnPoint = e.feature.getGeometry();
+                }
+                this.pendingAnnotationSource.once('addfeature', this.finishDrawAnnotation);
+            }
```
mzur, line 251: I feel all this can be added to the `if` above when refactored appropriately. This also would add the ´return` I asked for above.

resources/views/videos/show/sidebar-settings.blade.php
```diff
@@ -52,6 +52,10 @@
                 <div class="sidebar-tab__section">
                     <power-toggle :active="muteVideo" title-off="Mute video" title-on="Unmute video" v-on:on="handleMuteVideo" v-on:off="handleUnmuteVideo">Mute Video</power-toggle>
                 </div>
+
+                <div class="sidebar-tab__section">
+                    <power-toggle :active="singleAnnotation" title-off="Enable Single-Frame Annotation" title-on="Disable Single-Frame Annotation" v-on:on="handleSingleAnnotation" v-on:off="handleDisableSingleAnnotation">Single-Frame Annotation</power-toggle>
```
mzur, line 57: The title is not very helpful. What about "Enable always creating single-frame annotations" (and "Disable ...")?

mzur (commented):
Two comments about the UI:

- The setting doesn't affect whole-frame annotations but it should.

- To be more user-friendly, the "finish annotation" and "track annotation" buttons (which are disabled) could get titles like "Disable the single-frame annotation option to create multi-frame annotations" ("Disable the single-frame annotation option to track annotations"). This way users are reminded of what to do if they want to use the buttons again. You could implement this with `v-if` showing a completely different set of buttons which are always disabled and have the new title.

Otherwise it works great. See some comments about the code below.

resources/assets/js/videos/components/videoScreen.vue
```diff
@@ -429,6 +488,12 @@ export default {
         jumpForwardMessage() {
             return `Advance video by ${this.jumpStep} s 𝗖𝘁𝗿𝗹+𝗥𝗶𝗴𝗵𝘁 𝗮𝗿𝗿𝗼𝘄`;
         },
+        disableFinishTrackAnnotation() {
+            return this.cantFinishTrackAnnotation || this.disableJobTracking || this.singleAnnotation
+        },
+        disableFinishDrawAnnotation() {
+            return this.cantFinishDrawAnnotation || this.singleAnnotation
+        },
```
mzur, line 496: These are probably no longer required.

mzur (changes requested):
While the setting now affects whole frame annotations, I can't create these any more. The tool is activated but I no longer have a way to "finish" it. When the setting is activated, the whole frame annotation should be immediately created when the button is clicked.

Apart from that, only minor comments are left.

resources/assets/js/videos/components/videoScreen.vue
```diff
@@ -77,16 +77,30 @@
...
+                            icon="fa-check"
+                            title="Disable the single-frame annotation option to create multi-frame annotations"
+                            :disabled="true"
+                            ></control-button> 
+                        <control-button
+                            v-else
                             icon="fa-check"
                             title="Finish the point annotation 𝗘𝗻𝘁𝗲𝗿"
-                            :disabled="cantFinishDrawAnnotation"
+                            :disabled="this.cantFinishDrawAnnotation"
```
mzur, line 89: Please don't use `this` anywhere in this template. It works without, too: `v-if="cantFinishDrawAnnotation"`.

mzur (commented):
One last general comment:

When the new settings option is active, the title text of the buttons still reads "Start a point annotation", "Start a rectangle annotation", etc. It should change to something similar to the image  annotation tool ("Set a point", "Draw a rectangle", ...).

resources/assets/js/videos/components/videoScreen.vue
```diff
@@ -229,15 +229,15 @@
                 </control-button>
                 <control-button
                     icon="icon-wholeframe"
-                    title="Start a whole frame annotation 𝗛"
+                    :title="(singleAnnotation ? 'Draw a whole frame' : 'Start a whole frame annotation') + ' 𝗛'"
```
mzur, line 232:
```suggestion
                    :title="(singleAnnotation ? 'Create a whole frame annotation' : 'Start a whole frame annotation') + ' 𝗛'"
```

PR #1024, opened 2024-12-20
Add temporary linter fix

mzur:
This will no longer be nexessary when biigle/largo was merged into the core.


mzur:
The fix did not work.

PR #1078, opened 2025-02-10
Add a file filter using a comma-separated list for volume cloning

### LINKED ISSUE #1052: Clone subset of files with filename list
As an alternative to the `*` wildcard, allow users to paste a comma-separated list of filenames in the clone volume form. This may be useful if users want to clone a random subset for QA/QC. Allow only one of the options, either the wildcard or the comma-separated list. Also validate that users can't submit more filenames than the DB parameter limit (or chunk a longer list in the backend).

lehecht:
When cloning volumes, the selected files can be filtered using a comma-separated list.

Closes #1052.

### DIFF
--- a/resources/views/volumes/clone.blade.php
+++ b/resources/views/volumes/clone.blade.php
@@ -74,7 +74,7 @@
                                        v-on:keydown.enter="loadFilesMatchingPattern">
                             @endif
                             <span class="help-block">
-                                Filter by using a list of comma-separated file names or pattern that matches specific file names. A pattern may contain the wildcard character * that matches any string of zero or more characters
+                                Filter by using a list of comma-separated file names or a pattern that matches specific file names. A pattern may contain the wildcard character * that matches any string of zero or more characters
                             </span>
                             <div class="volume-files-panel">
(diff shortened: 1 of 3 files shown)


mzur (changes requested):
Please also update the help text of the input field for the filenames.

resources/views/volumes/clone.blade.php
```diff
@@ -74,7 +74,7 @@
                                        v-on:keydown.enter="loadFilesMatchingPattern">
                             @endif
                             <span class="help-block">
-                                Filter by using a pattern that matches specific file names. A pattern may contain the wildcard character * that matches any string of zero or more characters
+                                Filter by using a list of comma-separated file names or pattern that matches specific file names. A pattern may contain the wildcard character * that matches any string of zero or more characters
```
mzur, line 77:
```suggestion
                                Filter by using a list of comma-separated file names or a pattern that matches specific file names. A pattern may contain the wildcard character * that matches any string of zero or more characters
```

mzur (approved):
Thanks!

PR #1081, opened 2025-02-12
Pie chart HSL value format

### LINKED ISSUE #1006: Total Contribution Pie Chart
When hovering over or moving the cursor on the pie chart, it flickers or disappears.

https://github.com/user-attachments/assets/01ffa318-2866-42dc-a16c-5690f7c86b41

gkourie:
Using comma-separated HSL values to generate unique colors for IDs instead of space-separated ones. Although both formats should be supported ([hsl](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/hsl)), the space-separated format causes flickering when hovering over the pie chart. Fixes #1006.


mzur (approved):
Thanks!

PR #1089, opened 2025-02-24
Fix contribution chart user colors

### LINKED ISSUE #1051: Fix contribution chart user colors
In the "total contribution" chart the user colors are correct in the pie chart but not in the line chart:

Image

Fix the colors to match the pie chart.

Maybe related to https://github.com/biigle/core/issues/497

aneust:
Fixes #1051 by adjusting end index calculation

- Ensure 'end' correctly reference the last element by accounting for 0-based indexing and possible length differences between 'xAxis-array' (sourcedata[0]) and 'annotation-array-user1' (sourcedata[1]).


lehecht (commented):
Works perfect!

resources/assets/js/projects/components/charts/timelinePlot.vue
```diff
@@ -181,7 +181,7 @@ export default {
         createTimelineSeries() {
             // create a series Array with entries for each user, used for the timeline-plot
             let series = [];
-            let end = this.sourcedata[0].length;
```
mzur, line 184: The `-1` does seem to fix the issue. But what do you mean with possible length differences between index 0 and 1? The arrays should all have the same length.

Did you use an AI coding assistant for this? I don't mind as long as it makes you more effective and produces working results 😉 The PR description got me curious.

aneust: I printed out how the sourcedata looks like, and in cases when the coloring does not work the first array (the one for the x-Axis) had the same length as the array that contains the information to the user-annotations. And in cases when it already worked the length were different, the annotation-user was longer by one element. To be honest, I did not get more in to it.

That is how the source output looks like if the coloring is working on master. 
StillBugBetterData

And this is how the sourcedata looks like, when the coloring don't work.
Screenshot from 2025-02-26 16-55-23


I did not know how to describe it shorty, so asked AI. The description was one of the outcomes.

aneust: ... now I see that there is the '2025-01' entry doubled in xAxis-array in the not working output

aneust:
Now it is possible to use both calculations for 'end'.

mzur (commented):
I see! So part of this bug was introduced in https://github.com/biigle/core/pull/910.

The code to "fill in missing dates" can probably be implemented more elegantly. It's purpose is to add dates where no annotations were done so the x axis has even steps. A more elegant solution could take the first and the last date from the (sorted) data and then just generate all the dates for the x axis instead of filling the gaps in the existing data. What do you think about this?

Then the `sourcedata` arrays should all have the same length and the `end` calculation can use index 0 again.

mzur (approved):
Perfect :ok_hand: Much more elegant than before.

PR #1091, opened 2025-02-26
Prevent thumbnail overflow in file info

### LINKED ISSUE #1016: Fix thumbnail display in image/video info view
The larger image/video thumbnails can exceed the element size in the image/video info view:

image

Make sure that the thumbnail is contained in the element.

lehecht:
Closes #1016.

### DIFF
--- a/resources/assets/sass/volumes/main.scss
+++ b/resources/assets/sass/volumes/main.scss
@@ -26,6 +26,12 @@ $volume-image-padding: 0.5em;
     background-color: $gray-lighter;
 }
 
+.prevent-overflow {
+    object-fit: contain;
+    width:100%;
+    height:100%;
+}
+
 .volume-storage-disk-btn {
     overflow-x: hidden;
     text-overflow: ellipsis;
--- a/resources/views/volumes/images/index.blade.php
+++ b/resources/views/volumes/images/index.blade.php
@@ -15,7 +15,7 @@
     <div class="row">
         <div class="col-sm-6 col-lg-4">
             <div class="panel panel-default panel-image">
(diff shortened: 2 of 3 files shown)


lehecht:
I had to add the `width` and `height` properties because `object-fit: contain` wouldn’t work otherwise. However, this caused portrait thumbnails to enlarge, so I need to check whether a thumbnail is in portrait or landscape mode.

mzur:
I assume this is no longer a draft?

mzur (commented):
I think a single `max-width: 100%` would work here, too? Then there is no need for `object-fit`, `height` or the `@if`.

lehecht:
@mzur 
Indeed, `object-fit` is not needed for landscape thumbnails, but portrait thumbnails are still displayed enlarged when using `max-width: 100%`.

**Left**: before change
**Right**: after change

 |

mzur:
Maybe add a `max-height` with a fixed pixel value then? I'd just like to avoid the `@if` check in the template, as this is basically just a CSS issue.

lehecht:
Sorry, I overlooked that I was still using `width` instead of `max-width`, so my previous comment was wrong. Using only `width` would have worked for landscape thumbnails only, but using `object-fit` with `max-width: 100%` and `max-height: 100%` works for all thumbnails without Blade directives.

mzur (commented):
Is the `object-fit` required at all now? Also I think a `max-height: 100%` has no effect. Better would be a `max-height: 270px` (as this is the default thumbnail size).

lehecht:
> Is the `object-fit` required at all now? Also I think a `max-height: 100%` has no effect. Better would be a `max-height: 270px` (as this is the default thumbnail size).

I removed `object-fit` and `max-height` since they are unnecessary for preventing the thumbnail from overflowing the container.

PR #1109, opened 2025-03-18
Extend abundance report

### LINKED ISSUE #1041: Include empty images/labels in reports
Add a switch to the abundance report to include empty images (annotation count 0) and "empty labels" in the report. Basically this means that the report includes all images and all labels that are available for a volume. Maybe this could be the new default behavior and there is no need for a switch.

This was requested by Catherine Borremans.

lehecht:
- [x] Add tests for old option combinations

`Add all_labels option`
- [x] include all images for abundance report
  - [x] Add test
- [ ] use all labels for separated label trees
  - [ ] Add test
 - [ ] use all labels for separated users
    - [ ] Add test
- [ ] use all labels if label trees and users shouldn't be separated
   - [ ] Add test
- [ ] use all labels when labels should be aggregated
  - [ ]  Add test
- [ ] check option combinations with all_labels: 
  - [ ] `isRestrictedToAnnotationSession`
  - [ ] `isRestrictedToNewestLabel`
  - [ ] `isRestrictedToLabels`
- [ ] check project report generator classes
- [ ] check project report generator tests

**Volume reports view**
- [ ] add button for `all_labels` option
- [ ] check when new option should be disabled
- [ ] add rules for new option

`Project reports view`
- [ ] add button for `all_labels` option
- [ ] check when new option should be disabled
- [ ] add rules for new option

Closes #1041

### DIFF
--- a/app/Services/Reports/Volumes/ImageAnnotations/AbundanceReportGenerator.php
+++ b/app/Services/Reports/Volumes/ImageAnnotations/AbundanceReportGenerator.php
@@ -42,22 +42,35 @@ class AbundanceReportGenerator extends AnnotationReportGenerator
 
         if ($this->shouldSeparateLabelTrees() && $rows->isNotEmpty()) {
             $rows = $rows->groupBy('label_tree_id');
-            $trees = LabelTree::whereIn('id', $rows->keys())->pluck('name', 'id');
-
+            $treeIds = $rows->keys()->reject(fn($k) => !$k);
+            $trees = LabelTree::whereIn('id', $treeIds)->pluck('name', 'id');
             foreach ($trees as $id => $name) {
                 $rowGroup = $rows->get($id);
                 $labels = Label::whereIn('id', $rowGroup->pluck('label_id')->unique())->get();
-                $this->tmpFiles[] = $this->createCsv($rowGroup, $name, $labels);
+                $this->tmpFiles[] = $this->createCsv($rows->flatten(), $name, $labels);
             }
         } elseif ($this->shouldSeparateUsers() && $rows->isNotEmpty()) {
             $labels = Label::whereIn('id', $rows->pluck('label_id')->unique())->get();
+            $allFilenames = $rows->pluck('filename')->unique();
             $rows = $rows->groupBy('user_id');
-            $users = User::whereIn('id', $rows->keys())
+            $userIds = $rows->keys()->reject(fn($k) => !$k);
+            $users = User::whereIn('id', $userIds)
                 ->selectRaw("id, concat(firstname, ' ', lastname) as name")
                 ->pluck('name', 'id');
 
             foreach ($users as $id => $name) {
                 $rowGroup = $rows->get($id);
+                $userFilenames = $rowGroup->pluck('filename')->unique();
+                $missingFiles = $allFilenames->diff($userFilenames);
+                // Create empty entries to show all images
+                foreach ($missingFiles as $f) {
+                    $rowGroup->add((object) [
+                        'filename' => $f,
+                        'count' => 0,
+                        'label_id' => null,
+                        'user_id' => $id
+                    ]);
+                }
                 $this->tmpFiles[] = $this->createCsv($rowGroup, $name, $labels);
             }
         } else {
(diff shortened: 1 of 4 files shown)


app/Services/Reports/Volumes/ImageAnnotations/AbundanceReportGenerator.php
```diff
@@ -38,31 +38,66 @@ class AbundanceReportGenerator extends AnnotationReportGenerator
...
+                $labels = $this->getVolumeLabels();
+            }
 
             foreach ($users as $id => $name) {
+                // Get query for images that have no annotations
+                $emptyImagesQuery = $query->clone()->whereNull('label_id')->select('filename');
+                $usedImageFilenames = $rows->get($id)->pluck('filename')->unique();
+                $usersEmptyImagesQuery = $query->clone()->select('filename')->whereNotIn('filename', $usedImageFilenames);
+                // Add query to process unannotated images (by current user) in chunks later
+                $emptyImagesQuery->union($usersEmptyImagesQuery);
```
mzur, line 85: I don't get this part. So you want to add the images that weren't annotated by the current user. Here you first get the query for all empty images. Then you get the filenames of images annotated by the user. Then you get the query of all (annotated and empty) images that were not annotated by the user. Then you union the first with the last query. But shouldn't this be enough?

```php
$usedImageFilenames = $rows->get($id)->pluck('filename')->unique();
$emptyImagesQuery = $query->clone()->select('filename')->whereNotIn('filename', $usedImageFilenames);
```

Or even simpler:

```php
$usedImageFilenames = $rows->get($id)->pluck('filename')->unique();
$emptyImagesQuery = $this->source->images()->select('filename')->whereNotIn('filename', $usedImageFilenames);
```

It's also faster to use the image ID for the constraint because it is indexed but you'd have to add it to the original query as a select:

```php
$usedImageIds = $rows->get($id)->pluck('image_id')->unique();
$emptyImagesQuery = $this->source->images()->select('filename')->whereNotIn('id', $usedImageIds);
```

app/Services/Reports/Volumes/ImageAnnotations/AbundanceReportGenerator.php
```diff
@@ -38,31 +38,66 @@ class AbundanceReportGenerator extends AnnotationReportGenerator
...
+
+                $this->tmpFiles[] = $this->createCsv($rowGroup, $name, $labels, $emptyImagesQuery);
             }
         } else {
-            $labels = Label::whereIn('id', $rows->pluck('label_id')->unique())->get();
-            $this->tmpFiles[] = $this->createCsv($rows, $this->source->name, $labels);
+            // Get query for images that have no annotations
+            $emptyImagesQuery = $query->clone()->whereNull('label_id')->select('filename');
+            $allLabelIds = $rows->pluck('label_id')->reject(fn ($k) => !$k);
+            $labels = $this->shouldUseAllLabels() ? $this->getVolumeLabels() : Label::whereIn('id', $allLabelIds)->get();
```
mzur, line 99: Don't collect all label IDs if you may not need them:

```suggestion
            if ($this->shouldUseAllLabels()) {
                $labels = $this->getVolumeLabels();
            } else {
                $allLabelIds = $rows->pluck('label_id')->filter()->unique();
                $labels = Label::whereIn('id', $allLabelIds)->get();
            }
```

app/Services/Reports/Volumes/ImageAnnotations/AbundanceReportGenerator.php
```diff
@@ -133,6 +222,17 @@ protected function createCsv($rows, $title, $labels)
...
+        $labelsCount = $labels->count();
+        $emptyImagesQuery->orderBy('filename')->chunk(1000, function ($images) use ($labelsCount, $csv) {
+            foreach ($images as $image) {
+                $row = [$image->filename];
+                for ($i = 0; $i < $labelsCount; $i++) {
+                    $row[] = 0;
+                }
+                $csv->putCsv($row);
+            }
+        });
```
mzur, line 234: Using a [lazy stream](https://laravel.com/docs/11.x/queries#streaming-results-lazily) could make this slightly more readable:

```suggestion
        foreach ($emptyImagesQuery->orderBy('filename')->lazy() as $image) {
            $row = [$image->filename];
            for ($i = 0; $i < $labelsCount; $i++) {
                $row[] = 0;
            }
            $csv->putCsv($row);
        }
```

mzur (commented):
Please also update the manual article about the abundance report schema.

I didn't look at all the tests, to be honest :grin:

mzur (changes requested):
With the change to the `prohibited_if` validation rule the UI shows an error even if the respective fields are `false`. Please make sure everything works before requesting a review again.

PR #1139, opened 2025-04-28
Fix remaining Python packages for Numpy 2 compatibility

mzur:
Now packages are installed via the requirements.txt which was only used for automatic security checks before. The Apline system packages are no longer used because they are not compatible with Numpy 2. Numpy 2 is required by Torch 2.6.

See: https://github.com/biigle/core/pull/1138

### DIFF
--- a/.docker/requirements.txt
+++ b/.docker/requirements.txt
@@ -1,13 +1,11 @@
-# This file is just used to get security alerts from GitHub. Make sure the versions match
-# in worker.dockerfile.
 numpy==2.1.*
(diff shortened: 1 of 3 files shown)


mzur:
This was done to fix the incompatibility of matplotlib with Numpy 2, which made the basic report generator fail. However, now the report seems to be generated incorrectly (see below). It generates two pages in the PDF and the x axis labels are positioned wrong. Also the style of the bars seems to be wrong.

image

Compare this with the reference from the manual:

image

PR #1175, opened 2025-05-12
Arrow keys navigation for single video in annotation tool

### LINKED ISSUE #878: Video annotation tool arrow keys for single video
When pressing the left/right arrow keys in the video annotation tool, it starts to load even if only a single video is present in the volume. This shouldn't happen and worked correctly in the past. Find the commit that introduced the bug and fix it.

aneust:
Conditional Keyboard Shortcuts Based on showPrevNext

- Fixes a bug where pressing the left/right arrow keys in the video annotation tool triggers a reload, even if only a single video is available
- Updated adaptKeyboardShortcuts() to conditionally add/remove navigation-related keyboard shortcuts based on the showPrevNext flag

Resolve #878

### DIFF
--- a/resources/assets/js/videos/components/videoScreen.vue
+++ b/resources/assets/js/videos/components/videoScreen.vue
@@ -643,21 +643,33 @@ export default {
             this.resetInteractionMode();
         },
         adaptKeyboardShortcuts() {
-            if(this.enableJumpByFrame) {
+            if (this.enableJumpByFrame) {
                 Keyboard.off('ArrowRight', this.emitNext, 0, this.listenerSet);
                 Keyboard.off('ArrowLeft', this.emitPrevious, 0, this.listenerSet);
-                Keyboard.on('Shift+ArrowRight', this.emitNext, 0, this.listenerSet);
-                Keyboard.on('Shift+ArrowLeft', this.emitPrevious, 0, this.listenerSet);
                 Keyboard.on('ArrowRight', this.emitNextFrame, 0, this.listenerSet);
                 Keyboard.on('ArrowLeft', this.emitPreviousFrame, 0, this.listenerSet);
+                if (this.showPrevNext) {
+                    Keyboard.on('Shift+ArrowRight', this.emitNext, 0, this.listenerSet);
+                    Keyboard.on('Shift+ArrowLeft', this.emitPrevious, 0, this.listenerSet);
+                }
+                else {
+                    Keyboard.off('Shift+ArrowRight', this.emitNext, 0, this.listenerSet);
+                    Keyboard.off('Shift+ArrowLeft', this.emitPrevious, 0, this.listenerSet);
+                }
             }
             else {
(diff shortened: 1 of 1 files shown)


aneust:
Possible source of error is #823

resources/assets/js/videos/components/videoScreen.vue
```diff
@@ -643,21 +643,33 @@ export default {
...
-                Keyboard.on('ArrowRight', this.emitNext, 0, this.listenerSet);
-                Keyboard.on('ArrowLeft', this.emitPrevious, 0, this.listenerSet);
+                if (this.showPrevNext) {
+                    Keyboard.on('ArrowRight', this.emitNext, 0, this.listenerSet);
+                    Keyboard.on('ArrowLeft', this.emitPrevious, 0, this.listenerSet);
+                }
+                else {
+                    Keyboard.off('ArrowRight', this.emitNext, 0, this.listenerSet);
+                    Keyboard.off('ArrowLeft', this.emitPrevious, 0, this.listenerSet);
+                }
```
lehecht, line 672: You can make this a bit more readable when you use a method like `switchVideosWithKeys(left, right)` where you turn on the keyboard shortcuts. In the else branch you can use `Keyboard.off` without the if-statement, because it will do nothing if the keys don't exist. Additionally, you could also use the `switchVideosWithKeys` method in the created method instead of `adaptKeyboardShortcuts`.

So it looks like this:

```js
        adaptKeyboardShortcuts() {
            if (this.enableJumpByFrame) {
                Keyboard.off('ArrowRight', this.emitNext, 0, this.listenerSet);
                Keyboard.off('ArrowLeft', this.emitPrevious, 0, this.listenerSet);
                Keyboard.on('ArrowRight', this.emitNextFrame, 0, this.listenerSet);
                Keyboard.on('ArrowLeft', this.emitPreviousFrame, 0, this.listenerSet);
                this.switchVideosWithKeys('Shift+ArrowLeft', 'Shift+ArrowRight');
            }
            else {
                Keyboard.off('Shift+ArrowRight', this.emitNext, 0, this.listenerSet);
                Keyboard.off('Shift+ArrowLeft', this.emitPrevious, 0, this.listenerSet);
                Keyboard.off('ArrowRight', this.emitNextFrame, 0, this.listenerSet);
                Keyboard.off('ArrowLeft', this.emitPreviousFrame, 0, this.listenerSet);
                this.switchVideosWithKeys('ArrowLeft', 'ArrowRight');
            }
        }
```

lehecht (changes requested):
I'm not entirely sure whether the bug was actually introduced in #823. I checked out commit bbc30e06 (and several others prior to the first commit of #823), and the bug was still there.

Can you confirm whether you're seeing the same issue? And if yes, look for the real cause? You can use the git history of [videoScreens.vue](https://github.com/biigle/core/commits/master/resources/assets/js/videos/components/videoScreen.vue?after=92c07b9a51e1cec90e7e57a8598d2f949390d376+34) to see changes have been made over time.

PR #1179, opened 2025-05-20
Improve tile upload performance for S3 clients

### LINKED ISSUE #945: Improve TileSingle image performance
A tiled image can easily produce 100.000 files. It is extremely slow to upload them sequentially to an S3 backend here:

https://github.com/biigle/core/blob/e4756dffaef0c4db3d36a3074b43676a6c3b97d5/app/Jobs/TileSingleImage.php#L92-L107

To speed this up we could use asynchronous requests if an S3 backend is used. This circumvents Laravel's file system abstraction and uses the S3 SDK directly. Basically in the place shown above, detect if an S3 adapter is used. Then extract the S3 client from it and upload the files asynchronously as described here: https://stackoverflow.com/a/65365224/1796523

Make the number of parallel connections configurable. The default number should be 10.

Code from StackOverflow:
```php
$files = glob('/path/to/your/files/*'); // This will return an array of all files in your folder

try {
    // Init of S3 client
    $s3Client = new \Aws\S3\S3Client([
        'version'       => 'latest',
        'region'        => '', // Desired AWS region
        'credentials'   => [
            'key'       => '', // Your AWS key
            'secret'    => '', // Your AWS key secret
        ],
    ]);
    
    // Logic about your requests and how to execute them
    $uploads = function($files) use ($s3Client) {
        foreach ($files as $file) {
            yield $s3Client->putObjectAsync([
                'Bucket'        => '', // Name of your bucket
                'Key'           => basename($file),
                'SourceFile'    => $file,
            ]);
        }
    };
    
    // Execute your requests with Guzzle because $s3Client->putObjectAsync() returns \GuzzleHttp\Promise\Promise
    \GuzzleHttp\Promise\Each::ofLimit(
        $uploads($files),
        3, // How much concurrent request to start
        function($response, $index) { // Callback on success
            var_dump('Success: ' . $index); 
        },
        function($reason, $index) { // Callback on failure
            var_dump('Error: ' . $index);
        }
    )->wait();
} catch (\Aws\S3\Exception\S3Exception $e) {
    var_dump($e->getMessage());
}
```

lehecht:
Improve the tile upload performance for S3 clients by sending requests concurrently.

Closes #945.

### DIFF
--- a/app/Jobs/TileSingleImage.php
+++ b/app/Jobs/TileSingleImage.php
@@ -6,6 +6,7 @@ use File;
 use Exception;
 use FileCache;
 use Biigle\Image;
+use ArrayIterator;
 use FilesystemIterator;
 use GuzzleHttp\Promise\Each;
 use RecursiveIteratorIterator;
@@ -14,9 +15,7 @@ use Illuminate\Support\Facades\Log;
 use Jcupitt\Vips\Image as VipsImage;
 use Illuminate\Queue\SerializesModels;
 use Illuminate\Support\Facades\Storage;
-use GuzzleHttp\Promise\PromiseInterface;
 use Illuminate\Queue\InteractsWithQueue;
-use Illuminate\Filesystem\AwsS3V3Adapter;
 use Illuminate\Contracts\Queue\ShouldQueue;
 
 class TileSingleImage extends Job implements ShouldQueue
@@ -67,8 +66,8 @@ class TileSingleImage extends Job implements ShouldQueue
     {
         try {
             FileCache::getOnce($this->image, [$this, 'generateTiles']);
-            $adapter = Storage::disk(config('image.tiles.disk'))->getAdapter();
-            if ($adapter instanceof AwsS3V3Adapter) {
+            if (config('filesystems.disks.tiles.driver') === 's3') {
+                Log::info("upload to s3");
                 $this->uploadToS3Storage();
             } else {
                 $this->uploadToStorage();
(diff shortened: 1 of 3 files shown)


lehecht:
I think, I found a solution to handle files that were already uploaded.
Maybe you can give your opinion on my idea [here](https://github.com/biigle/core/blob/0bdb61fa93a54fc807dc1e6b57641b096613f81f/app/Jobs/TileSingleImage.php#L142-L147) and [here](https://github.com/biigle/core/blob/0bdb61fa93a54fc807dc1e6b57641b096613f81f/app/Jobs/TileSingleImage.php#L197-L200)?

mzur:
In the regular case if there was an error uploading the files, [any uploaded files were deleted](https://github.com/biigle/core/blob/0bdb61fa93a54fc807dc1e6b57641b096613f81f/app/Jobs/TileSingleImage.php#L114) so the failed job does not leave any trace in the storage disk. The S3 upload should behave the same way so there should not be any previously uploaded files if a job is retried. If there are some anyway, this is an error condition and the job can fail.

app/Jobs/TileSingleImage.php
```diff
@@ -106,6 +116,107 @@ public function uploadToStorage()
...
+                    'Bucket' => $bucket,
+                    'Key' => "{$root}{$prefix}{$path}",
+                    'SourceFile' => $file,
+                    // Return with error if file already exist
+                    '@http' => [
+                        'headers' => [
+                            'If-None-Match' => '*'
+                        ]
+                    ]
+                ]);
```
mzur, line 150: I don't get why the path/key is built differently to `uploadToStorage()` here. Doesn't it work the same way?

```php
$prefixLength = strlen($this->tempPath) + 1;
$iterator = $this->getIterator($this->tempPath);
$fragment = fragment_uuid_path($this->image->uuid);

// ...

foreach ($iterator as $pathname => $fileInfo) {
    yield $client->putObjectAsync([
        'Bucket' => $bucket,
        'Key' => "{$prefix}{$fragment}/".substr($pathname, $prefixLength),
        'SourceFile' => $fileInfo,
        // Return with error if file already exist
        '@http' => [
            'headers' => [
                'If-None-Match' => '*'
            ]
        ]
    ]);
}
```

app/Jobs/TileSingleImage.php
```diff
@@ -106,6 +116,107 @@ public function uploadToStorage()
...
+        return $disk->getConfig()['bucket'];
+    }
+
+    /**
+     * Upload files to the S3 storage.
+     *
+     * @param \Iterator $files The files to upload.
+     * @param callable $onFullfill Callback for successful uploads.
+     * @param callable $onReject Callback for failed uploads.
+     * @throws Exception If retry limit is reached
```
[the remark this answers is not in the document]
mzur: `@throws Exception If retry limit is reached`

app/Jobs/TileSingleImage.php
mzur: I see that your approach has the same result but I still don't see an advantage over the implementation in `uploadToStorage()` (rather the opposite as your `$prefix` is determined anew in each loop whereas `$fragment` is set only once). But I leave this up to you as it's only a minor detail.

app/Jobs/TileSingleImage.php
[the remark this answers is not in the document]
lehecht: Can you tell me what is wrong?
I looked into the [Guzzle code](https://github.com/caseyamcl/guzzle_retry_middleware/blob/362bdd5e6d19e54e1f4c8c27757e7598201efa3c/src/GuzzleRetryMiddleware.php#L205-L238), and the middleware first retries each request and eventually uses the reject callback where I throw an exception.

app/Jobs/TileSingleImage.php
```diff
@@ -125,21 +126,22 @@ public function uploadToStorage()
...
 
             foreach ($files as $file) {
-                $path = substr($file, $tmpLength);
-                $prefix = $path[0] . $path[1] . '/' . $path[2] . $path[3] . "/";
+                $path = substr($file, $dirLength);
                 // @phpstan-ignore-next-line
                 yield $client->putObjectAsync([
                     'Bucket' => $bucket,
-                    'Key' => "{$root}{$prefix}{$path}",
+                    'Key' => "{$prefix}{$path}",
```
mzur, line 144: So the former `$root` is now called `$prefix`. But don't we need the former `$prefix` here still? The `$path` is the UUID of the image but we still need the "fagment" parts before it.

app/Jobs/TileSingleImage.php
```diff
@@ -50,7 +54,8 @@ class TileSingleImage extends Job implements ShouldQueue
     public function __construct(Image $image)
     {
         $this->image = $image;
-        $this->tempPath = config('image.tiles.tmp_dir')."/{$image->uuid}";
+        $fragment = fragment_uuid_path($image->uuid);
+        $this->tempPath = config('image.tiles.tmp_dir') . "/{$fragment}";
```
mzur, line 58: Ah now I see this. This will clutter the temp directory, as a new tempPath is generated like this: `/tmp/ba/da/bada55-uuid/` but `deleteDirectory` below will only delete the last `bada55-uuid`. At some point you end up with possibly 256² empty directories.

The "fragment" stuff is only needed at the target storage location. For temporary storage here the UUID is totally fine.

app/Jobs/TileSingleImage.php
```diff
@@ -79,6 +90,7 @@ public function handle()
      */
     public function generateTiles(Image $image, $path)
     {
+        File::ensureDirectoryExists(path: $this->tempPath, recursive: true);
```
mzur, line 93: Maybe without nested temp directories this is also not needed any more?

mzur (approved):
Looks good. I've remove the type annotation that caused the linter error, so no `@phpstan-ignore` is necessary any more.

PR #1190, opened 2025-06-10
Set scale limit for NetMap

### LINKED ISSUE #1128: Limit NetMap zoom in/out in charts
When zooming all the way out on the `NetMap`, it becomes too small and difficult to zoom back in, as we need to hover directly over the `NetMap` to do so. Maybe we can:

- Limit the minimum and maximum zoom levels, or
- Add a "Reset Zoom" button to return the `NetMap` to its default size.

https://github.com/user-attachments/assets/eb0cbe47-ce3a-416c-a842-6a113ffe0717

gkourie:
closes #1128


mzur (approved):
Thanks!

Issue #957, opened 2024-10-22
Favourite label presets


mzur:
Users asked to be able to configure more favorite labels. It can work like this: There are up to 10 sets of favorite labels. By default, set 0 is active and new favorite labels are added to this set (current behavior). But now users can switch to a different set by pressing <kbd>Ctrl</kbd>+<kbd>0</kbd>-<kbd>9</kbd>. Switching to a different empty set effectively clears the favorite label tree and allows the user to add another 10 favorite labels.

In the UI this should be shown with a dropdown button next to the favorite label tree title. In the button, the current number of the selected label set is shown. Users can click this and get a dropdown to select another label set. Favorite label sets should be remembered in local storage similar to favorite labels now.

This was a user request.

I close the idea with [virtual label trees](https://github.com/biigle/core/issues/554) in favor of this one.

PR #1192, opened 2025-06-12
Max favorites at 20

marinmarcillat:
Augment the number of favorites to 20 to ease annotation of pictures with high diversity

### DIFF
--- a/resources/assets/js/label-trees/constants.js
+++ b/resources/assets/js/label-trees/constants.js
@@ -1,1 +1,1 @@
-export const MAX_FAVOURITES = 10;
+export const MAX_FAVOURITES = 20;


mzur:
Hi, thanks for the suggestion! We had some requests for more favorite labels before but we didn't want to just increase the limit because each favorite label should be accessible via a shortcut key. The current labels can be activated with <kbd>0</kbd>-<kbd>9</kbd>. To allow more favorite labels we came up with the idea for [label presets](https://github.com/biigle/core/issues/957), which are basically "pages" of favorite labels. Do you think this could also work for your use case? If yes, please let me know if you would consider to implement the idea :wink:

PR #1194, opened 2025-06-13
Improve selected annotation style

### LINKED ISSUE #1112: Improve selected annotation style
When selecting an annotation in the image annotation tool, the annotation is highlighted in red. If an annotation is already red, it becomes hard to distinguish between the selected and unselected annotations.

Unselected annotations:

Image

After selecting the "label 2" annotation:

Image

Proposed solution:
Maybe adding an animation, such as dashed lines around the selected annotation, to make it more distinguishable from the unselected ones.

gkourie:
closes #1112

### DIFF
--- a/resources/assets/js/annotations/annotatorContainer.vue
+++ b/resources/assets/js/annotations/annotatorContainer.vue
@@ -437,7 +437,7 @@ export default {
                 };
 
                 let promise = AnnotationsStore.attachLabel(annotation, annotationLabel);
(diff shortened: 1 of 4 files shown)


mzur (changes requested):
The selected circle stroke has to be updated as well (for point annotations). But you have to make sure the inner area of the circle (with the highlight color fill) does not vanish if you just make the stroke thicker :wink:

PR #1195, opened 2025-06-13
Add visual feedback when attaching labels with volare

### LINKED ISSUE #253: Visual feedback for attached label with Volare
The user has noo feedback if the selected label was attached (with enter) to the selected annotation in Volare. Think of some visual feedback that can be used here (e.g. flashing of the annotation) and implement it.

gkourie:
closes #253

### DIFF
--- a/resources/assets/js/annotations/annotatorContainer.vue
+++ b/resources/assets/js/annotations/annotatorContainer.vue
@@ -437,17 +437,11 @@ export default {
                 };
 
                 if (this.isVolareAnnotationMode) {
-                    this.$refs.canvas.setEditingStyle(annotation);
+                    this.$refs.canvas.blinkAnnotation(annotation);
                 }
 
                 let promise = AnnotationsStore.attachLabel(annotation, annotationLabel);
-                promise
-                    .catch(handleErrorResponse)
-                    .finally(() => {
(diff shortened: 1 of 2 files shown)


resources/assets/js/annotations/annotatorContainer.vue
```diff
@@ -435,8 +435,19 @@ export default {
                     // TODO: confidence control
                     confidence: 1,
                 };
+
+                if (this.isVolareAnnotationMode) {
+                    this.$refs.canvas.setEditingStyle(annotation);
+                }
+
```
mzur, line 442: I'd prefer this to have a fixed "blink rate", i.e. when the label is attached, the annotation blinks with the same speed independent from the API request. You can do this here with something like:
```suggestion

                if (this.isVolareAnnotationMode) {
                    this.$refs.canvas.blinkAnnotation(annotation);
                }
```
Then there is no need for the code change below. If there is an error with the API request, it will be handled by `handleErrorResponse`.

For the blinking animation maybe apply the editing style for 250 ms?

mzur (changes requested):

mzur (changes requested):
The result is great :+1:  Code needs tiny changes.

mzur (approved):

PR #1196, opened 2025-06-15
Safeguards when deleting a volume or project

### LINKED ISSUE #1074: Implement safeguards when deleting a volume or project
When users delete a volume or project, lots of annotations can be deleted accidentally. While BIIGLE shows a confirmation dialog if the volume/project contains annotations, many users just dismiss these. Update the confirmation dialog as follows:

When the dialog is shown (i.e. when there are annotations to be deleted), show an *input* dialog and ask the user to type the volume/project name to confirm the deletion. Offer the name in the dialog text so users could copy/paste it for convenience.

This should make users pause and think if the project/volume is really the one that should be deleted.

aneust:
Improve deletion confirmation dialog for volumes and projects that contain annotations. Added confirmation dialog includes an input field that requires the user to manually enter the name of the volume or project to confirm the deletion.

- Updated confirmation dialog for deletion of annotated volumes/projects.
- Input prompt added that requires typing the volume/project name.
- Dialog only shown when annotations are present and would be deleted.

closes #1074

### DIFF
--- a/resources/assets/js/projects/title.vue
+++ b/resources/assets/js/projects/title.vue
@@ -61,15 +61,12 @@ export default{
             setTimeout(() => location.href = this.redirectUrl, 2000);
         },
         deleteProject() {
-            let inputproject = prompt(`Do you want to delete ${this.project.name}?`);
+            let inputproject = prompt(`Are you scure you want to delete "${this.project.name}"? Please type the project name to confirm.`);
             if (inputproject == this.project.name) {
-                let confirmed = confirm(`Do you really want to delete the project ${this.project.name}?`);
-                if (confirmed) {
-                    this.startLoading();
-                    ProjectsApi.delete({ id: this.project.id })
-                        .then(this.projectDeleted, this.maybeForceDeleteProject)
-                        .finally(this.finishLoading);
-                }
+                this.startLoading();
+                ProjectsApi.delete({ id: this.project.id })
+                    .then(this.projectDeleted, this.maybeForceDeleteProject)
+                    .finally(this.finishLoading);
             }
         },
         maybeForceDeleteProject(response) {
--- a/resources/assets/js/projects/title.vue
+++ b/resources/assets/js/projects/title.vue
@@ -61,8 +61,8 @@ export default{
             setTimeout(() => location.href = this.redirectUrl, 2000);
         },
         deleteProject() {
-            let inputproject = prompt(`Are you scure you want to delete "${this.project.name}"? Please type the project name to confirm.`);
-            if (inputproject == this.project.name) {
+            let inputproject = prompt(`Are you sure you want to delete "${this.project.name}"? Please type the project name to confirm.`);
+            if (inputproject === this.project.name) {
                 this.startLoading();
                 ProjectsApi.delete({ id: this.project.id })
                     .then(this.projectDeleted, this.maybeForceDeleteProject)
--- a/resources/assets/js/projects/title.vue
+++ b/resources/assets/js/projects/title.vue
@@ -61,7 +61,7 @@ export default{
             setTimeout(() => location.href = this.redirectUrl, 2000);
         },
         deleteProject() {
-            let inputproject = prompt(`Do you realy want to delete ${this.project.name} ?`);
+            let inputproject = prompt(`Do you want to delete ${this.project.name}?`);
             if (inputproject == this.project.name) {
                 let confirmed = confirm(`Do you really want to delete the project ${this.project.name}?`);
                 if (confirmed) {
--- a/resources/assets/js/projects/title.vue
+++ b/resources/assets/js/projects/title.vue
@@ -24,6 +24,7 @@ export default{
             description: null,
             userId: null,
             redirectUrl: null,
+            volumes: [],
         };
     },
     computed: {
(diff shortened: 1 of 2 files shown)


resources/assets/js/projects/title.vue
```diff
@@ -61,23 +62,39 @@ export default{
...
-                    ProjectsApi.delete({id: this.project.id}, {force: true})
-                        .then(this.projectDeleted, handleErrorResponse)
-                        .finally(this.finishLoading);
+                let volNames = [];
+                for (let i = 0; i < this.volumes.length; i++) {
+                    volNames[i] = this.volumes[i].name;
+                }
+                let inputVolume = prompt(`Do you realy want to delete ${volNames.toString()}?`);
+                if (inputVolume.replace(/\s/g, '') == volNames.toString().replace(/\s/g, '')) {
+                    let confirmed = confirm('Deleting this project will delete one or more volumes with all annotations! Do you want to continue?');
```
aneust, line 91: @mzur  Is it needed to ask about each or all volumes that are in a project that should be deleted?

mzur: If a volume should be deleted, users should enter the volume name. If a project should be deleted, users should enter the project name.

lehecht:
I noticed that when you're trying to delete a volume, there's no description what to do to proceed. Maybe add "Please enter the volume name to proceed" or something similar so the user knows what's expected.

Also, is there a reason why we need two more prompts after entering the volume name? Wouldn't it be enough to only have two prompts in total here?

aneust:
@lehecht I thought about having a better 'stop' to consider, but maybe it's enough just to read and copy-paste the name of the volume/project.

resources/assets/js/projects/title.vue
```diff
@@ -61,11 +61,10 @@ export default{
             setTimeout(() => location.href = this.redirectUrl, 2000);
         },
         deleteProject() {
-            let confirmed = confirm(`Do you really want to delete the project ${this.project.name}?`);
-
-            if (confirmed) {
+            let inputproject = prompt(`Are you scure you want to delete "${this.project.name}"? Please type the project name to confirm.`);
```
lehecht, line 64:
```suggestion
            let inputproject = prompt(`Are you sure you want to delete "${this.project.name}"? Please type the project name to confirm.`);
```
Please fix the typo.

resources/assets/js/projects/volumesContainer.vue
```diff
@@ -139,10 +139,12 @@ export default {
...
         removeVolume(id) {
             this.startLoading();
-            ProjectsApi.detachVolume({id: this.project.id, volume_id: id})
+            ProjectsApi.detachVolume({ id: this.project.id, volume_id: id })
                 .then(() => this.volumeRemoved(id), (response) => {
+                    let volumeToDelete = this.volumes.find(volume => volume.id === id).name;
                     if (response.status === 400) {
-                        if (confirm('The volume you are about to remove belongs only to this project and will be deleted. Are you sure you want to delete this volume?')) {
+                        let inputVolume = prompt(`The volume you are about to remove belongs only to this project and will be deleted. Are you sure you want to delete "${volumeToDelete}"? Please type the volume name to confirm.`);
+                        if (inputVolume == volumeToDelete) {
```
lehecht, line 147:
```suggestion
                        if (inputVolume === volumeToDelete) {
```
See comment above,

resources/assets/js/projects/title.vue
```diff
@@ -61,11 +61,10 @@ export default{
...
+            if (inputproject == this.project.name) {
```
lehecht, line 65:
```suggestion
            if (inputproject === this.project.name) {
```
The identity operator `===` also checks the arguments types, so it's a bit safer to use.

resources/assets/js/projects/volumesContainer.vue
```diff
@@ -139,10 +139,12 @@ export default {
     methods: {
         removeVolume(id) {
             this.startLoading();
-            ProjectsApi.detachVolume({id: this.project.id, volume_id: id})
+            ProjectsApi.detachVolume({ id: this.project.id, volume_id: id })
                 .then(() => this.volumeRemoved(id), (response) => {
+                    let volumeToDelete = this.volumes.find(volume => volume.id === id).name;
                     if (response.status === 400) {
-                        if (confirm('The volume you are about to remove belongs only to this project and will be deleted. Are you sure you want to delete this volume?')) {
+                        let inputVolume = prompt(`The volume you are about to remove belongs only to this project and will be deleted. Are you sure you want to delete "${volumeToDelete}"? Please type the volume name to confirm.`);
```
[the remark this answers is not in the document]
aneust: It is the same problem. I tried to put the volume/project name into the input field. But then the user can just confirm without really reading the prompt.

resources/assets/js/projects/volumesContainer.vue
```diff
@@ -139,10 +139,12 @@ export default {
     methods: {
         removeVolume(id) {
             this.startLoading();
-            ProjectsApi.detachVolume({id: this.project.id, volume_id: id})
+            ProjectsApi.detachVolume({ id: this.project.id, volume_id: id })
                 .then(() => this.volumeRemoved(id), (response) => {
+                    let volumeToDelete = this.volumes.find(volume => volume.id === id).name;
```
lehecht, line 144: Please move this line below the if-statement.

lehecht, line 144: Please move this line below the if-statement.

lehecht (changes requested):
>Dialog only shown when annotations are present and would be deleted.

Currently volumes and projects without annotations show the input dialogue, too. Does this happen on purpose?

PR #1197, opened 2025-06-16
Fix the missing timeline for remote videos in firefox

lehecht:
Remote fragmented .mp4 files can trigger multiple durationchange events, updating the duration incrementally. In firefox the  loadedmetadata event is fired after the first duration update. To avoid an invalid or incorrect duration, use the server-provided video duration when the loadedmetadata event fires.

### DIFF
--- a/resources/assets/js/videos/components/videoTimeline.vue
+++ b/resources/assets/js/videos/components/videoTimeline.vue
@@ -175,8 +175,9 @@ export default {
             this.currentTime = this.video.currentTime;
         },
         setDuration() {
-            // Use the video duration the from server response, since Firefox's loadedmetadata event
-            // can be fired before the final durationchange event when handeling non-standard fragmented mp4.
+            // Use the video duration the from server response, since Firefox's loadedmetadata event is not reliable.
+            // The loadedmetadata event can be fired before the final durationchange event leading to an invalid or incorrect duration
+            // when requesting remote fragmented .mp4 files.
             this.duration = this.videoDuration;
         },
         emitSeek(time) {
--- a/resources/assets/js/videos/components/videoTimeline.vue
+++ b/resources/assets/js/videos/components/videoTimeline.vue
@@ -59,6 +59,10 @@ export default {
             type: HTMLVideoElement,
             required: true,
         },
+        videoDuration: {
+            type: Number,
+            required: true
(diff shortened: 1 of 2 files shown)


resources/assets/js/videos/components/thumbnailPreview.vue
```diff
@@ -266,6 +266,14 @@ export default {
...
     },
     watch: {
+        duration() {
+            this.setSpritesFolderpath();
+            this.updateThumbnailInterval();
+
+            this.spriteIdx = Math.floor(this.hoverTime / (this.thumbnailInterval * this.thumbnailsPerSprite));
+            this.updateSprite();
+            this.viewHoverTimeBar();
+        },
```
mzur, line 276: Why is this needed? With the new implementation the duration should never change, except at first from 0 to >0. The thumbnail preview is only shown once the duration is >0.

lehecht: Can you explain why the duration should never change? Every video has a different duration. Or did you mean something different?

lehecht: I added the watcher because otherwise the thumbnails would no longer be updated. Previously, when the video changed, the duration was reset to 0, which caused the thumbnail component to be destroyed and recreated. This ensured that the data was always freshly loaded. Now, the duration is set to 0 only once at the beginning, so the thumbnails need to be updated when the video changes.

mzur: Makes sense, thanks. I didn't consider multiple videos in the same volume :facepalm:

resources/assets/js/videos/components/thumbnailPreview.vue
```diff
@@ -266,6 +266,14 @@ export default {
         }
     },
     watch: {
+        duration() {
+            this.setSpritesFolderpath();
+            this.updateThumbnailInterval();
+
+            this.spriteIdx = Math.floor(this.hoverTime / (this.thumbnailInterval * this.thumbnailsPerSprite));
```
mzur, line 273: The spriteIdx is set directly in `updateSprite()`, so there is no need to set it here. It's the same in `mounted()` where the line can be removed.

PR #1207, opened 2025-07-02
Labelbot refactor

mzur:
This moves most logic related to the LabelBOT popup into the component itself. It also introduces the following changes:

- More natural feel of the popup position during zoom.

- The first time <kbd>Esc</kbd> is pressed, the timeout stops. The second time, the popup is closed.


gkourie:
Thanks! The UI feels smoother.

> Please test the new web worker with WebGPU if you can

I tested it on Chrome and it works :ok_hand: 

The problem of not hitting the rate limit when creating lots of annotation (points) persists:
[Screencast from 2025-07-04 12-12-13.webm](https://github.com/user-attachments/assets/9f18c5d0-768d-49eb-a2ac-4bd35095e718)

Additionally the editing features (light blue dots) vanish before the popup is opened.


This happens when no other annotations exist before drawing the first one. Here I drew an annotation before drawing lots of points:
[Screencast from 2025-07-04 12-16-52.webm](https://github.com/user-attachments/assets/dd6120b4-7e6c-402b-977f-95ce6cd2a1da)

gkourie (approved):

PR #1216, opened 2025-07-10
Support small TIFFs in image annotation tool

### LINKED ISSUE #1151: Support small TIFFs in image annotation tool
In https://github.com/biigle/user-storage/pull/54 we introduce the utif2 package to check if an uploaded TIFF is too small. Small TIFFs are not supported because they are not preprocessed by BIIGLE and browsers cannot display them natively. However, with utif2 it is possible to decode TIFF images in JS. If this is fast enough (which it may be for small images), we can add full TIFF support to BIIGLE! This is just a frontend issue. The backend can work with any image format.

Here is the example code for decoding:

```js
function imgLoaded(e) {
  var ifds = UTIF.decode(e.target.response);
  UTIF.decodeImage(e.target.response, ifds[0])
  var rgba  = UTIF.toRGBA8(ifds[0]);  // Uint8Array with RGBA pixels
  console.log(ifds[0].width, ifds[0].height, ifds[0]);
}

var xhr = new XMLHttpRequest();
xhr.open("GET", "my_image.tif");
xhr.responseType = "arraybuffer";
xhr.onload = imgLoaded;   xhr.send();
```

The Uint8Array can be used to draw the image to the [canvas of the image wrapper](https://github.com/biigle/core/blob/6517625286014592d0dca3bede2cec5a5ce6fc64/resources/assets/js/annotations/stores/images.js#L204). This only works if CORS is configured correctly, so if users load small TIFFs from an incorrectly configured remote location, an error has to be shown.

aneust:
Add support for displaying small TIFF images (<1GB) in the browser using the UTIF.js library. TIFFs are decoded with utif2 and drawn onto the image wrapper’s canvas. 

- [x] Implement TIFF display in the image annotation tool
- [x] ~~Implement a basic TIFF viewer in the biigle/user-storage module so admins can review uploaded TIFFs~~
- [x] Update biigle/user-storage module to allow all TIFFs https://github.com/biigle/user-storage/issues/65
- [x] Update helpdesk knowledge base and inform reviewers

Closes #1151

### DIFF
--- a/resources/assets/js/annotations/stores/images.js
+++ b/resources/assets/js/annotations/stores/images.js
@@ -210,7 +210,7 @@ class Images {
                 resolve(imageWrapper);
             };
 
-            img.onerror = function (e) {
+            img.onerror = function () {
                 // If the image is loaded without CORS and decoding failed, it may be
                 // a TIFF.
                 if (imageWrapper.crossOrigin) {


resources/assets/js/annotations/annotatorContainer.vue
```diff
@@ -156,8 +156,12 @@ export default {
             return imagesIds;
         },
         hasCrossOriginError() {
-            return !this.loading && this.crossOriginError;
+            return !this.loading && this.crossOriginError && !this.crossOriginTiffError;
         },
+        hasCrossOriginErrorTiff() {
+            return this.crossOriginError && this.crossOriginTiffError;
```
lehecht, line 162:
```suggestion
            return this.loading && this.crossOriginTiffError;
```


I think you can replace `crossOriginError` by `loading`, because 
1. `crossOriginTiffError` is independent of `crossOriginError` and 
2. the error message should only be shown when it finished loading

resources/assets/js/annotations/stores/images.js
```diff
@@ -310,7 +338,6 @@ class Images {
     }
 
     drawImage(image) {
-
```
lehecht, line 313: Please add the new line again.

resources/views/annotations/show.blade.php
```diff
@@ -64,6 +64,13 @@
                 <button type="button" class="btn btn-default" v-on:click="dismissCrossOriginError">Continue at own risk</button>
             </div>
         </message-curtain>
+        <message-curtain v-if="hasCrossOriginErrorTiff" v-cloak>
+            <div class="message-curtain--text text-danger">
+                <div class="lead">
+                    Unable to display the image.<br>Please configure <a href="/manual/tutorials/volumes/remote-volumes#cors">cross origin resource sharing</a> on your remote image location.
```
lehecht, line 70:
```suggestion
                    Image can't be displayed.<br>Please configure <a href="/manual/tutorials/volumes/remote-volumes#cors">cross origin resource sharing</a> on your remote image location.
```

Issue #1166, opened 2020-05-12
Filter annotations in Largo


mzur:
Largo could be used (by an expert) to review the annotations of a single other user. To do this, the annotations need to be filterable by users. The reviewer/expert can select one (or more) user(s) to see only the annotations of the selected user(s) in Largo.

**Summary:** Filter by
- [x] user
- [x] annotation shape
- [x] file
- [ ] date of the image
- [ ] date of the annotation
- [ ] date of the annotation label
- [ ] annotation session (volume Largo only)
- [x] volume(s) (project Largo only) https://github.com/biigle/core/issues/1271

Ideally, the filters could be combined just like in the volume overview and support less than/equals/more than rules (see https://github.com/biigle/core/issues/359).

mzur:
The next filter to implement should be "filter by file".

PR #1237, opened 2025-07-23
Implements Confidence filtering for Largo

dlangenk:
Allows to filter annotations in largo by confidence as proposed in #1166.

### DIFF
--- a/resources/assets/js/largo/components/annotationFilter.vue
+++ b/resources/assets/js/largo/components/annotationFilter.vue
@@ -109,11 +109,19 @@ export default {
         return {
             filterValues: {
                 Shape: availableShapes,
-                User: {}
+                User: {},
+                Confidence: {
+                    0: 'Very low (0 - 0.25)',
+                    1: 'Low (0.25 - 0.5)', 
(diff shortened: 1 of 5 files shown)


app/Services/AnnotationSizeCalculator.php
mzur: Why is this here?

resources/assets/js/largo/components/annotationFilter.vue
```diff
@@ -109,11 +109,19 @@ export default {
...
-                User: {}
+                User: {},
+                Confidence: {
+                    0: 'Very low (0 - 0.25)',
+                    1: 'Low (0.25 - 0.5)', 
+                    2: 'Medium (0.5 - 0.75)',
+                    3: 'High (0.75 - 0.9)',
+                    4: 'Very high (0.9 - 1.0)'
+                }
             },
```
mzur: This could be a chance to add `>` and `<` operators to filter conditions. Instead of fixed intervals, users can enter one rule `>0.25` and another `<0.5` to get the same thing.

PR #1249, opened 2025-08-21
Video object tracking counter

### LINKED ISSUE #1086: Fix video object tracking counter
When a video object tracking job is started, a cache value is incremented that is used to limit the number of concurrent tracking jobs for each user. The value is decremented, when the tracking job finishes or fails. The job class also has [`$deleteWhenMissingModels = true`](https://github.com/biigle/core/blob/02a383447dfac28a70023ac4840fcc08caba687a/app/Jobs/TrackObject.php#L53), though. This could delete a tracking job without decrementing the cache value (if the new annotation was deleted before the tracking job started). This could stack up so users won't be allowed to start new tracking jobs although none are actually running.

Remove `$deleteWhenMissingModels` from the job class and handle a missing annotation or use model manually. In all cases, the cache value must be decremented or deleted again.

aneust:
This PR ensures that the concurrent tracking job counter for a user is always correctly decremented, even when the related annotation model is missing before the job starts.
Previously, the $deleteWhenMissingModels = true property in the job class could cause jobs to be discarded without executing any cleanup logic, which left the cache counter incremented indefinitely. As a result, users could become blocked from starting new tracking jobs despite having none running.

Removed $deleteWhenMissingModels = true from the tracking job class.
Added explicit check in the handle() method for annotation existence.

Closes #1086

### DIFF
--- a/app/Http/Controllers/Api/VideoAnnotationController.php
+++ b/app/Http/Controllers/Api/VideoAnnotationController.php
@@ -18,6 +18,7 @@ use Illuminate\Validation\ValidationException;
 use Queue;
 use Symfony\Component\HttpFoundation\StreamedJsonResponse;
 use Symfony\Component\HttpKernel\Exception\TooManyRequestsHttpException;
+use Illuminate\Support\Facades\Log;
 
 class VideoAnnotationController extends Controller
 {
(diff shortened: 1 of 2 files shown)


aneust:
It is still a draft

Issue #1271, opened 2025-10-06
Volume filter in project Largo


mzur:
In the Largo view of a whole project we want to implement a "volume" filter. Users can select one volume from a list of all volumes of the project and the shown annotations will be limited to this volume. This can be combined with multiple filter rules so users will be able to evaluate a subset of volumes of the same project in Largo.

PR #1273, opened 2025-10-10
Add volume to largo filters

dbrembilla:
Targets #1271 - Introduces a Volume filter. This allows users to filter for volumes when using Largo on a Project.


mzur (approved):
All good, thanks!

Issue #1040, opened 2020-05-26
Split full report if one file would have too many rows


mzur:
Reject a full report request if it should contain more than 50000 annotations for a single volume. Even though the full report Python script is now more memory efficient, such a report can generate extremely large temporary files (>1.5 GB for a volume with 80000 annotations) and can contain more than the number of rows that can be handled by Excel or Calc (if freehand polygons are used).

dbrembilla:
> The check for how "big" a volume is would be very expensive and would have to be performed both for the UI (to show the relevant elements) and in the backend (to validate the request for the new report).

I tried this SQL query for the check:
```
select 1 from image_annotations
join images on images.id = image_annotations.image_id
join project_volume on images.volume_id = project_volume.volume_id 
where project_volume.project_id = 2
limit 1 offset 5000;
```
for the project above it takes 4 ms on a hit, if there are many volumes it can take much more time, especially if it does not hit.

```
select 1 from image_annotations
join images on images.id = image_annotations.image_id
where images.volume_id = 6
limit 1 offset 5000;
```
for the biggest volume and it takes 1.3 ms

These are all very little, so I cannot really say what happens for big projects/volumes with thousands of images.

Using count is not an option here, it takes too long. A cool idea would be to use a stored procedure / prepared statement here (counting how many annotations there are in a project is something that I think is used in the charts as well and might benefit from some type of optimisation), but it's probably overkill.

Another possibility is changing the behaviour of the class behind the scenes (e.g. the class counts the number of annotations and decides to split the excel in multiple files during the job), but that's very ugly.

> What about dynamically adding a new worksheet in the [Python script](https://github.com/biigle/core/blob/master/resources/scripts/reports/full_report.py) if the number of rows is too large?

This would of course make the files easier to handle for Excel. However this also makes the files a little bigger (see attachment, one changes sheet after 500 annotations, the other once it reaches 500 lines in the Excel).

I'd say this would be the way to go if we just want the row problem to be handled. We can apply changes to the Full report/create a new report that handles multiple files in the future.

[2_full_image_annotation_report_newsheetper500ann.zip](https://github.com/user-attachments/files/23120696/2_full_image_annotation_report_newsheetper500ann.zip)
[2_full_image_annotation_report_newsheetper500lines.zip](https://github.com/user-attachments/files/23120697/2_full_image_annotation_report_newsheetper500lines.zip)

mzur:
[...]

PR #1281, opened 2025-10-29
Split big xlsx reports

dbrembilla:
This PR addresses #1040 - It changes the python scripts `full_report.py` and `csvs_to_xslx.py` so that reports that are too big are split into multiple sheets.
For now, the limit of rows is set to 5000 - I am not sure if that is too low or too big. It could also be set via environment variable using `python-dotenv` potentially.


mzur (approved):
Thanks. I updated the row limit to 100k.

Issue #1035, opened 2022-06-20
Add options to include metadata and/or attributes in CSV reports


mzur:
Add a new option for annotation CSV reports to include the `attributes` column in the reports. If the option is not checked, the column is left out. This could break existing scripts that parse the CSV report and use one of the two columns after `attributes`. Maybe think about this again?

Also add a new option to include image/video metadata for each annotation. While the metadata would be included in `attribtes` as well, it's easier to parse as separate columns. In the case of video annotations, metadata could be interpolated. Make this clear in the documentation or add an option to either interpolate or use the "nearest" data point.

mzur:
Adding metadata columns to the report is now tracked in https://github.com/biigle/core/issues/1065

This issue is only about adding a switch to remove the attributes column from CSV reports (opt-out).

PR #1295, opened 2025-11-05
Allow exclude attributes report

dbrembilla:
This PR addresses [#1035](https://github.com/biigle/core/issues/1035) - Adds a checkbox in the CSV reports that makes so that the `attributes` column is skipped when generating the report.
image

 PS: This PR contains changes from the https://github.com/biigle/core/pull/1281, I will keep this in draft until that PR is done

### DIFF
--- a/resources/views/manual/tutorials/reports/reports-schema.blade.php
+++ b/resources/views/manual/tutorials/reports/reports-schema.blade.php
@@ -193,7 +193,7 @@
             <li>
                 <strong>Additional attributes of the image</strong>
                 <p>
-                    The additional attributes of the image are encoded as a JSON object. The content may vary depending on the BIIGLE modules that are installed and the operations performed on the image (e.g. a laser point detection to calculate the area of an image).
+                    The additional attributes of the image are encoded as a JSON object. The content may vary depending on the BIIGLE modules that are installed and the operations performed on the image (e.g. a laser point detection to calculate the area of an image). This column can be hidden via the <code>Hide the attributes column</code> checkbox.
                 </p>
             </li>
             <li><strong>Annotation ID</strong></li>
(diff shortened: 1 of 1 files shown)


resources/views/projects/reports.blade.php
```diff
@@ -150,6 +151,17 @@
...
                     </div>
                 </div>
+
+                <div v-cloak v-if="wantsCombination('ImageAnnotations', 'Csv') || wantsCombination('VideoAnnotations', 'Csv')" class="form-group">
+                    <div class="checkbox">
+                        <label>
+                            <input type="checkbox" v-model="options.skip_attributes"> Skip attribute column
+                        </label>
+                    </div>
+                    <div v-else class="help-block">
```
mzur:
```suggestion
                    <div class="help-block">
```

resources/views/volumes/reports.blade.php
```diff
@@ -170,6 +170,16 @@
...
                     </div>
                 </div>
+                <div v-cloak v-if="wantsCombination('ImageAnnotations', 'Csv') || wantsCombination('VideoAnnotations', 'Csv')" class="form-group">
+                    <div class="checkbox">
+                        <label>
+                            <input type="checkbox" v-model="options.skip_attributes"> Skip attribute column
+                        </label>
+                    </div>
+                    <div v-else class="help-block">
+                        Skip the attribute column in the CSV report
```
mzur:
```suggestion
                        Skip the attribute column
```

mzur (changes requested):
I would call it "hide attribute column" rather than "skip attribute column". What do you think?

If the column should be hidden, `images.attrs` or `videos.attrs` could be removed from the DB queries, too.

Please add a hint to the manual article that the column can be hidden for a more compact report. This info can be added to the description of the attributes column in the list.

mzur (approved):
Thanks!

Issue #1232, opened 2025-07-18
LabelBOT for tiled images


mzur:
LabelBOT could extract a screenshot of the OpenLayers canvas for inference if a tiled image is shown (regular images keep using the original). This could be done in the annotationCanvas, i.e. if LabelBOT is active the annotationCanvas passes on either the image patch or the whole feature vector in the event to the annotatorContainer along with the OpenLayers feature.

PR #1313, opened 2025-11-19
Resolve LabelBOT for tiled images #1232

yannik131:
Resolves Issue #1232 by taking a screenshot of the map if the image is tiled.

### DIFF
--- a/resources/assets/js/annotations/components/screenshotButton.vue
+++ b/resources/assets/js/annotations/components/screenshotButton.vue
@@ -12,6 +12,7 @@
 import Events from '@/core/events.js';
 import Messages from '@/core/messages/store.js';
 import Keyboard from '@/core/keyboard.js';
+import { trimCanvas } from '../utils';
 
 /**
  * A button that produces a screenshot of the map


mzur:
Ja das passt so, kein Problem :+1:

resources/assets/js/annotations/mixins/labelbot.vue
mzur: I don't like that the bounding box is now used only to identify the message from the worker. This can be done more efficiently with a global message ID, for example. Add `labelbotMessageId: 0` to the data method and increase the ID for each message sent to the worker. Send the current ID along with `postMessage()` and use it to identify the message in `addLabelbotWorkerListener()`. Then you can move `getBoundingBox` from `utils.js` into the `labelBot.vue` mixin (of the annotation canvas).

resources/assets/js/annotations/components/annotationCanvas/labelBot.vue
```diff
@@ -2,6 +2,8 @@
 import LabelbotIndicator from '../labelbotIndicator.vue';
 import LabelbotPopup from '../labelbotPopup.vue';
 import { LABELBOT_STATES } from '../../mixins/labelbot.vue';
+import { makeBlob } from '../../makeBlobFromCanvas.js';
```
mzur: I think it's ok to put this into `utils.js` too, so we don't have an extra file for this.

mzur (changes requested):
This works great already! My comments are mostly about stylistic issues.

resources/assets/js/annotations/components/annotationCanvas.vue
```diff
@@ -570,6 +571,20 @@ export default {
...
+            {
+                try {
+                    labelbotImage = await this.createLabelbotImage(points);
+                } 
+                catch(error) {
+                    this.annotationSource.removeFeature(e.feature);
+                    Messages.danger(error.message);
+                    return;
+                }
+            }
```
mzur:
```suggestion
            if (this.labelbotIsActive) {
                try {
                    labelbotImage = await this.createLabelbotImage(points);
                } catch(error) {
                    this.annotationSource.removeFeature(e.feature);
                    Messages.danger(error.message);
                    return;
                }
            }
```

resources/assets/js/annotations/components/annotationCanvas/labelBot.vue
```diff
@@ -50,6 +54,126 @@ export default {
...
+            const height = maxY - minY;
+
+            return [minX, minY, width, height];
+        },
+        makeMapScreenshot() {
+            const promise = new Promise(resolve => {
+                this.tiledImageLayer.once('postrender', event => {
+                    makeBlob(event.context.canvas).then(blob => {
+                        resolve(createImageBitmap(blob));
+                    });
```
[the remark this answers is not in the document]
yannik131: I moved the screenshot code into `createLabelbotImageFromTiledImage` to handle the canvas synchronously directly after the postrender event and now it works.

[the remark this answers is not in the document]
yannik131: It's already ready. I applied all your suggestions and tested it locally. Everything seems to works fine.

Issue #787, opened 2024-02-21
Allow users to adjust the order of label trees


mzur:
Users wished to be able to adjust the order of label trees. This is probably best implemented as a per-user setting. One way how this could be done is to allow users to sort the label trees in the project label trees tab. The order is then remembered in local storage and applied everywhere the label trees are displayed (volume overview, annotation tools, largo, ...). If this is implemented in the label trees component (which only requires the project ID as key), it should automatically work everywhere.

PR #1323, opened 2025-12-05
787 allow users to adjust the order of label trees

dbrembilla:
Addresses #787 - Allows users to adjust the order of label trees

### DIFF
--- a/resources/assets/js/label-trees/components/labelTrees.vue
+++ b/resources/assets/js/label-trees/components/labelTrees.vue
@@ -192,9 +192,6 @@ export default {
         hasFavourites() {
             return this.favourites.length > 0;
         },
-        hasCustomOrder() {
-            return this.customOrder.length > 0;
-        },
         ownId() {
             if (this.id) {
                 return this.id;


resources/assets/js/label-trees/components/labelTrees.vue
```diff
@@ -20,126 +17,151 @@
...
+            .flat()
+
+        let customOrderStorageKeys = [];
+        projectIds.forEach(
+            function (el) {
+                if (Number.isInteger(el)) {
+                    return customOrderStorageKeys.push(`biigle.label-trees.${el}.custom-order`)
+                }
+            }
+        );
```
mzur: You should add these as a new prop of the component instead of hard-coding the variable names.

mzur (changes requested):
- There are lots of coding style changes in here (e.g. changing `'` to `"`, rmoving trailing commas, adding spaces). Please revert all these.

- Please add the underline style of the label tree title and show the sorting buttons only when the user hovers over the title (not when hovering over the entire tree).

- ~~Please also add the sorting to the label trees tab of the project overview.~~

- You implemented the new behavior as opt-in in the component. I would choose always on instead (as the user has to actively change the order first before anything changes from the default).

    You can still use the new property to disable the sorting for the "Favourites" label tree in the label trees component. Currently this is handled inside the label tree component but this component has nothing to to with controlling (or knowing about) the Favourites tree.

- With some label tree names, the buttons can loose alignment and overflow to the bottom (see screenshot below). You should use a fixed `padding-right` for the label tree title and wrap the buttons in an element with `position: absolute`.

    Image

- Please see the comment about linking the issue above.

resources/assets/js/label-trees/components/labelTrees.vue
```diff
@@ -189,11 +214,17 @@ export default {
...
+            return ids.sort().join('-');
         },
         favouriteStorageKey() {
             return `biigle.label-trees.${this.ownId}.favourites`;
         },
+        treeIds() {
+            return this.trees.reduce((els, obj) => {
+                els.push(obj.id);
+                return els;
+            }, []);
```
mzur:
```suggestion
            return this.trees.map(tree => tree.id);
```

resources/views/largo/show.blade.php
```diff
@@ -6,6 +6,7 @@
     <script type="module">
         biigle.$declare('largo.user', {!! $user !!});
         biigle.$declare('largo.volumeId', {!! $volume->id !!});
+        biigle.$declare('largo.projectIds', {!! $volume->projects()->pluck('id') !!});
```
mzur: Similar to above but here you can just do:

```suggestion
        biigle.$declare('largo.projectIds', {!! $projects->pluck('id') !!});
```

dbrembilla:
I tried to increase the padding to `3 em`, seems to be okay in my testing

resources/assets/js/label-trees/components/labelTrees.vue
```diff
@@ -242,6 +239,7 @@ export default {
             this.events.emit('clear');
         },
         handleAddFavourite(label) {
+            //TODO: here label.favourite does not get set to true. THIS MIGHT BE BECAUSE WE DONT USE A PROP ANYMORE TO CREATE LABELTREES
```
mzur: Is this still TODO?

resources/assets/js/label-trees/components/labelTrees.vue
```diff
@@ -242,6 +271,27 @@ export default {
...
+            return arr;
+        },
+        changeCustomOrder() {
+            let customOrder = JSON.parse(
+                localStorage.getItem(this.customOrderStorageKey)
+            );
+            if (customOrder) {
+                this.customOrder = customOrder;
+            }
+        }
```
mzur: This also seems to be used nowhere. Please check your changes for any other unused stuff.

resources/assets/js/label-trees/components/labelTrees.vue
```diff
@@ -157,6 +146,17 @@ export default {
...
+            let customOrderStorageKeys = [];
+            this.projectIds.forEach(
+                function (el) {
+                    if (Number.isInteger(el)) {
+                        customOrderStorageKeys.push(`biigle.label-trees.${el}.custom-order`)
+                    }
+                }
+            );
+            return customOrderStorageKeys;
+        },
```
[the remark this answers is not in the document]
mzur: The key could also be confusing. What about:
```js
`biigle.projects.${id}.label-trees.custom-order`
```

resources/assets/js/label-trees/components/labelTree.vue
```diff
@@ -174,11 +194,20 @@ export default {
             return this.compiledLabels[null];
         },
         collapseTitle() {
-            return this.collapsed ? 'Expand' : 'Collapse';
+            if (this.collapsible) {
+                return this.collapsed ? 'Expand' : 'Collapse';
```
mzur:
```suggestion
                return this.collapsed ? 'Expand the label tree' : 'Collapse the label tree';
```

mzur (dismissed):

PR #1334, opened 2025-12-16
1110 strip directory traversal instructions from volume and file paths

### LINKED ISSUE #1110: Strip directory traversal instructions from volume and file paths
Consider the following path with a directory traversal instruction `user-1://my_images/../my_other_images`. This is currently a valid volume path in BIIGLE. It also works as part of a filename. While it is not possible to access files uploaded by another user this way (which would be a security issue), I want to disallow all `..` from volume URLs and file paths.

Sanitize volume URLs and file paths on any create and update operations by removing the `..`. Alternatively, update the VolumeUrl and VolumeFiles validation rules to disallow any occurrence of `\.\.(\/|$)` in the URL or filename (see regex below). [Here](https://github.com/biigle/core/blob/73bf7f6d0b8b64b1c64d6d76ec1057225fa74fd1/app/Http/Requests/StoreVolume.php#L141) (both), [here](https://github.com/biigle/core/blob/73bf7f6d0b8b64b1c64d6d76ec1057225fa74fd1/app/Http/Requests/UpdateVolume.php#L50), [here](https://github.com/biigle/core/blob/73bf7f6d0b8b64b1c64d6d76ec1057225fa74fd1/app/Http/Requests/UpdatePendingVolume.php#L75) (both) and [here](https://github.com/biigle/core/blob/73bf7f6d0b8b64b1c64d6d76ec1057225fa74fd1/app/Http/Requests/StoreVolumeFile.php#L56).

Example code to strip `../`:
```php
preg_replace('/\.\.(\/|$)/', '', $path);
```

Also add tests with a couple of combinations (`..` at the start, end, without training slash, multiple occurrences).

dbrembilla:
Resolves #1110

### DIFF
--- a/app/Rules/VolumeFiles.php
+++ b/app/Rules/VolumeFiles.php
@@ -5,7 +5,6 @@ namespace Biigle\Rules;
 use Biigle\FileCache\GenericFile;
 use Biigle\MediaType;
 use Biigle\Volume;
-use Biigle\Rules\VolumeUrl;
 use Exception;
 use FileCache;
 use Illuminate\Contracts\Validation\Rule;


app/Rules/VolumeUrl.php
```diff
@@ -173,4 +186,19 @@ protected function isDeniedProvider($value)
...
+     *
+     * @return boolean
+     */
+    protected function pathIsValid($path)
+    {
+        if (preg_match('/(\/|\\\\)*(\.\.)+(\/|\\\\)*(.)*/', urldecode($path)) !== 0) {
+            return false;
+        }
+        return true;
+    }
```
mzur: Maybe set this as public static function so it can be reused in the `VolumFiles` rule?

app/Rules/VolumeUrl.php
```diff
@@ -84,6 +84,11 @@ protected function passesRemoteUrl($value)
             return false;
         }
 
+        if (!$this->pathIsValid($value)) {
+            $this->message = 'The URL inserted is not valid. Please do not use path traversals.';
```
mzur:
```suggestion
            $this->message = 'Volume URLs with path traversal instructions are not allowed.';
```

app/Rules/VolumeUrl.php
```diff
@@ -124,14 +129,22 @@ protected function passesDiskUrl($value)
             return false;
         }
 
-        if (Gate::denies('use-disk', $url[0])) {
+        $diskName = $url[0];
+        $path = $url[1];
+
+        if (!$this->pathIsValid($path)) {
+            $this->message = "The path inserted is not valid. Please do not use path traversals";
```
mzur:
```suggestion
            $this->message = "Volume URLs with path traversal instructions are not allowed.";
```

Issue #1233, opened 2025-07-18
LabelBOT for videos


mzur:
Implement LabelBOT in the video annotation tool. Everything should look and work exactly as in the image annotation tool. It should work with object tracking as well as multi-frame annotations (where it will use the first frame to generate the feature vector).

To get the feature vector of the first frame, it may have to be extracted right when the shape of the first frame is drawn. Otherwise the video will no longer show the correct frame. The vector then can be held back until the annotation is finished.

There could be a problem is users quickly continue playing the video or if "play and pause while drawing" is active. Getting a picture of the current frame before the video continues playing must be real quick. Computing the vector from the picture then can take longer while the video already continues.

yannik131:
My notes from working on the issue:
```
Videos hierarchy:

videos.show.php 
    <div id="video-container" .. -> VideoContainer.vue
        includes videos/show/content.blade.php 
            <video-screen ... -> VideoScreen.vue
        includes videos/show/sidebar-labels.blade.php 
            Contains the labelbot button

Adding an annotation for an image:
--------------------------------------

resources/assets/js/annotations/components/annotationCanvas/drawInteractions.vue:
    .on('drawend'): this.handleNewFeature(e)

resources/assets/js/annotations/components/annotationCanvas.vue:
    async handleNewFeature(e):
        this.$emit('new', {
                shape: geometry.getType(),
                points: points,
                labelbotImage: labelbotImage
            }, removeCallback);
            
resources/assets/js/annotations/annotatorContainer.vue:
    handleNewAnnotation(annotation, removeCallback):
        promise = this.storeLabelbotAnnotation(annotation);
        promise.then(this.setLastCreatedAnnotation);
        
resources/assets/js/annotations/mixins/labelbot.vue:
    storeLabelbotAnnotation(annotation):
        return this.generateFeatureVector(annotation.labelbotImage)
            .then(featureVector =>  annotation.feature_vector = featureVector)
            .then(() => AnnotationsStore.create(currentImageId, annotation))
            .then((annotation) => {
                if (currentImageId === this.imageId) {
                    this.showLabelbotPopup(annotation);
                }
                return annotation;
            })
                    
resources/assets/js/annotations/mixins/labelbot.vue:
    generateFeatureVector(labelbotImage):
        this.labelbotWorker.postMessage({ type: 'run', ... });
        return this.addLabelbotWorkerListener(
                e => labelbotMessageID === e.data?.labelbotMessageID
            ).then(e => e.data.vector);
            
[...]

PR #1336, opened 2026-01-05
Resolve LabelBOT for videos #1233

yannik131:


mzur:
Thanks! The number of popups actually no longer has a config option. It's set with a constant [here](https://github.com/biigle/core/blob/4ea4ecfc7caa7f823dc11af7f9ce91ff7457232c/resources/assets/js/annotations/mixins/labelbot.vue#L10).

PR #1343, opened 2026-01-09
923 has no number of images incorrect volume filter

### LINKED ISSUE #923: "Has no" number of images incorrect (Volume filter)
If you want to filter a volume and e.g. select "has annotations", the number of images containing annotations is correctly shown. However if you select "has no annotations" the same number is still displayed.

image

dbrembilla:
Resolves #923 
image

I took the liberty of doing some renaming, I think that without double negations it is a little clearer to understand what is going on. If you don't like it I can go back.
In general, I think the functioning of `filterTab` and its interactions with `filterListComponent` etc. could be improved, but I wanted this PR to be smaller in scope. Let me know what you think. 
I am also unsure about how is the method `refreshRule` triggered, so I patched it to be aligned with `ruleAdded` but without adding the rule to `this.rules`.

### DIFF
--- a/app/Rules/VolumeFiles.php
+++ b/app/Rules/VolumeFiles.php
@@ -107,6 +107,12 @@ class VolumeFiles implements Rule
 
                 return false;
             }
+
+            if (VolumeUrl::pathHasDirectoryTraversal($filename)) {
+                $this->message = 'Filenames with path traversal instructions are not allowed.';
+
+                return false;
+            }
         }
 
         if ($this->typeId === MediaType::imageId()) {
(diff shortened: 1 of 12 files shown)


resources/assets/js/volumes/components/filterTab.vue
```diff
@@ -74,46 +74,46 @@ export default {
                 return this.fileIds;
             }
 
-            let only = {};
-            let except = {};
-            let negatedRules = 0;
-            let nonNegatedRules = 0;
+            let filteredFiles = {};
+            let exceptionFiles = {};
```
mzur: I propose:

```suggestion
            let filesToKeep = {};
            let filesToExclude = {};
```

PR #1358, opened 2026-01-29
Added requirements and tracking script for SAM-2

mwuel:
A custom frame by frame video predictor [https://github.com/Gy920/segment-anything-2-real-time](url) has been used to implement tracking using the SAM-2 model by Facebook. The original predictor [https://github.com/facebookresearch/sam2](url) was also considered but not implemented because it required loading the whole video or at least chunks of it in-memory, making it less elegant for larger video files.

The script works by extracting a bounding box from the predicted mask by SAM-2. Tracking via a bounding box is therefore very straightforward to implement in the future. For point and circle tracking, this also means that the point is just an interpolated mask and might show unwanted behavior. Especially point tracking may look very odd, as the point given as output will just be the average over the whole mask. The point might end up outside of the original object if the shape is not simple enough (i.e. big wingspan of birds might enlargen the bounding box, shifting the middle point outside of bounds). 
The bounding box will always keep the dimensions of the initial input so it can be easily adjusted. Using polygons as an input for tracking has not been tried but might be possible for even more accurate results. Scripts can be easily switched between old and new by modifying the _videos.php_ config file. 
Currently, the script is set to track every second frame only. Tracking every frame significantly hinders computational performance and skipping more than one frame showed worse tracking results because of missing context. The number of frames to be skipped can be easily adjusted in the script.

### DIFF
--- a/.docker/gpu-worker.dockerfile
+++ b/.docker/gpu-worker.dockerfile
@@ -26,13 +26,30 @@ RUN apt-get update \
         build-essential \
         git \
         libvips \
+        wget \
+    && wget -qO /tmp/cuda-keyring.deb https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2204/x86_64/cuda-keyring_1.1-1_all.deb \
+    && dpkg -i /tmp/cuda-keyring.deb \
+    && apt-get update \
+    && apt-get install -y --no-install-recommends \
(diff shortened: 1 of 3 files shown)


mzur:
Revert the logic that keeps the same size for circles. It should be possible to change the circle size during tracking (e.g. when the camera moves closer to the object).

PR #1390, opened 2026-03-16
Annotation Guideline

### LINKED ISSUE #235: Annotation strategy
Spin off of BiodataMiningGroup/biigle-core#93

Let users upload a PDF with their annotation strategy for a project. The PDF can be displayed in-app or be supplied as download.

dbrembilla:
Resolves #235 
(to be merged only when the feature is complete)
- [x] Adds Annotation <del>Strategy</del> Guideline to the available tabs in the project view.
- [x] Annotation <del>Strategy</del> Guideline can be edited by admins and viewed by editors.
- [x] Annotation <del>Strategy</del> Guidelines can have labels associated. A description, a preferred shape and a reference image can be added
- [x] Users can choose a set of allowed shapes (independent of a label) for enforced guidelines. Only these tools will be available/allowed for annotation.
- [ ] Annotation <del>Strategy</del> Guidelines can be enforced for the annotation process, making it so that only labels within a strategy can be used and the preferred shapes are  enforced (validated in the backend)
- [ ] Enforcement applies to Largo as well, making it so that only labels with a  <del>Strategy</del> Guideline can be used and the preferred shapes are  enforced (i.e. a label with shape point cannot be changed to a label with shape circle)
- [ ] Enforcement applies to LabelBOT as well, so it only suggests "allowed" labels (for the current shape)
- [ ] When annotating, description and reference image are displayed


mzur (commented):
High-level review first:

I thought about the name "annotation strategy" again. This is not really a "plan of action" (although this could be explained in the description) but rather a "set of rules" with some additional information attached. If this is enforced, it is more like a "policy". But since enforcement should be optional, maybe it could be a "guideline"? The thought occurred to me when I saw the tab "strategy" and had no idea what it really meant.

---

I now definitely prefer the alternative UI I mentioned earlier. Below is a mockup. We can also discuss this again before you start the implementation, so it will turn out "right" and you don't have to do duplicate work.

2026-04-09-mockup-annotation-guideline

The UI consists of a block with the description at the top and a block for the labels below. The description can turn into "edit mode" with a click on the edit button. There, the description becomes a textarea and can be changed. Also a "power toggle" appears where the admin can toggle "enforcement" of the strategy/guideline. The edit button is replaced with a discard and a save button. When any of these is pressed, edit mode is disabled again. If a guideline is enforced, this is indicated with something next to the title (also for non-admins).

On the left of the labels block there is the label trees component with all label trees of the project. When a label is selected, the label description, shape and reference image appear to the right. Admins can edit these. If a label has a description/shape/image, this is indicated with an info icon in the label trees. This can be an example for the UI in the annotation tools later. Since a label can also have a star icon if it is favorite, both icons must work.

If a guideline is enforced, labels without description/reference/image are shown with a muted color in the label trees component. Maybe we need an extra property to "include" a label in a guideline even if it has no description/shape/image?

Each label should be saved individually after it was edited, not only after the whole strategy/guideline was saved (as it is done now).

---

The body of the strategy/guideline tab overflows the main body container on two levels (see below). Watch the use of `container` and `row` classes.

Image

---

Currently when I create a strategy with only a description, I get the error "The reference images field is required." The strategy is saved though.

dbrembilla:
Thanks for the feedbacks and for the prospect of the UI! FYI, that's a PNG and the background is transparent so it's kinda hard to read on the dark background in github. 

The guideline naming also makes sense.

To signal that a guideline is enforced, could a [bolt](https://fontawesome.com/icons/bolt?f=classic&s=solid) icon make sense? 

>On the left of the labels block there is the label trees component with all label trees of the project. When a label is selected, the label description, shape and reference image appear to the right. Admins can edit these


Should the labels have an `Edit`  and `Save` button as well, so that it can be individually saved (similarly to what it is now)? I would show it besides the `Label description` title, similarly to the annotation guideline.

>Each label should be saved individually after it was edited, not only after the whole strategy/guideline was saved (as it is done now).

In order to have the annotation guideline before the label, I would probably show the label components after the strategy is initially saved, what do you think?

>If a label has a description/shape/image, this is indicated with an info icon in the label trees. 

This makes sense, we could also make it so that when hovering on the info icon the information about the annotation guideline for the label appear, or that clicking that would either send you to the label in the annotation guideline or display the info somewhere in the UI. I would implement this after the tab is set up, for now I am just gonna add the icon.

>If a guideline is enforced, labels without description/reference/image are shown with a muted color in the label trees component. Maybe we need an extra property to "include" a label in a guideline even if it has no description/shape/image?

We could also simply make it so that all properties of a label within a guideline are nullable, so that users that want to leave all the freedoms do not have to use all of them.

mzur:
> Should the labels have an Edit and Save button as well, so that it can be individually saved (similarly to what it is now)? I would show it besides the Label description title, similarly to the annotation guideline.

When an admin selects a label, it is automatically in "edit" mode. But we need a save button, yes.

> In order to have the annotation guideline before the label, I would probably show the label components after the strategy is initially saved, what do you think?

This can be handled by the JS in the background for better UX. If the users saves a label before they created the guideline, the JS creates the guideline first (with empty description) and the saves the label.

> This makes sense, we could also make it so that when hovering on the info icon the information about the annotation guideline for the label appear, or that clicking that would either send you to the label in the annotation guideline or display the info somewhere in the UI. I would implement this after the tab is set up, for now I am just gonna add the icon.

I think I prefer to show the info in the label indicator that currently only shows the label name at the bottom right of the annotation tools. But let's talk about this when we get there.

> We could also simply make it so that all properties of a label within a guideline are nullable, so that users that want to leave all the freedoms do not have to use all of them.

Sounds good.

mzur:
Fyi, GitHub now supports [stacked PRs](https://github.github.com/gh-stack/). I think you don't even need the `gh` CLI for that. Just open a new PR targeting the branch of this PR and it will be a stack. We can use this when annotation guidelines are implemented in the annotation tools and Largo.

dbrembilla:
> Fyi, GitHub now supports [stacked PRs](https://github.github.com/gh-stack/). I think you don't even need the gh CLI for that. Just open a new PR targeting the branch of this PR and it will be a stack. We can use this when annotation guidelines are implemented in the annotation tools and Largo.

Pretty cool feature! It's good to know

Btw, with the new UI there is some more space for the image. I would cap the maximum size to display to 300 pixels.  
Would be something like this:

https://github.com/user-attachments/assets/c0d3d2e3-54a7-42d3-9b7d-5108e30886a5

mzur:
Neat. I may jump in at the end and move some buttons around. Just let me know when this is ready to review again.

config/annotation_guideline.php
mzur, line 1: Please call the file `projects.php` and the key `annotation_guideline_storage_disk`.

tests/files/test-image-small.jpg
mzur, line 1: Please use one of the two existing test images.

resources/assets/js/projects/components/annotation_guideline/annotationGuidelineLabelImage.vue
```diff
@@ -0,0 +1,132 @@
...
+                title="Remove this image"
+            >
+                <span class="fa fa-times"></span>
+            </button>
+        </span>
+    </div>
+    <div v-else>
+        <span>No reference image was provided</span>
+        <div v-if="isAdmin && editable">
+            <label>Select a reference image (max 5 MB)</label>
```
mzur: As far as I can see the 5 MB limit is never enforced. BIIGLE allows much larger uploads by default. With the client-side resizing, the file size should be fine in most cases and we don't need the hint here. We should add a server-side validation for the size (e.g. the 5 MB) to be sure and only show this as a validation error here if the size is actually exceeded.

mzur (changes requested):
Another high-level review. I'll also request a new Copilot review for good measure. Next time I'll have a more detailed look. Comments:

- Allow guidelines with an empty description. If the user adds a label before the description, create an "empty" guideline in the background first and then attach the saved label to the guideline.

- The "label saved" notification pops up each time I select/unselect a recently saved label.

- When I click on any label that has no image I get a request "`http://localhost:8000/storage/annotation-guideline-reference-images/1/7.jpg?v=NaN`. The UI should know about all saved labels and if they have images and only send requests if the label should actually have an image.

tests/files/test-image-small.jpg
dbrembilla: I added the same image but resized because the test image was too big

mzur: There is a smaller `test-image.png`.

dbrembilla: Up to now I limited the possibility to upload images only to jpg, which format should be allowed?

mzur: jpg, png and webp.

resources/assets/js/label-trees/components/labelTrees.vue
```diff
@@ -170,8 +176,12 @@ export default {
                 Array.prototype.push.apply(labels, tree.labels);
             });
 
+            if (this.labelsInGuideline.length > 0) {
```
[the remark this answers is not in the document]
dbrembilla: Solved by adding an "enforceGuideline" attribute

config/annotation_guideline.php
mzur, line 1: The `projects.php` config is missing now.

app/Http/Requests/StoreAnnotationGuidelineLabel.php
```diff
@@ -0,0 +1,50 @@
...
+     * The rules that the request should follow
+     *
+     * @return array
+     */
+    public function rules(): array
+    {
+        return [
+            'label' => ['required', 'integer'],
+            'description' => ['nullable', 'string'],
+            'shape' => ['nullable', 'integer'],
```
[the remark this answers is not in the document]
mzur: Still relevant.

mzur (commented):
I'll continue this PR so the comments are for me.

mzur:
@dbrembilla I will implement the "enforce" switch and the "users can choose a set of allowed shapes (independent of a label)" feature in the show/edit annotation guideline view. After that you can continue to actually implement the guideline in the annotation tools and Largo.

mzur:
Thinking about this, "enforcing" a guideline has so many implications. Maybe we should discuss these in the group meeting again:

- Is the set of shapes is limited (independent of labels), can the admin still choose one of these for a single label in the guideline? Example: Only circles and polygons are allowed, label A is limited to circles, label B to polygons and label C is not configured so it can be annotated either with circles or polygons.

-  Users cannot attach a label to an annotation that has an "incorrect" shape (using the attach or swap tools).
    - We need a helpful error message for this. Users should create a new annotation instead (and maybe delete the old one). 

- Largo has to check the shape of each annotation to decide if the label may be attached (or swapped). If anything is wrong, the whole Largo session is not saved.
   - Ignore and not save these cases? Similar to changing the label of other users? Maybe add a new feedback message that this happened. If some selections were not changed, these should still be selected in Largo. Add a "preflight" parameter to check a request without applying it.

- LabelBOT can only suggest labels that have the shape of the (already drawn) annotation. So by choosing the "wrong" shape, users can limit LabelBOT in a way that it can't suggest the correct label any more.
   - LabelBOT should suggest labels as usual but the UI does not allow them to choose labels with "incorrect" shape (or if the labels are not allowed by the guideline in the first place).

mzur:
@dbrembilla I finished the UI to create/edit annotation guidelines. Going forward I suggest that you implement enforcement of the guidelines (for the annotation tools and Largo) in the backend and I do the frontend. We can talk this through on Monday.

I also thought if "annotation instructions" would fit even better because when I explained this feature in the manual article I was always talking about "instructions".

---

Thoughts about the backend implementation:

- The API endpoints to create image/video annotations (also the bulk endpoint) have to check if there is an enforced guideline. If the volume belongs to many projects and one of the projects has an enforced guideline, the request has to provide the guideline ID that should be applied. If one project has an enforced guideline and the other has none, the request can provide `null` to say that no guideline should be applied. If all projects have enforced guidelines, one of these has to be chosen.
- The API endpoints to update annotation labels also have to apply a guideline similar to above.
- Largo should validate the annotation labels and shapes based on the (chosen, see above) guideline. If a label is not allowed or an annotation shape does not match the chosen label, this change should be ignored similar to how currently changes are ignored that would delete labels of other users (unless "force" is enabled). Largo should collect all ignored changes (also the ignored changes because of a different user) and return these to the UI, so these can be displayed there along with a warning/info message.

Issue #1264, opened 2025-09-18
Implement iterative index scans for LabelBOT


mzur:
Currently we have a workaround with `DROP INDEX` in a transaction to perform a slower exact search if the vector search using the HNSW index returns no result. While issuing `DROP INDEX` in application logic is dangerous on its own, I found that the statement also seems to create a locked state during database backups with `pg_dump`. So we need a better solution for this.

pgvector 0.8.0 added [iterative index scans](https://github.com/pgvector/pgvector?tab=readme-ov-file#iterative-scan-options) to solve this issue. We should implement these here, too.

@gkourie could you look into this when you have time?

PR #1401, opened 2026-03-23
Fix LabelBOT iterative index scan logic

gkourie:
Removed the fallback within‑subquery‑filtering query because the iterative index scan does not rely on it as I thought.

### DIFF
--- a/app/Services/LabelBot/LabelBotService.php
+++ b/app/Services/LabelBot/LabelBotService.php
@@ -121,8 +121,10 @@ class LabelBotService
     {
         // This has to be wrapped in a transaction because it uses SET LOCAL.
         return DB::transaction(function () use ($featureVector, $trees, $model) {
+            $k = (int) config('labelbot.K');
+
             // Perform ANN search.
-            $topNLabels = $this->performAnnSearch($featureVector, $trees, $model);
+            $topNLabels = $this->performAnnSearch($featureVector, $trees, $model, $k);
 
             // Perform ANN search with iterative index scan as a fallback if ANN search returns no results.
             if (empty($topNLabels)) {
@@ -133,10 +135,13 @@ class LabelBotService
                 See https://github.com/pgvector/pgvector?tab=readme-ov-file#iterative-index-scans for more details
 
                 We will use relaxed order because it's slightly faster and we are sorting the subquery results anyway.
+
+                For the choice of $k*2, see:
+                https://github.com/biigle/core/pull/1401#issuecomment-4163049404
                 */
                 DB::statement("SET LOCAL hnsw.iterative_scan = relaxed_order");
 
-                $topNLabels = $this->performAnnSearch($featureVector, $trees, $model, fallback: true);
+                $topNLabels = $this->performAnnSearch($featureVector, $trees, $model, $k*2);
             }
 
             return $topNLabels;
@@ -153,17 +158,12 @@ class LabelBotService
      * @param Vector $featureVector The input feature vector to search for nearest neighbors.
      * @param int[] $trees The label tree IDs to filter the data by.
      * @param string $model Class name of the annotation type.
-     * @param bool $fallback Indicates whether this search is being performed as a fallback after no-result ANN search.
+     * @param int $k Number of desired results.
      *
      * @return array The array of label IDs representing the top nearest neighbors.
     */
-    protected function performAnnSearch($featureVector, $trees, $model, $fallback = false)
+    protected function performAnnSearch($featureVector, $trees, $model, $k)
     {
-        $k = (int) config('labelbot.K');
-
-        if ($fallback) {
-            $k *= 2;
-        }
         // Size of the dynamic candidate list during the search process.
         // K is always bounded by this value so we set it to K.
         DB::statement("SET LOCAL hnsw.ef_search = $k");
--- a/app/Services/LabelBot/LabelBotService.php
+++ b/app/Services/LabelBot/LabelBotService.php
@@ -106,11 +106,10 @@ class LabelBotService
     }
 
     /**
-     * Perform vector search using the Dynamic Index Switching (DIS) technique.
+     * Perform vector search using the ANN-Iterative-Fallback technique (ANN-Itr-FB).
      *
      * The search process first attempts to retrieve results using an Approximate Nearest Neighbor (ANN) search
-     * via the HNSW index. If the ANN search returns no results, it falls back to an exact KNN search using the
-     * B-Tree index for filtering, ensuring that results are always returned.
+     * via the HNSW index. If the ANN search returns no results, the iterative index scan is triggered and the ANN search is performed again as a fallback.
      *
      * @param vector $featureVector The input feature vector to search for nearest neighbors.
      * @param int[] $trees The label tree IDs to filter the data by.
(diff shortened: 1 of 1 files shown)


mzur (commented):
Can you please explain again why this change is possible? We thought that the iterative scan only works if the filter condition is in the same subquery than the HNSW index scan. This is now no longer the case.

Also if this indeed works correctly now, why not always enable iterative index scan? It's the same query now and if it returns enough results the first time, no iterative scanning should be triggered and the query should be as fast as before?

gkourie:
> Can you please explain again why this change is possible? We thought that the iterative scan only works if the filter condition is in the same subquery than the HNSW index scan. This is now no longer the case.
> 


It turns out my earlier assumption [#1264](https://github.com/biigle/core/issues/1264#issuecomment-3386221107) was incorrect due to how iterative index scans behave at the session level. What misled me was that once iterative index scan is enabled, it remains active for all following queries in the session. In my tests, after the first fallback occurred, iterative scanning stayed enabled, which reduced fallback rates for the following queries. This also skewed my interpretation of later results. When I encountered a second fallback, I re-enabled iterative index scan and reran the ANN query. Since nothing had actually changed (the query was identical and iterative scan was already active), it still didn’t return results unless the filter condition was placed inside the subquery. This led me to incorrectly conclude that iterative scans only work if the filter is applied within the same subquery as the HNSW scan. However, moving the filter into the subquery sometimes caused the planner to switch to the B-Tree index in some cases. Because this plan consistently returned results (albeit slower), I mistakenly attributed that behavior to iterative scanning working in the background.

> Also if this indeed works correctly now, why not always enable iterative index scan? It's the same query now and if it returns enough results the first time, no iterative scanning should be triggered and the query should be as fast as before?

In my tests, enabling the iterative index scan for all ANN queries increases the average query execution time by ~ 4x compared to the pure ANN query.

mzur:
Is your conclusion still correct now where we enable iterative scanning only within the transaction and reset later?

gkourie:
> Is your conclusion still correct now where we enable iterative scanning only within the transaction and reset later?

If you mean, the iterative scanning is enabled even without within-subquery-filtering, so yes that's correct and it's reset later to be turned off automatically after the transaction is finished.

In my experimental setup, I had to turn it off manually by setting `SET hnsw.iterative_scan = off;`

mzur:
Now I understood why this change could make sense:

- The iterative scan is not only useful if there is a filter condition in the subquery. It also works if the subquery would return less than `k` results. The iterative scan basically only ensures `k` results no matter if there is an additional filter condition or not.
- If we enable iterative scanning *with* filter condition in the subquery, the query can be extremely slow as it wants to ensure `k` results that match the filter condition (and it may have to search hard until it gets `k`). We don't always need `k`, we only want three different labels to suggest.
- If we enable iterative scanning *without* filter condition in the subquery, we get guaranteed `k` results and have a better chance of still getting results after the filtering in the outer query. But we could still end up with 0 results, too.

We want to test what happens if:
- We enable iterative scanning with subquery filtering but reduce `k` from 100 to 3. Is it still slow? Is the accuracy better than with filtering in the outer query? This will have a high chance of returning only a single unique label to suggest but this could be ok in the fallback case.
- We enable iterative scanning without subquery filtering but increase `k` from 100 to 200. Is it still fast? is the accuracy (meaningfully) better than with `k`=100?

gkourie:
The initial LabelBOT vector search is performed with a single query, which include a subquery for ANN and post-subquery label grouping and label tree filtering. So we get up to `k` (default = 100) results after ANN  --> filter by label tree --> group by label --> sort by minimum distance to input vector --> get Top-3.

Let's define some terms:

- **Post-Subquery-Filtering (PSF):** Filtering by label tree occurs outside the subquery (initial LabelBOT vector search).
- **Post-Filtering (PF):** Filtering by label tree occurs inside the subquery.
- **ANN-Itr-FB strategy:** Enable iterative index scan and re-run the ANN search when the initial search returns no results as a fallback. The fallback query can be PSF or PF.


> We enable iterative scanning with subquery filtering but reduce k from 100 to 3. Is it still slow? Is the accuracy better than with filtering in the outer query? This will have a high chance of returning only a single unique label to suggest but this could be ok in the fallback case.


1. **ANN-Itr-FB + PF** as fallback for the dataset X. When the initial ANN + PSF returns no results, we enable the iterative index scan and switch to ANN + PF:

| k | Top-1 | Top-3 | Avg. fallback execution time | Avg. overall execution time |
|--------|--------|--------|--------|--------|
| 3 | 60.57% | 85.35% | **14 ms** | **1.35 ms** |
| 10 | 60.57% | 85.42% | 28.60 ms | 1.51 ms |
| 25 | 60.58% | 85.42% | 42.51 ms | 1.65 ms |
| 50 | 60.58% | 85.42% | 52.81 ms | 1.76 ms |
| 100 | 60.58% | 85.42% | 71.84 ms | 1.97 ms | 
| Baseline (kNN) | 
| 100 | **60.59%** | **90.36%** | - | 35 s | 

These values come from an experimental setup. In the BIIGLE production environment, the average execution times are different, For example, the fallback execution time for some random input is about 7 seconds with `k = 100` and 1.5 seconds with `k = 3`. For another random input, the fallback execution time is about 900 ms with `k = 100` and 233 ms for `k = 3`, despite that, the differences between the results for different values of `k` stay similar. 

> We enable iterative scanning without subquery filtering but increase k from 100 to 200. Is it still fast? is the accuracy (meaningfully) better than with k=100?

2. **ANN-Itr-FB + PSF** as fallback for the dataset X.  When the initial ANN + PSF returns no results, we enable the iterative index scan and re-run the same initial ANN query and either with the same `k` (as current PR implementation) or increase `k`:

| k | Top-1 | Top-3 | Avg. fallback execution time | Avg. overall execution time |
|--------|--------|--------|--------|--------|
| 100 | 60.23% | 84.93% | **4.20 ms** | **1.25 ms** |
| 200 | 60.44% | 85.18% | 7.36 ms | 1.28 ms |
| 300 | **60.51%** | **85.27%** | 12.10 ms | 1.32 ms |


Final thoughts:
* ANN-Itr-FB + PF (even with `k = 3`) resolves most of fallback cases, but can be slow (seconds).
* ANN-Itr-FB + PSF resolves only 40% of the fallback cases (with `k = 100`), but guarantees an execution time of under one second. Setting `k = 200` resolves about 70% of fallback cases and also guarantees an execution time of under one second. `k = 300` could be the edge case for an under one second execution time.
* In this case, ANN-Itr-FB + PF (with `k = 3`) achieved marginal gains in accuracy (< 1%).
* I would go for efficiency over accuracy even if ANN-Itr-FB + PF achieved near kNN accuracy, as the overall number of annotations is increasing rapidly and maintaining an under one second execution time could be challenging.

@mzur what do you think?

mzur:
Ok final question. What is Top-1 and Top-3 if we don't use the fallback at all (i.e. no result=wrong result)?

gkourie:
Top-1: 59.98%, Top-3: 84.57% for this dataset. The fallback-rate for this test dataset is 1%.

mzur:
Ok thanks. Maybe we go with ANN-Itr-FB + PSF/k=200 as it is almost like ANN-Itr-FB + PF/k=3 but faster? Is it still fast (<1s) on the full database? Otherwise we can stick to the current approach with k=100 of this PR. You decide, as you have the full picture.

gkourie:
> Ok thanks. Maybe we go with ANN-Itr-FB + PSF/k=200 as it is almost like ANN-Itr-FB + PF/k=3 but faster? Is it still fast (<1s) on the full database?

yes, it is still fast.

mzur (approved):

PR #1416, opened 2026-04-07
500 modal with available keyboard shortcuts

### LINKED ISSUE #500: Modal with available keyboard shortcuts
Implement a modal with the available keyboard shortcuts in the image and video annotation tools. The modal could maybe be opened with a button from the settings tab. The content should be the same than in the manual (maybe implement this as partial template so the information is not redundant).

yannik131:
closes #500


resources/assets/js/core/components/shortcutsButton.vue
```diff
@@ -0,0 +1,69 @@
...
+        </button>
+
+        <teleport to="body">
+            <div v-if="showKeyboardShortcutsModal">
+                <div
+                    class="modal in settings-tab__shortcuts-modal"
+                    tabindex="-1"
+                    role="dialog"
+                    @click.self="closeKeyboardShortcutsModal"
+                >
```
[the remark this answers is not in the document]
yannik131: I added enable/disable to the Keyboard class and called them here to disable shortcuts when the modal is open

mzur:
Is this ready for review again?

resources/assets/js/annotations/components/screenshotButton.vue
```diff
@@ -5,7 +5,7 @@
     @click="capture"
     >
     <span class="fa fa-camera" aria-hidden="true"></span>
-    Capture screenshot
+    Screenshot
```
mzur: I prefer "Capture screenshot". You could display both buttons as `btn-block` and set the shortcuts button text to "Available shortcuts".

resources/assets/js/annotations/components/shortcutsButton.vue
mzur, line 1: Maybe also put this in `annotations/components/` rather than `core/`.

resources/assets/js/core/keyboard.js
mzur, line 153: Use `this.enabled` to only skip `handleKeyEvents` rather than skipping the recording of pressed keys, too. Otherwise you could e.g. disable the keyboard events, press <kbd>a</kbd>, enable the keyboard events, press <kbd>b</kbd> and an `a+b` listener would not fire because <kbd>a</kbd> was not recorded.

There will be no need to clear pressed keys or skip keyUp events if disabled, then.

resources/assets/sass/annotations/sidebar/_settingsTab.scss
mzur, line 1: Also move this to `annotations/sidebar/`.

resources/assets/sass/components/_settingsTab.scss
```diff
@@ -0,0 +1,19 @@
...
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    gap: 1em;
+}
+    
+.settings-tab__shortcuts-modal {
+    display: block;
+    overflow-y: auto;
+}
```
mzur: This class is never used.


```suggestion
```

mzur (changes requested):

resources/assets/js/core/keyboard.js
yannik131: The AI also noticed that `maybeRemoveModifierKeys` falsely deletes the modifier from the `pressedKeysSet` twice instead of once from `pressedKeysSet` and once form `pressedCodesSet`, I'll fix that as well.

mzur: Praise the AI :smile:

resources/assets/sass/videos/main.scss
```diff
@@ -56,4 +56,5 @@
 @import 'components/annotationTracks';
 @import 'components/annotationTrack';
 @import 'components/thumbnail';
+@import '../annotations/sidebar/settingsTab';
```
mzur: This must be removed. The annotation tools share the same CSS.


```suggestion
```

mzur (commented):
Looks good. Only one unnecessary import in the CSS which I'll remove myself.

mzur (approved):

PR #1418, opened 2026-04-08
Fine-grained error messages for annotation policy

### LINKED ISSUE #924: Improve error message for unauthorized label of new annotation
https://github.com/biigle/core/blob/5f6861a69eef53fa1e22b77be8e0d2c1e2231ccf/app/Policies/AnnotationPolicy.php#L124-L128

The policy checks if a label intended for a new annotation belongs to a label tree that is attached to the project(s) of the file. If this is not the case, the user now just receives a cryptic "unauthorized" response. This should be improved with a more helpful error message. Use [`Response::deny()`](https://laravel.com/docs/11.x/authorization#policy-responses) for this with the message: "You are not authorized to use label xxx because the label tree is not attached to the project."

Maybe we should look through the other policies if there are other places where we could improve error messages.

yannik131:
closes #924

### DIFF
--- a/app/Policies/AnnotationPolicy.php
+++ b/app/Policies/AnnotationPolicy.php
@@ -10,6 +10,7 @@ use Biigle\User;
 use Biigle\Volume;
 use DB;
 use Illuminate\Auth\Access\HandlesAuthorization;
+use Illuminate\Auth\Access\Response;
 
 class AnnotationPolicy extends CachedPolicy
 {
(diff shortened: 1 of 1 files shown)


app/Policies/AnnotationPolicy.php
```diff
@@ -118,14 +119,19 @@ public function attachLabel(User $user, Annotation $annotation, Label $label)
                     Role::adminId(),
                 ])
                 ->pluck('project_id');
+                
+            if ($projectIds->isEmpty()) {
+                return Response::deny("You are not authorized to use the label '{$label->name}' because you are not the editor, expert or admin of any projects this annotation belongs to.");
```
mzur:
```suggestion
                return Response::deny("Only project editors, experts or admins may attach a label to this annotation.");
```

app/Policies/AnnotationPolicy.php
```diff
@@ -118,14 +119,19 @@ public function attachLabel(User $user, Annotation $annotation, Label $label)
...
-            return $projectIds->isNotEmpty()
-                // Label must belong to a label tree that is used by one of the projects.
-                && DB::table('label_tree_project')
-                    ->whereIn('project_id', $projectIds)
-                    ->where('label_tree_id', $label->label_tree_id)
-                    ->exists();
+            if (!DB::table('label_tree_project')
+                ->whereIn('project_id', $projectIds)
+                ->where('label_tree_id', $label->label_tree_id)
+                ->exists()) {
```
mzur: I prefer to store the query result in a separate variable so we don't have a multi-line block in the if:


```suggestion
            $labelBelongsToProject = DB::table('label_tree_project')
                ->whereIn('project_id', $projectIds)
                ->where('label_tree_id', $label->label_tree_id)
                ->exists();
            if (!$labelBelongsToProject) {
```

mzur (commented):
Only small comments again which I can apply myself.

mzur (approved):

PR #1419, opened 2026-04-08
Order-agnostic result assertion

yannik131:
The AnnotationReportGenerator returns a list of all labels attached to an annotation within a given session. The order of the rows is undefined (no ORDER BY), so the test can't rely on it.

### DIFF
--- a/tests/php/Services/Reports/Volumes/ImageAnnotations/AnnotationReportGeneratorTest.php
+++ b/tests/php/Services/Reports/Volumes/ImageAnnotations/AnnotationReportGeneratorTest.php
@@ -575,7 +575,7 @@ class AnnotationReportGeneratorTest extends TestCase
             'annotation_id' => ImageAnnotationTest::create([
                 'image_id' => $image->id,
                 'created_at' => '2016-10-05 09:15:00',
-            ]),
+            ])->id,
             'user_id' => $userId,
         ]);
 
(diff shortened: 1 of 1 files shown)


tests/php/Services/Reports/Volumes/ImageAnnotations/AnnotationReportGeneratorTest.php
```diff
@@ -571,13 +571,11 @@ public function testAnnotationSessionSeparateLabelTree()
...
-            'image_id' => $image->id,
-            'created_at' => '2016-10-05 09:15:00',
-        ]);
-
         $al1 = ImageAnnotationLabelTest::create([
-            'annotation_id' => $a->id,
+            'annotation_id' => ImageAnnotationTest::create([
+                'image_id' => $image->id,
+                'created_at' => '2016-10-05 09:15:00',
+            ]),
```
mzur: I'm surprised that this worked, since this expects an ID. Maybe there is some magic going on under the hood.


```suggestion
            'annotation_id' => ImageAnnotationTest::create([
                'image_id' => $image->id,
                'created_at' => '2016-10-05 09:15:00',
            ])->id,
```

mzur (commented):
I didn't know about `assertEqualsCanonicalizing`. Seems like a good solution for these sometimes failing tests! I only have one suggestion for a small change, which I can apply myself.

mzur (approved):

Issue #992, opened 2024-11-28
Improve metadata file validation


mzur:
The current volume metadata file validation is performed in two steps. First, the format-specific validation is performed in `recognizesFile()`. Second, the validation of the parsed metadata is performed.

The first validation step is not implemented very well because a user uploading a malformed file only gets a message "file is not valid". We should implement a mechanism for metadata parsers to return more meaningful error messages for the validation (e.g. "category x is missing the name attribute"). The `recognizesFile()` method can be replaced with a `validate()` method that may return a special `MetadataValidationException` with a message that is passed on to the frontend. Any other exception is handled like an unexpected server error.

mzur:
The validation is performed here:
- https://github.com/biigle/core/blob/929fcb62ed33f51f1a6ff284f05ee594927955b5/app/Http/Requests/StorePendingVolume.php#L85
- https://github.com/biigle/core/blob/929fcb62ed33f51f1a6ff284f05ee594927955b5/app/Http/Requests/StoreVolume.php#L119
- https://github.com/biigle/core/blob/929fcb62ed33f51f1a6ff284f05ee594927955b5/app/Http/Requests/StoreVolumeMetadata.php#L79

The parsers implement it here:
- https://github.com/biigle/core/blob/929fcb62ed33f51f1a6ff284f05ee594927955b5/app/Services/MetadataParsing/MetadataParser.php#L34
- https://github.com/biigle/core/blob/929fcb62ed33f51f1a6ff284f05ee594927955b5/app/Services/MetadataParsing/CsvParser.php#L52
- https://github.com/biigle/metadata-ifdo/blob/da0ab36ddce12c402d9db0c9873c087c9a3b360d/src/IfdoParser.php#L43
- https://github.com/biigle/metadata-coco/blob/4b7da60a2e00acc8f88fd3360b8e1f7c0a697243/src/CocoParser.php#L48

PR #1428, opened 2026-04-14
Improve point validation message

### LINKED ISSUE #1067: Improve annotation point validation message
If there is a [validation error](https://github.com/biigle/core/blob/f151e5dd42dc85272795f8eac5dfed335546c0fd/app/Traits/HasPointsAttribute.php#L27-L50) for annotation points, the message is "Invalid (number of) points for shape ...". Update the validation so it can return the messages:

- Too few points for shape ...
- Too many points for shape ...
- Invalid points for shape ...

References https://github.com/orgs/biigle/discussions/1066

yannik131:
Closes #1067

### DIFF
--- a/app/Services/MetadataParsing/Annotation.php
+++ b/app/Services/MetadataParsing/Annotation.php
@@ -66,7 +66,7 @@ abstract class Annotation
     }
 
     /**
-     * {@inheritdoc}
+     * Sets the points and rounds them if they are numeric.
      */
     protected function setPointsAttribute(array $points)
     {
--- a/app/Services/MetadataParsing/Annotation.php
+++ b/app/Services/MetadataParsing/Annotation.php
@@ -27,6 +27,7 @@ abstract class Annotation
         public array $labels,
     ) {
         $this->shape_id = $shape->id;
+        $this->validatePoints($points); // TODO Without this, non-numeric strings would be passed directly to round()
         $this->setPointsAttribute($points);
 
         array_walk($labels, function ($label) {
@@ -51,7 +52,7 @@ abstract class Annotation
     }
 
     /**
-     * Validatethe points and labels.
+     * Validate the points and labels.
      *
      * @throws Exception If something is invalid.
      */
@@ -67,7 +68,7 @@ abstract class Annotation
     /**
      * {@inheritdoc}
      */
-    public function setPointsAttribute(array $points)
+    private function setPointsAttribute(array $points)
     {
         $this->points = array_map(
             fn ($a) => is_array($a)
--- a/app/Services/MetadataParsing/Annotation.php
+++ b/app/Services/MetadataParsing/Annotation.php
@@ -66,7 +66,7 @@ abstract class Annotation
     }
 
     /**
-     * {@inheritDoc}
+     * {@inheritdoc}
      */
     protected function setPointsAttribute(array $points)
     {
--- a/app/Services/MetadataParsing/Annotation.php
+++ b/app/Services/MetadataParsing/Annotation.php
@@ -4,6 +4,7 @@ namespace Biigle\Services\MetadataParsing;
 
 use Biigle\Shape;
 use Biigle\Traits\HasPointsAttribute;
+use Biigle\Traits\InvalidCoordinateTypeException;
 use Exception;
 
 // Abstract becaus it should not be used directly.
@@ -27,7 +28,6 @@ abstract class Annotation
         public array $labels,
     ) {
         $this->shape_id = $shape->id;
-        $this->validatePoints($points); // TODO Without this, non-numeric strings would be passed directly to round()
         $this->setPointsAttribute($points);
 
         array_walk($labels, function ($label) {
@@ -72,9 +72,18 @@ abstract class Annotation
     {
         $this->points = array_map(
             fn ($a) => is_array($a)
-                ? array_map(fn ($b) => round($b, 2), $a)
-                : round($a, 2),
+                ? array_map(fn ($b) => $this->roundValue($b), $a)
+                : $this->roundValue($a),
             $points
         );
+    }
+    
+    private function roundValue($value) 
+    {
+        if (!is_int($value) && !is_float($value)) {
+            throw new InvalidCoordinateTypeException('Coordinate "'.$value.'" is not of type int or float.');
+        }
+        
+        return round($value, 2);
     }
 }
--- a/app/Services/MetadataParsing/Annotation.php
+++ b/app/Services/MetadataParsing/Annotation.php
@@ -66,7 +66,7 @@ abstract class Annotation
     }
 
     /**
-     * @inheritDoc
+     * {@inheritDoc}
      */
     protected function setPointsAttribute(array $points)
     {
--- a/app/Services/MetadataParsing/Annotation.php
+++ b/app/Services/MetadataParsing/Annotation.php
@@ -78,7 +78,7 @@ abstract class Annotation
         );
     }
     
-    private function roundValue($value) 
+    private function roundValue($value)
     {
         if (!is_int($value) && !is_float($value)) {
             throw new InvalidCoordinateTypeException('Coordinate "'.$value.'" is not of type int or float.');
--- a/app/Services/MetadataParsing/Annotation.php
+++ b/app/Services/MetadataParsing/Annotation.php
@@ -68,22 +68,13 @@ abstract class Annotation
     /**
      * {@inheritdoc}
      */
-    private function setPointsAttribute(array $points)
+    protected function setPointsAttribute(array $points)
     {
         $this->points = array_map(
             fn ($a) => is_array($a)
-                ? array_map(fn ($b) => $this->roundValue($b), $a)
-                : $this->roundValue($a),
+                ? array_map(fn ($b) => is_numeric($b) ? round($b, 2) : $b, $a)
+                : (is_numeric($a) ? round($a, 2) : $a),
             $points
         );
-    }
-    
-    private function roundValue($value)
-    {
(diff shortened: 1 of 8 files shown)


app/Traits/HasPointsAttribute.php
```diff
@@ -5,6 +5,19 @@
...
+class InvalidNumberOfCoordinatesException extends Exception
+{
+};
+class InvalidNumberOfPointsException extends Exception
+{
+};
+class InvalidShapeException extends Exception
+{
+};
+
```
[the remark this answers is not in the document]
yannik131: - The comment does not match the code suggestion
- The code is bad: Aliases don't create new types, so tests cannot distinguish between `InvalidNumberOfCoordinatesException` and `InvalidNumberOfPointsException` if they are aliases of the same underlying exception type, this example here will output "Caught A" twice, which is not what we want

```php
<?php

// Two “different” exception names via aliasing
use LengthException as A;
use LengthException as B;

try {
	throw new A();
} catch(A $a) {
	echo "Caught A\n";
} catch(B $b) {
	echo "Caught B\n";
}

try {
	throw new B();
} catch(A $a) {
	echo "Caught A\n";
} catch(B $b) {
	echo "Caught B\n";
}
```

@mzur I'm not sure if we have a source file where we collect exception types, should these exception types stay local to this trait or be put somewhere else?

tests/php/Traits/HasPointsAttributeTest.php
```diff
@@ -0,0 +1,108 @@
...
+    }
+    
+    public static function invalidPointsProvider()
+    {
+        return [
+            'point' => ['point', [0, 0, 1, 1]],
+            'rectangle' => ['rectangle', [0, 0, 1, 1, 2, 2]],
+            'ellipse' => ['ellipse', [0, 0, 1, 1, 2, 2, 3, 3, 4, 4]],
+            'polygon' => ['polygon', [0, 0, 1, 1, 0, 0]],
+            'line' => ['line', [0, 0]]
```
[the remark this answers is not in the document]
yannik131: I specifically created custom exception types to avoid testing against exact strings because that's brittle.

app/Traits/HasPointsAttribute.php
```diff
@@ -15,45 +28,102 @@ trait HasPointsAttribute
...
+                if ($pointCount !== 1) {
+                    throw new InvalidNumberOfPointsException('Invalid number of points: Need exactly 1 point for shape point, but '.$pointCount.' were given.');
+                }
                 break;
+            case Shape::circleId(): return;
+            case Shape::rectangleId():
+            case Shape::ellipseId():
+                if ($pointCount !== 4) {
+                    throw new InvalidNumberOfPointsException('Invalid number of points: Expected 4, got '.$pointCount.'');
+                }
```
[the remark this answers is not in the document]
yannik131: I think users can infer too few/many by comparing 2 numbers.

app/Traits/HasPointsAttribute.php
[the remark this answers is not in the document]
mzur: All classes must live in their own file and the file path and name must match the namespace. If you create new exception classes, put them into `app/Exceptions/`.

yannik131:
@mzur In my backend code, I have so far not used type declarations, as they seem to be inconsistently used across the codebase. I feel like I made a mistake and should use them wherever possible. What do you think?

mzur:
As we just discussed, you can add them whenever you modify old code that doesn't have them yet.

mzur:
btw is this ready for review again? You added another commit after requesting my review.

yannik131:
@mzur Yes it's ready

app/Services/MetadataParsing/Annotation.php
```diff
@@ -67,13 +68,22 @@ public function validate(): void
...
     }
+    
+    private function roundValue($value)
+    {
+        if (!is_int($value) && !is_float($value)) {
+            throw new InvalidCoordinateTypeException('Coordinate "'.$value.'" is not of type int or float.');
+        }
+        
+        return round($value, 2);
+    }
```
mzur: The class already uses `validatePoints` so no additional validation should be necessary.

app/Traits/HasPointsAttribute.php
```diff
@@ -15,45 +19,105 @@ trait HasPointsAttribute
...
+    {
+        $invalidCoordinates = array_values(array_filter(
+            $points,
+            fn ($coordinate) => !is_int($coordinate) && !is_float($coordinate)
+        ));
 
-        if (!$valid) {
-            throw new Exception('Point coordinates must be of type float or integer.');
+        if (!empty($invalidCoordinates)) {
+            throw new InvalidCoordinateTypeException('Invalid coordinate type: The following coordinates aren\'t int or float: '.implode(', ', $invalidCoordinates).'');
```
mzur: I'm not sure if it's practical to list the offending coordinates here. I'd expect that either all coordinates are correct or all are incorrect and it's not helpful to add a potentially very large list of coordinates to the error message.

app/Traits/HasPointsAttribute.php
```diff
@@ -15,45 +19,105 @@ trait HasPointsAttribute
...
+        
+        switch ($this->shape_id) {
+            case Shape::circleId():
+                if ($size !== 3) {
+                    throw new InvalidNumberOfCoordinatesException('Invalid number of values for shape circle: Expected 3, got '.$size.'.');
+                }
+                break;
+            default:
+                if ($size % 2 !== 0) {
+                    throw new InvalidNumberOfCoordinatesException('Even number of coordinates expected but got '.$size.' coordinates instead. Note that each pair of coordinates is interpreted as a point.');
```
mzur:
```suggestion
                    throw new InvalidNumberOfCoordinatesException('Even number of coordinates expected but got '.$size.' coordinates instead.');
```

app/Traits/HasPointsAttribute.php
```diff
@@ -15,45 +19,105 @@ trait HasPointsAttribute
...
+                }
+                
+                if ($size == 0) {
```
mzur: Prefer strict equality checks.


```suggestion
                if ($size === 0) {
```

app/Traits/HasPointsAttribute.php
```diff
@@ -15,45 +19,105 @@ trait HasPointsAttribute
...
+        $pointCount = intval(count($points) / 2);
+        
         switch ($this->shape_id) {
             case Shape::pointId():
-                $valid = $size === 2;
+                if ($pointCount !== 1) {
+                    throw new InvalidNumberOfPointsException('Invalid number of points for shape point: Need exactly 1 point, but '.$pointCount.' were given.');
+                }
+                break;
+            case Shape::circleId(): return;
```
mzur:
```suggestion
```

tests/php/Traits/HasPointsAttributeTest.php
mzur, line 1: Better call this `Traits\HasPointsAttributeTest` so the connection between the tests and the application logic is immediately clear.

mzur (changes requested):
Please add a docblock to each new method you create.

Please prefer `protected` methods over `private` methods, as these can be overridden (e.g. in test stubs).

app/Services/MetadataParsing/Annotation.php
yannik131: If  the points are initialized 
- using the constructor (frequently done in tests)
- or by creating an empty annotation and then setting them without calling `validatePoints` first (which is never enforced)

then `round` will give a slightly cryptic `TypeError` instead of the `InvalidCoordinateTypeException` if passed something non-numeric.

yannik131:
@mzur 
> Please add a docblock to each new method you create.
> 
> Please prefer `protected` methods over `private` methods, as these can be overridden (e.g. in test stubs).

Docblocks:
- Should the docblock be more of a one-liner that describes what the function does or should it always cover every aspect like arguments and their types, return values and their types, exceptions thrown, what the function does etc..
- In my experience, docblocks on *every* function duplicate information, quickly lead to stale comments and maintance work and don't add any value if the function is kept short and uses descriptive naming

Protected:
I would prefer `private` as a default unless the method is an intended extension point in subclasses (which `roundValue` isn't I think). I found [this discussion](https://www.reddit.com/r/PHP/comments/3jdbkd/protected_or_private_as_a_default_visibility_and/) interesting.

That said, if defaulting to `protected` and using `docblocks` everywhere are project conventions, I’m happy to follow them. These could also be written down in some document for future references.

app/Services/MetadataParsing/Annotation.php
mzur: A valid point. For now I'd like to stick to the existing validation pattern where all annotations are validated as part of `StorePendingVolumeImport`. If they are always validated when instantiated, validation errors could happen at unexpected times.

To avoid TypeErrors we could do something like this for now:
```php
$this->points = array_map(
    fn ($a) => is_array($a)
        ? array_map(fn ($b) => is_numeric($b) ? round($b, 2) : $b, $a)
        : (is_numeric($a) ? round($a, 2) : $a),
    $points
);
```

Then as part of https://github.com/biigle/core/issues/992 we can improve the overall validation of metadata imports.

mzur:
Both are project conventions at the moment. While you have valid points (again) I'd like to stick to them for now to keep the codebase consistent.

Docblocks should include a one-liner explaining the method and any type annotations that may be missing from the actual code (e.g. there are lots of methods with only types in the docblocks because PHP didn't have native annotations back then).

yannik131:
I assume neither convention applies to tests, the other tests don't seem to follow them as far as I can tell, so I didn't change that.

mzur (approved):

PR #1429, opened 2026-04-15
fix: validate file accessibility when updating volume URL

### LINKED ISSUE #829: Fix file validation when volume URL is updated
If the URL of an existing volume is [updated](https://github.com/biigle/core/blob/a80062ea6ce790023e40257d237e612b0d243fd8/app/Http/Controllers/Api/VolumeController.php#L132) the validation should also check for existing volume files (similar to the check for a new volume). If a sample of the volume files cannot be found with the new URL, the change should be rejected.

mvanhorn:
## Summary

Closes #829. Volume creation runs the `VolumeFiles` rule, which probes a sample of the supplied filenames at the supplied URL and rejects the request if any sample file isn't reachable. When an existing volume's URL is updated through `VolumeController::update`, no equivalent check ran — a typo or mistaken disk reference would silently re-point the volume at empty storage, leaving the volume's existing image/video records pointing at files that no longer exist.

## Changes

- **`app/Rules/VolumeAccessibleAtUrl.php`** (new): rule that takes a `Volume` and probes a configurable sample (default 5) of its existing files via `FileCache::exists` against a candidate URL. Mirrors the URL-normalisation in `VolumeFiles` (strip trailing slash, leave bare scheme alone).
- **`app/Http/Requests/UpdateVolume.php`**: when the request changes the URL, the new rule is appended to the `url` validation chain after `VolumeUrl`. Under `bail` so it only runs once the URL passes shape and allowlist checks.
- **`tests/php/Rules/VolumeAccessibleAtUrlTest.php`** (new): 5 cases covering pass/fail/empty-volume/trailing-slash/non-string-input.

## Edge cases handled

- **Volume with no files yet**: passes silently. The subsequent `ProcessNewVolumeFiles` job will populate from the new URL and surface per-file problems then.
- **URL unchanged**: the new rule isn't appended, so name-only / handle-only updates and the existing no-op-update path are unaffected.
- **Trailing slash on the URL**: stripped before probing, matching the normalisation in `VolumeFiles::__construct`.
- **Non-string `url` input**: rule returns false (validation fails cleanly).

## Test plan

- [x] PHP syntax (manual review — no `php` binary on this machine to run `php -l`).
- [x] Test factories (`VolumeTest::create`, `ImageTest::create`) verified to exist under `Biigle\Tests`.
- [ ] `php artisan test --filter VolumeAccessibleAtUrl` — would like to run locally but the test harness needs the project's Postgres + Docker setup. Happy to iterate if maintainer CI surfaces anything.
- [x] Existing `testUpdateUrl` and `testUpdateInvalidUrl` paths in `VolumeControllerTest` are unaffected because they don't change the URL to a path where the volume's existing files would be missing — the new rule only fires when the volume actually has registered files.

Closes #829

### DIFF
--- a/app/Http/Requests/UpdateVolume.php
+++ b/app/Http/Requests/UpdateVolume.php
@@ -3,6 +3,7 @@
 namespace Biigle\Http\Requests;
 
 use Biigle\Rules\Handle;
+use Biigle\Rules\VolumeAccessibleAtUrl;
 use Biigle\Rules\VolumeUrl;
 use Biigle\Volume;
 use Illuminate\Foundation\Http\FormRequest;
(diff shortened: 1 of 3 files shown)


mzur:
Hi, thanks for the contribution. Another user/AI already proposed a simpler fix for the same issue in #1430, so I'm closing this a duplicate.

PR #1430, opened 2026-04-15
Validate sample volume files when updating the URL

### LINKED ISSUE #829: Fix file validation when volume URL is updated

tmchow:
## Summary

Issue #829: when a volume's URL is updated, the store path validates that a sample of the submitted files exist at the URL via `VolumeFiles`, but the update path does not — so pointing an existing volume at a URL that does not contain its files is accepted, and the downstream `ProcessNewVolumeFiles` job silently fails when it cannot locate the files.

This mirrors the store-side sample check on the update path:

- `UpdateVolume::withValidator` runs the `VolumeFiles` rule against the existing volume's filenames with the new URL
- Only fires when the URL is actually changing and the volume already has files attached (preserves the "create-first, add files later" flow)
- Runs *after* the other rules so `filled|max:256|VolumeUrl` errors surface first

## Testing

- Added `VolumeControllerTest::testUpdateUrlFilesMissing`:
  - Sets up `file.jpg` on `admin-test://volumes` (the volume's current URL)
  - Attaches `file.jpg` to the volume via `ImageTest::create`
  - Submits a URL update to `editor-test://volumes` (where `file.jpg` does not exist)
  - Asserts 422 and that the URL on the volume is unchanged
  - Asserts no `ProcessNewVolumeFiles` is queued

- The existing `testUpdateUrl` and `testUpdateInvalidUrl` still pass — those use a volume with no attached files, so the new early-return on empty filenames preserves their behaviour.

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)

### DIFF
--- a/app/Http/Requests/UpdateVolume.php
+++ b/app/Http/Requests/UpdateVolume.php
@@ -63,7 +63,12 @@ class UpdateVolume extends FormRequest
                 return;
             }
 
-            $filenames = $this->volume->files()->pluck('filename')->all();
+            $filenames = $this->volume
+                ->files()
+                ->inRandomOrder()
+                ->limit(5)
+                ->pluck('filename')
+                ->all();
             if (empty($filenames)) {
                 return;
             }
--- a/app/Http/Requests/UpdateVolume.php
+++ b/app/Http/Requests/UpdateVolume.php
@@ -3,6 +3,7 @@
 namespace Biigle\Http\Requests;
 
 use Biigle\Rules\Handle;
+use Biigle\Rules\VolumeFiles;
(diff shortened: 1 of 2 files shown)


mzur (approved):
Thanks!

PR #1439, opened 2026-04-17
Fix HTML encoding of imported filenames

### LINKED ISSUE #1438: CSV metadata import HTML-encodes & to &amp; in filenames
## Issue
When creating a volume with a CSV metadata file containing filenames with `&` in them (e.g. URLs like `images?cruise_id=X&lowering_id=Y&image_path=Z.jpg)`, the `&` is stored as `&amp;` in the database. The HTML-encoded filename populates to the Volume files form and breaks the import URLs.

Image

## Cause
In resources/views/volumes/create/step2.blade.php, line 11, the filenames extracted from the metadata CSV are injected into a JavaScript variable using Blade's {{ }} syntax, which HTML-encodes the output:

`biigle.$declare('volumes.filenames', '{{ $filenames }}')`

## Possible Fix
Use Js::from() instead, consistent with how some other variables are declared on the same page:

`biigle.$declare('volumes.filenames', {!! Js::from($filenames) !!})`

This handles `&` as well as other characters like single quotes and backslashes if users have complex filenames.

It also appears that it re-encodes the filename and the bug repeats itself on subsequent attempts to create the volume if there is a validation error, leading to a humorous `&amp;amp;amp;...` 

Image

Happy to open a PR if desired.

ThomasMorrow-SOI:
In resources/views/volumes/create/step2.blade.php, filenames extracted from the metadata CSV were injected into a JavaScript variable using Blade's {{ }} syntax, which HTML-encoded the output.

HTML encoding broke some of the filenames by converting characters such as '&' to '&amp;'.

Replaced `'{{ $filenames }}'` with `Js::from($filenames)` which properly encodes the string and preserves special characters.

Addresses #1438

### DIFF
--- a/resources/views/volumes/create/step2.blade.php
+++ b/resources/views/volumes/create/step2.blade.php
@@ -8,7 +8,7 @@
         biigle.$declare('volumes.url', '{!! $oldUrl !!}');
         biigle.$declare('volumes.handle', {{ Js::from($oldHandle) }});
         biigle.$declare('volumes.mediaType', '{!! $mediaType !!}');
-        biigle.$declare('volumes.filenames', '{{ $filenames }}');
+        biigle.$declare('volumes.filenames', {{ Js::from($filenames) }});
         biigle.$declare('volumes.filenamesFromMeta', {{ $filenamesFromMeta ? 'true' : 'false' }});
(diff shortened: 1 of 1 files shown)


mzur (approved):
Thanks!

PR #1453, opened 2026-04-29
Change the temporary annotation color to fit match the selected label

### LINKED ISSUE #1359: Change the temporary annotation color to fit match the selected label
- A common error is that a different label is activated than the one you intended to use
- when creating e.g. a box annotation the color of the temporary dashed annotation is always blue
- if it would match the label color the user would easier recognize that the wrong label is activated

yannik131:
Closes #1359

### DIFF
--- a/resources/assets/js/annotations/components/annotationCanvas/drawInteractions.vue
+++ b/resources/assets/js/annotations/components/annotationCanvas/drawInteractions.vue
@@ -166,12 +166,34 @@ export default {
 
             return never;
         },
+        updateDraftAnnotationColor(label) {
+            if (!drawInteraction) {
+                return;
+            }
+
+            const overlay = drawInteraction.getOverlay();
+            const source = overlay && overlay.getSource();
+            const features = source ? source.getFeatures() : [];
+
+            features.forEach((feature) => {
+                if (label && label.color && this.draftAnnotationUsesLabelColor) {
+                    feature.set('color', label.color);
+                } else {
+                    feature.unset('color');
+                }
+            });
+        },
     },
     watch: {
         selectedLabel(label) {
             if (this.isDrawing && !label && !this.labelbotIsActive) {
                 this.resetInteractionMode();
             }
+            
+            this.updateDraftAnnotationColor(label);
+        },
+        draftAnnotationUsesLabelColor() {
+            this.updateDraftAnnotationColor(this.selectedLabel);
         },
         interactionMode(mode) {
             this.maybeUpdateDrawInteractionMode(mode)
--- a/resources/assets/js/annotations/components/annotationCanvas/drawInteractions.vue
+++ b/resources/assets/js/annotations/components/annotationCanvas/drawInteractions.vue
@@ -110,8 +110,12 @@ export default {
                 });
                 this.map.addInteraction(drawInteraction);
 
-                drawInteraction.on('drawstart', () => {
+                drawInteraction.on('drawstart', (event) => {
                     this.drawEnded = false;
+                    
+                    if (this.draftAnnotationUsesLabelColor && this.selectedLabel) {
+                        event.feature.set('color', this.selectedLabel.color);
+                    }
                 });
 
                 drawInteraction.on('drawend', (e) => {
--- a/resources/assets/js/annotations/components/annotationCanvas/drawInteractions.vue
+++ b/resources/assets/js/annotations/components/annotationCanvas/drawInteractions.vue
@@ -112,7 +112,7 @@ export default {
 
                 drawInteraction.on('drawstart', (event) => {
                     this.drawEnded = false;
-                    
+
                     if (this.draftAnnotationUsesLabelColor && this.selectedLabel) {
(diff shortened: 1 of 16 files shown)


yannik131:
A new `draftAnnotationUsesLabelColor` variable was introduced. This is the data flow for the image annotation tool:

Using it to change the color:
- Defined in data() of `annotatorContainer.vue`
- Passed to `annotationCanvas.vue` from `annotatorContainer.vue` as a prop
- The prop is used in the `annotationCanvas.vue`-mixin `drawInteractions.vue` to actually set the color of the draft annotation to the one of the `selectedLabel`

Managing it with a toggle:
- Defined also in `data()` of `settingsTab.vue` 
- Passed to the `active` prop of `powerToggle.vue` to show the current value in `settings.blade.php`
- The `<power-toggle/>` changes the value using callbacks defined in `settingsTab.vue`
- The watcher `settingsTab.vue:draftAnnotationUsesLabelColor` emits the settings change
- `annotatorContainer.vue:handleSettingsChange` receives the change, keeping `drawInteractions.vue` up to date

Restoring it at startup:
- Set as a key in `data()::restoreKeys` of `settingsTab.vue`
- The watcher `settingsTab.vue:draftAnnotationUsesLabelColor` also sets the value in settings
- The value is reloaded in `settingsTab.vue:created()` at mounting

yannik131:
This is the data flow for the video annotation tool:

Using it to change color:
- Defined in `data():settings` of `videoContainer.vue`
- Passed to `videoScreen.vue` from `videoContainer.vue` as a prop
- Used in the `videoScreen.vue`-mixin `videoScreen/drawInteractions.vue` to actually set the color of the draft annotation

Restoring it at startup:
- Defined in `defaults` of `videos/stores/settings.js`
  - Passed to the constructor of `Settings`
  - Used as a fallback when loading the variable from browser storage
- `videos/components/settingsTab.vue` sets the value from `restoreKeys` at `created()`
- This triggers the watcher at `videos/components/settingsTab.vue` to emit the settings signal
- Set in `videoContainer.vue:data():settings` from `videoContainer.vue:handleUpdatedSettings` 

Managing it with a toggle:
Identical to image annotation, but now changes are in `videos/components/settingsTab.vue` and `videos/show/sidebar-settings.blade.php`. This is not DRY! Changes are almost identical to the image annotation tool, just the watcher is slightly different. 

@mzur The changes to the `videos/components/settingsTab.vue` and `annotations/components/settingsTab.vue` are almost identical and should probably be factored out into a mixin. The mixin would have to infer somehow if it's used in the video or image annotation settings tab.

resources/assets/js/annotations/components/annotationCanvas/drawInteractions.vue
```diff
@@ -110,8 +110,12 @@ export default {
                 });
                 this.map.addInteraction(drawInteraction);
 
-                drawInteraction.on('drawstart', () => {
+                drawInteraction.on('drawstart', (event) => {
                     this.drawEnded = false;
+
```
mzur: Our linter is currently not complaining about this but please avoid whitespaces in empty lines.

```suggestion

```

resources/views/annotations/show/tabs/settings.blade.php
```diff
@@ -84,6 +84,18 @@
                     </power-toggle>
             </div>
         @endcan
+
```
mzur:
```suggestion

```

resources/views/videos/show/sidebar-settings.blade.php
```diff
@@ -89,6 +89,18 @@
     <div class="sidebar-tab__section">
         <power-toggle :active="singleAnnotation" title-off="Enable always creating single-frame annotations" title-on="Disable always creating single-frame annotations" v-on:on="handleSingleAnnotation" v-on:off="handleDisableSingleAnnotation">Single-Frame Annotation</power-toggle>
     </div>
+
```
mzur:
```suggestion

```

mzur (commented):
It's possible to change the selected label while the annotation is still a draft. The draft annotation color should update in this case.

Also it's unlikely but possible to enable/disable the settings option while the annotation is a draft. Should we update the color here, too?

Finally, please update the respective manual articles about the sidebar with a description of the new button.

yannik131:
@mzur Disabling/Enabling this in the settings now updates the color immediately as well. 
Also, I've noticed that there are URLParams for the image and video annotation tools, but not all settings are reflected there. Should this new option be included there (and in the manual entries) as well?

mzur (approved):
Looks good, thanks 👍 

All options are automatically available as URL params. But I don't think we have to add everything to the manual. The documented params are mostly for users who want to send links so the receiver sees exactly their viewport or they use the same annotation mode.

I noticed an error here, though. If I set `draftAnnotationUsesLabelColor=false` in the URL, the `false` is set as a string, triggering Vue type warnings. Maybe we should update the URL param parsing to handle `true` and `false` correctly. But this can be done in a separate issue.

PR #1465, opened 2026-05-20
Console error volare

### LINKED ISSUE #1121: Measurement tool throws console error after using volume label review tool
The console error is thrown after these actions:

1. Open an image from a volume that contains multiple images
2. Use the measurement tool and fixate the line by double click
3. Add the measurement line as line annotation (by selecting a label and hitting enter)
4. Click on the volume label review tool button
5. Use the left or right arrow key

yannik131:
Closes #1121

### DIFF
--- a/resources/assets/js/annotations/components/annotationCanvas/measureInteraction.vue
+++ b/resources/assets/js/annotations/components/annotationCanvas/measureInteraction.vue
@@ -92,7 +92,7 @@ export default {
             }
         },
         image() {
-            if (this.isMeasuring && this.image) {
+            if (this.isMeasuring && this.image && this.measureFeature) {
                 // Wait for the new image to be propagated down to the measureTooltip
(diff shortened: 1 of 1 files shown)


yannik131:
@mzur The issue was this:
- When a measurement is started with left mouse click, the total length is displayed in a tooltip (the `measureFeature`)
- When jumping to the next image (regardless if in volare mode or not), to preserve the tooltip, `updateMeasureFeature` with the measure feature [is called](https://github.com/biigle/core/blob/1121-console-error-volare/resources/assets/js/annotations/components/annotationCanvas/measureInteraction.vue#L95)
- If no measure feature exists because measuring hasn't started yet, this will cause an error trying to access a property of undefined
- To preserve current behaviour, I just added a check if the measure feature is actually present

mzur (approved):
👍

PR #1466, opened 2026-05-20
Filenames with slashes in volume filtering

### LINKED ISSUE #1123: Filtering filenames that begin with a slash results in an error
Filenames starting with "/" (e.g. remote videos) cause an error when using the filename filter in the volume overview.

yannik131:
Closes #1123

### DIFF
--- a/resources/assets/js/volumes/stores/filters.js
+++ b/resources/assets/js/volumes/stores/filters.js
@@ -205,7 +205,7 @@ let filenameFilter = {
     getSequence(volumeId, pattern) {
         return VolumesApi.queryFilesWithFilename({
             id: volumeId,
-            pattern: pattern,
+            pattern: encodeURIComponent(pattern),
         });
     },
 };
--- a/resources/assets/js/volumes/stores/filters.js
+++ b/resources/assets/js/volumes/stores/filters.js
@@ -205,7 +205,7 @@ let filenameFilter = {
     getSequence(volumeId, pattern) {
         return VolumesApi.queryFilesWithFilename({
             id: volumeId,
-            pattern: encodeURIComponent(pattern),
+            pattern: pattern,
         });
     },
 };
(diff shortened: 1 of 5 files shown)


resources/assets/js/volumes/stores/filters.js
```diff
@@ -205,7 +205,7 @@ let filenameFilter = {
     getSequence(volumeId, pattern) {
         return VolumesApi.queryFilesWithFilename({
             id: volumeId,
-            pattern: pattern,
+            pattern: encodeURIComponent(pattern),
         });
     },
```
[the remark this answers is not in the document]
yannik131: I tried

```js
queryFilesWithFilename: {
    method: 'GET',
    url: ({ id, pattern }) =>
        `api/v1/volumes/${id}/files/filter/filename/${encodeURIComponent(pattern)}`,
},
```

but that didn't work, so I'll keep it this way.

routes/api.php
```diff
@@ -432,7 +432,7 @@
 
     $router->get('{id}/files/filter/filename/{pattern}', [
         'uses' => 'Filters\FilenameController@index',
-    ]);
+    ])->where('pattern', '.*');
```
mzur: Why this if it matches anything?

yannik131: After additional investigation:
- Laravel already encodes/decodes URI components
- However, laravel decodes the route parameters and throws an exception if the decoded pattern contains invalid characters such as slashes
- 2 possible fixes: 
  - encode/decode manually on top of vue/laravel encoding/decoding, then laravel can't "see" the invalid characters after decoding the pattern just once
  - add the `->where('pattern', '.*');` to explicitly allow any characters, including slashes

The second approach is actually cleaner so I'll remove the manual encoding/decoding

mzur: Please add a comment here why this is used, then. And you think we shouldn't use `encodeURIComponent` in JS as well? Is seems a safer bet.

yannik131: I added the comment. I feel like additional encoding would be redundant and make testing more difficult.

mzur (approved):

PR #1467, opened 2026-05-21
Option for unit preference

### LINKED ISSUE #784: Settings option for unit preference
In the image annotation tool settings, users should be able to set a unit preference (cm, m, cm², m² etc.). If the preference is set, the lengths/areas are always displayed in the configured unit.

yannik131:
Closes #784

### DIFF
--- a/resources/assets/js/annotations/annotatorContainer.vue
+++ b/resources/assets/js/annotations/annotatorContainer.vue
@@ -594,9 +594,6 @@ export default {
                 case 'draftAnnotationUsesLabelColor':
                     this.draftAnnotationUsesLabelColor = value;
                     break;
-                case 'preferredUnit':
-                    this.preferredUnit = value;
-                    break;
             }
         },
         handleAnnotationModeChange(mode, data) {
--- a/resources/assets/js/annotations/annotatorContainer.vue
+++ b/resources/assets/js/annotations/annotatorContainer.vue
@@ -86,6 +86,7 @@ export default {
             showScaleLine: false,
             showExampleAnnotations: true,
             draftAnnotationUsesLabelColor: true,
+            preferredUnit: 'auto',
             imagesArea: null,
             openTab: null,
             userUpdatedVolareResolution: false,
--- a/resources/assets/js/annotations/annotatorContainer.vue
+++ b/resources/assets/js/annotations/annotatorContainer.vue
@@ -593,6 +593,9 @@ export default {
                 case 'draftAnnotationUsesLabelColor':
                     this.draftAnnotationUsesLabelColor = value;
                     break;
+                case 'preferredUnit':
+                    this.preferredUnit = value;
+                    break;
             }
         },
         handleAnnotationModeChange(mode, data) {
--- a/resources/assets/js/annotations/components/measureTooltip.vue
+++ b/resources/assets/js/annotations/components/measureTooltip.vue
@@ -93,7 +93,8 @@ export default {
         },
         formatMeasurement(measurement, unit, decimals) {
             return new Intl.NumberFormat("en-US", {
-                maximumSignificantDigits: decimals
+                minimumFractionDigits: decimals,
+                maximumFractionDigits: decimals
             }).format(measurement) + ' ' + unit;
         },
         unitNearest(measurement, multipliers, min) {
--- a/resources/assets/js/annotations/components/measureTooltip.vue
+++ b/resources/assets/js/annotations/components/measureTooltip.vue
@@ -97,12 +97,12 @@ export default {
             }).format(measurement) + ' ' + unit;
         },
         unitNearest(measurement, multipliers, min) {
-            if (measurement === 0) {
-                return multipliers.length - 1;
-            }
-
             if (UnitNames.indexOf(this.preferredUnit) !== -1) {
                 return UnitNames.indexOf(this.preferredUnit);
+            }
+
+            if (measurement === 0) {
+                return multipliers.length - 1;
             }
 
             min = min || 1;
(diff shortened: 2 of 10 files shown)


resources/assets/js/annotations/components/measureTooltip.vue
```diff
@@ -28,10 +30,13 @@ export default {
     },
     computed: {
         areaUnitMultipliers() {
-            return this.unitMultipliers.map(function (multiplier) {
+            return UnitMultipliers.map(function (multiplier) {
                 return Math.pow(multiplier, 2);
             });
         },
+        preferredUnit() {
+            return Settings.get('preferredUnit');
```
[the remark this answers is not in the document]
yannik131: The store is reactive and I tested that the change is immediately reflected by the UI, it's working

resources/assets/js/annotations/components/measureTooltip.vue
```diff
@@ -79,22 +84,25 @@ export default {
...
+                length = length / UnitMultipliers[index];
             }
 
-            return this.formatMeasurement(length, unit);
+            return this.formatMeasurement(length, unit, 1);
         },
         formatMeasurement(measurement, unit, decimals) {
-            decimals = Math.pow(10, decimals || 1);
-            return (Math.round(measurement * decimals) / decimals) + ' ' + unit;
+            return new Intl.NumberFormat("en-US", { notation: "standard", maximumSignificantDigits: decimals }).format(measurement) + ' ' + unit;
```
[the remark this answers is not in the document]
yannik131: The original rounding logic doesn't make sense if the user can select the unit himself, I don't want 0.00000023 km² to turn into 0.0 km², which is why I switched to siginificant digits.

resources/assets/js/annotations/components/scaleLineIndicator.vue
```diff
@@ -21,7 +22,7 @@ export default {
     },
     computed: {
         scaleLineProperties() {
-            return new ScaleLineProperties(this.resolution, this.hasArea, this.pxWidthInMeter, this.unitMultipliers, this.unitNames);
+            return new ScaleLineProperties(this.resolution, this.hasArea, this.pxWidthInMeter, Settings.get('preferredUnit'));
         },
```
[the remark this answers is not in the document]
yannik131: See above, this is reactive

yannik131:
@mzur The preferred unit can be set using a dropdown in the settings tab, where `auto` is the default value at the top of the list (previous behaviour). Let me know if you prefer a switch to disable a preferred unit instead, I didn't really know where to put it (maybe to the left of the dropdown, disable/enable the dropdown accordingly?).
To accomodate large/small numbers I switched from the `Math.round` logic to `Intl.NumberFormat("en-US", { maximumSignificantDigits: ... })` which will display leading zeros in small numbers and group digits of larger numbers using commas (388000.02 mm² -> 388,000.02 px²).
Both the scale line and measure tooltips already do a decent job of displaying larger numbers, so I didn't change anything there.

resources/assets/js/annotations/components/measureTooltip.vue
```diff
@@ -79,22 +84,27 @@ export default {
...
+            }).format(measurement) + ' ' + unit;
         },
         unitNearest(measurement, multipliers, min) {
             if (measurement === 0) {
                 return multipliers.length - 1;
             }
 
+            if (UnitNames.indexOf(this.preferredUnit) !== -1) {
+                return UnitNames.indexOf(this.preferredUnit);
+            }
```
mzur: If there is a preferred unit it should be used even if the measurement is 0.

resources/assets/js/annotations/components/measureTooltip.vue
```diff
@@ -79,22 +84,27 @@ export default {
...
-                let index = this.unitNearest(length, this.unitMultipliers);
-                unit = this.unitNames[index];
-                length = length / this.unitMultipliers[index];
+                let index = this.unitNearest(length, UnitMultipliers);
+                unit = UnitNames[index];
+                length = length / UnitMultipliers[index];
             }
 
-            return this.formatMeasurement(length, unit);
+            return this.formatMeasurement(length, unit, 1);
```
mzur, line 92: Maybe we should change the number of decimals to 3 too? What do you think?

resources/assets/js/annotations/annotatorContainer.vue
```diff
@@ -593,6 +594,9 @@ export default {
                 case 'draftAnnotationUsesLabelColor':
                     this.draftAnnotationUsesLabelColor = value;
                     break;
+                case 'preferredUnit':
+                    this.preferredUnit = value;
+                    break;
```
mzur: This is never used. The components read the value directly from the settings instead. Either pass this as a prop (or use prop drilling) or remove it here and keep using Settings directly.

resources/assets/js/annotations/components/measureTooltip.vue
yannik131: Turns out `maximumSignificantDigits: 3` is the wrong thing to adjust here. I compared the old and new behaviour for some values:

| Old            | New        |
|----------------|------------|
| 3241.164 px²   | 3,240 px²   |
| 6.5 px           | 6,54 px     |
| 12.4 px          | 12.4 px     |
| 112.2 px         | 112 px      |
| 1113.5 px        | 1,110 px   |

To replicate the old precision, I switched to setting `minimumFractionDigits` and `maximumFractionDigits` to 1 for length and 3 for area. Now we have thousands separators + 1/3 digits after comma like before. I don't know if we should increase length precision to 3 digits after the comma as well, doesn't seem necessary if no one has complained about it so far.

mzur (approved):

PR #1470, opened 2026-05-22
Right click drag for polygon brush tools

### LINKED ISSUE #795: Move image/video with advanced polygon tools
If one of the polygon brush/eraser/fill or magic wand tools is active, the user can no longer grab and move the image/video because a click and drag activates the tool. If these tools are active, allow a click/drag interaction with the right mouse button.

Also mention this in the manual.

yannik131:
Closes #795


yannik131:
@mzur The fill/erase/magic wand/brush tools were all triggered with the right mouse button as well, I hope this doesn't break anybody's work flow. If it does we could add additional modifiers like shift + right click for dragging.
Also, I made right click drag work only when no modifiers (shift, alt, ...) are present.

resources/assets/js/annotations/components/annotationCanvas.vue
```diff
@@ -283,6 +289,25 @@ export default {
...
             }));
 
+            map.addInteraction(new DragPan({
+                condition: (mapBrowserEvent) => {
+                    return mapBrowserEvent.originalEvent.button === 2 && noModifierKeys(mapBrowserEvent) && this.isBrushOrWandMode;
+                },
+            }));
+
+            map.getViewport().addEventListener('contextmenu', (e) => {
+                switch (this.interactionMode) {
```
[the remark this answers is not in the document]
yannik131: This is not a SPA

resources/assets/js/annotations/ol/MagicWandInteraction.js
```diff
@@ -205,6 +205,9 @@ class MagicWandInteraction extends PointerInteraction {
      * Start drawing of a sketch.
      */
     handleDownEvent(e) {
+        if (e.originalEvent.button === 2) {
+            return false;
+        }
```
mzur: Maybe implement this as a generic `condition` similar to the other interactions?

resources/assets/js/videos/components/videoScreen/drawInteractions.vue
```diff
@@ -185,7 +185,7 @@ export default {
                             style: Styles.editing,
                             indicatorPointStyle: Styles.editing,
                             indicatorCrossStyle: Styles.cross,
-                            simplifyTolerant: 0.1,
+                            simplifyTolerant: 0.1
```
mzur: The codebase prefers trailing commas everywhere.


```suggestion
                            simplifyTolerant: 0.1,
```

resources/assets/js/annotations/components/annotationCanvas/polygonBrushInteraction.vue
```diff
@@ -64,6 +64,7 @@ export default {
                     style: Styles.editing,
                     brushRadius: brushRadius,
                     resizeCondition: altKeyOnly,
+                    condition: (event) => event.originalEvent.button !== 2
```
mzur: Maybe we could also use the existing `primaryAction` condition as "must be something like left click" instead of "must not be right click"? Or we implement a custom `rightClick` condition (see `annotations/ol/events/conditions.js`) and (re-)use `!rightClick`.

Example:


```suggestion
                    condition: primaryAction
```

resources/assets/js/annotations/components/annotationCanvas.vue
```diff
@@ -283,6 +289,18 @@ export default {
...
+                    return mapBrowserEvent.originalEvent.button === 2 && noModifierKeys(mapBrowserEvent) && this.isBrushOrWandMode;
+                },
+            }));
+
+            map.getViewport().addEventListener('contextmenu', (e) => {
+                if (this.isBrushOrWandMode) {
+                    e.preventDefault();
+                }
+            });
+
```
mzur: I'd suggest to put this into the `polygonBrushInteraction` mixin but this wouldn't fit with the magic wand tool. Maybe add a DragPan for the polygon tools and another one for the magic wand interaction? This would repeat code but the separation is cleaner in the different files.

resources/assets/js/videos/components/videoScreen.vue
```diff
@@ -653,6 +660,18 @@ export default {
...
+                condition: (mapBrowserEvent) => {
+                    return mapBrowserEvent.originalEvent.button === 2 && noModifierKeys(mapBrowserEvent) && this.isBrushOrWandMode;
+                },
+            }));
+
+            map.getViewport().addEventListener('contextmenu', (e) => {
+                if (this.isBrushOrWandMode) {
+                    e.preventDefault();
+                }
+            });
```
mzur: Again maybe move this to the mixin files?

resources/views/partials/image-annotation-shortcuts.blade.php
```diff
@@ -158,6 +158,25 @@
...
+        </tr>
+    </thead>
+    <tbody>
+        <tr>
+            <td><kbd>Right click</kbd> (held)</td>
+            <td>Drag the map<br></td>
+        </tr>
+    </tbody>
+</table>
+
```
mzur: I'm not sure about this. Some kind of documentation is good but this is not really a keyboard shortcut. Maybe we should rather add a note about it to the manual articles on creating/editing annotations.

resources/views/partials/video-annotation-shortcuts.blade.php
```diff
@@ -172,6 +172,25 @@
...
+        </tr>
+    </thead>
+    <tbody>
+        <tr>
+            <td><kbd>Right click</kbd> (held)</td>
+            <td>Drag the map<br></td>
+        </tr>
+    </tbody>
+</table>
+
```
mzur: See above.

resources/assets/js/annotations/components/annotationCanvas.vue
yannik131: I left the contextmenu event-listener there and cut-pasted the other if to the mounted() methods in 

- `annotationCanvas/polygonBrushInteraction.vue`
- `videoScreen/polygonBrushInteractions.vue`
- `annotationCanvas/magicWandInteraction.vue`
- (is it needed in `ol/MagicWandInteraction.js` for the video tool? I'll check later with a mouse)

Is that what you meant?

resources/assets/js/annotations/ol/MagicWandInteraction.js
```diff
@@ -205,7 +206,7 @@ class MagicWandInteraction extends PointerInteraction {
      * Start drawing of a sketch.
      */
     handleDownEvent(e) {
-        if (e.originalEvent.button === 2) {
+        if (rightClick(e)) {
```
mzur: The new rightClick condition is great for the other cases but what about adding an "allow condition" to the MagicWandInteraction instead of a "deny condition"? With:
```js
new MagicWandInteraction({
   condition: primaryAction, // may also be the default
})
```
it only handles down events if the event matches the condition.

resources/assets/js/annotations/components/annotationCanvas.vue
mzur: You could also implement two contextmenu listeners in the separate mixins, one with `isUsingPolygonBrush` and one with `isMagicWanding`.

resources/assets/js/videos/components/videoScreen.vue
mzur: Same

resources/assets/js/annotations/components/annotationCanvas/magicWandInteraction.vue
```diff
@@ -84,6 +87,12 @@ export default {
         magicWandInteraction.on('drawend', this.handleNewFeature);
         magicWandInteraction.setActive(false);
         this.map.addInteraction(magicWandInteraction);
+
+        this.map.addInteraction(new DragPan({
+            condition: (mapBrowserEvent) => {
+                return rightClick(mapBrowserEvent) && noModifierKeys(mapBrowserEvent) && this.isBrushOrWandMode;
```
mzur:
```suggestion
                return rightClick(mapBrowserEvent) && noModifierKeys(mapBrowserEvent) && this.isMagicWanding;
```

resources/assets/js/annotations/components/annotationCanvas/polygonBrushInteraction.vue
```diff
@@ -159,6 +162,12 @@ export default {
         shiftClickSelectInteraction.on('select', this.handleFeatureSelect);
         shiftClickSelectInteraction.setActive(false);
         this.map.addInteraction(shiftClickSelectInteraction);
+
+        this.map.addInteraction(new DragPan({
+            condition: (mapBrowserEvent) => {
+                return rightClick(mapBrowserEvent) && noModifierKeys(mapBrowserEvent) && this.isBrushOrWandMode;
```
mzur:
```suggestion
                return rightClick(mapBrowserEvent) && noModifierKeys(mapBrowserEvent) && !this.isNotAPolygonTool;
```

resources/assets/js/videos/components/videoScreen/polygonBrushInteractions.vue
```diff
@@ -149,5 +152,12 @@ export default {
             this.keyboardOn('t', this.togglePolygonFill, 0, this.listenerSet);
         }
     },
+    mounted() {
+        this.map.addInteraction(new DragPan({
+            condition: (mapBrowserEvent) => {
+                return rightClick(mapBrowserEvent) && noModifierKeys(mapBrowserEvent) && this.isBrushOrWandMode;
```
mzur: Also handle only the polygon case here.

resources/assets/js/videos/components/videoScreen/polygonBrushInteractions.vue
```diff
@@ -149,5 +152,12 @@ export default {
             this.keyboardOn('t', this.togglePolygonFill, 0, this.listenerSet);
         }
     },
+    mounted() {
+        this.map.addInteraction(new DragPan({
```
mzur: And add something like this to the drawInteractions mixin.

resources/assets/js/annotations/ol/MagicWandInteraction.js
yannik131: `PointerInteraction` doesn't support this, so I added this explictly in the constructor and check in `handleDownEvent`.

resources/assets/js/videos/components/videoScreen/polygonBrushInteractions.vue
yannik131: @mzur You mean [like this](https://github.com/biigle/core/pull/1470/changes/f58ab1efad601cedd1c948ee68fe71f5f6852c0b)? Enable right-click drag for all other annotation tools as well? In that case we might as well get rid of all tool-specific event handlers and add a single global one in drawInteractions for both video and image to enable right-click drag-pan.

mzur: No I meant only in drawInteractions of the video annotation tool because this handles the Magic Wand tool. The DragPan interaction should only be available for the polygon (brush/eraser/fill) tools and Magic Wand.

resources/assets/js/annotations/ol/MagicWandInteraction.js
```diff
@@ -18,6 +17,8 @@ class MagicWandInteraction extends PointerInteraction {
         super(options);
         this.on('change:active', this.toggleActive);
 
+        this.condition = options.condition || (() => true);
```
mzur: Should be either `always` or `primaryAction` (from Openlayers) as fallback.

resources/assets/js/videos/components/videoScreen.vue
```diff
@@ -653,6 +659,12 @@ export default {
                 }),
             });
 
+            map.getViewport().addEventListener('contextmenu', (e) => {
+                if (this.isBrushOrWandMode) {
+                    e.preventDefault();
+                }
+            });
+
```
mzur: I still prefer this to be put into the separate mixins (once for polygons and once for magic wand) because this makes the intent clearer. Otherwise we need an explaining comment here.

resources/assets/js/annotations/components/annotationCanvas.vue
```diff
@@ -283,6 +289,12 @@ export default {
                 },
             }));
 
+            map.getViewport().addEventListener('contextmenu', (e) => {
+                if (this.isBrushOrWandMode) {
+                    e.preventDefault();
+                }
+            });
+
```
mzur: Same

mzur (commented):
Please resolve the branch conflicts.

yannik131:
@mzur I added a helper to `utils.js` for adding the right click panning. `utils.js` starts feeling crowded and mixes concerns. I propose we create named helper files at locations like `utils/map/addRightClickDragPanToMap.js`, `utils/scale/ScaleLineProperties.js`, ... (possibly without the intermediate directories like `map/`, `scale/`, ...).

mzur (approved):
Looks good, thanks!

PR #1478, opened 2026-05-29
Show LabelBOT popup despite no suggestions

### LINKED ISSUE #1337: Improve case with empty LabelBOT suggestions
Currently, if LabelBOT does not find any suggested labels, the pending annotation is deleted again. This could be frustrating for complicated (video) annotations because the user has to draw them again. Instead, we could show the LabelBOT popup that only contains the (focussed) typeahead so the user can choose a label manually. The popup could also contain the explanation text that LabelBOT was not able to suggest any label. Only if the user dismisses this popup, the annotation will be deleted.

### LINKED ISSUE #1510: LabelBOT crashes with videos and no suggestions
Fix this error:

https://github.com/user-attachments/assets/ce35d71f-dd04-4077-b766-6e51c9657ecb

gkourie:
When LabelBOT returns no suggestions, an empty array will be sent to the front end. The LabelBOT popup will show a message to type a label in the typeahead. For now the temp annotation will be saved in the database without a label and it will have the editing style in the front end until it's confirmed with a label, if not then it will be deleted.


mzur:
Please request my review again once you are through with the AI.

gkourie:
@mzur While trying to implement this for videos, I found some other issues:
1. When LabelBOT returns no results, the video disappears (this should be fixed with empty LabelBOT popup):
 
[Screencast from 2026-06-02 13-58-18.webm](https://github.com/user-attachments/assets/1bb0e9fd-ce6b-4ed8-9935-5bf30d89e1d0)

2. If LabelBOT popup is shown and the video is seeked, then the popup and its line feature are shown on the next frames without the annotation:

[Screencast from 2026-06-02 14-04-56.webm](https://github.com/user-attachments/assets/1ad1bec5-3733-4e73-b9b5-5151c44f395e)


Or was that intentional for 2 ? Otherwise, I could try to fix it here.

resources/assets/js/annotations/components/labelbotPopup.vue
```diff
@@ -10,7 +10,12 @@
         <i class="fas fa-grip-lines"></i>
     </div>
     <ul class="labelbot-labels">
-        <li
+        <li v-if="noLabels" class="labelbot-popup__message">
```
mzur, line 13:
```suggestion
        <li v-if="noLabels" class="labelbot-popup__message text-muted">
```

resources/assets/js/annotations/components/labelbotPopup.vue
```diff
@@ -10,7 +10,12 @@
...
+            <p>
+                LabelBOT could not find similar annotations. Enter a label manually below, or close this popup to delete the annotation.
```
mzur:
```suggestion
                LabelBOT could not find similar annotations. Enter a label below, or close this popup to delete the annotation.
```

app/Http/Controllers/Api/ImageAnnotationController.php
```diff
@@ -86,6 +86,13 @@ public function index(Request $request, $id)
         } else {
             $yieldAnnotations = function () use ($image, $load): Generator {
                 foreach ($image->annotations()->with($load)->lazy() as $annotation) {
+                    // Delete saved unlabeled LabelBOT annotations
+                    // Such annotations are only created when LabelBOT returns no results
+                    // and the user refreshes/closes the BIIGLE session without interacting with the empty LabelBOT popup.
+                    if ($annotation->labels->isEmpty()) {
+                        $annotation->delete();
+                        continue;
+                    }
```
mzur: We can't do it like this. The index endpoint must not delete anything. We have to update the behavior of the store endpoint instead.

app/Http/Controllers/Api/ImageAnnotationController.php
```diff
@@ -236,15 +243,19 @@ public function store(StoreImageAnnotation $request, LabelBotService $labelBotSe
...
-            $annotationLabel->confidence = $request->input('confidence');
-            $annotation->labels()->save($annotationLabel);
+            if ($label) {
+                $annotationLabel = new ImageAnnotationLabel;
+                $annotationLabel->label_id = $label->id;
+                $annotationLabel->user_id = $request->user()->id;
+                $annotationLabel->confidence = $request->input('confidence');
+                $annotation->labels()->save($annotationLabel);
+            }
         });
```
mzur, line 244: Without any label, no annotation should be created here. We could instead respond with a "204 No Content" to signal the UI that no label could be found. The UI can then keep the temporary annotation. If the user tries again with a manually chosen label, the annotation could be saved and the temporary annotation replaced. if the popup is closed then the temporary annotation is just discarded in the UI.

app/Http/Controllers/Api/VideoAnnotationController.php
```diff
@@ -241,15 +241,19 @@ public function store(StoreVideoAnnotation $request, LabelBotService $labelBotSe
...
-                'user_id' => $request->user()->id,
-                'annotation_id' => $annotation->id,
-            ]);
+            if ($label) {
+                VideoAnnotationLabel::create([
+                    'label_id' => $label->id,
+                    'user_id' => $request->user()->id,
+                    'annotation_id' => $annotation->id,
+                ]);
+            }
```
mzur: Same here. If no label is found, don't create an annotation and respond with a 204. Then handle the rest in the UI.

app/Http/Controllers/Api/VideoAnnotationController.php
mzur, line 1: Also update the test class.

mzur (changes requested):
Good that you fixed issue 1. Regarding issue 2 I feel that we thought about this when LabelBOT was implemented for videos but I might be wrong (or there was a regression). Please fix that too.

Comments below:

gkourie:
No unlabeled annotations will be saved to the database anymore. The OpenLayers feature element is now temporarily attached to the annotation element. If LabelBOT returns no results, the feature element ID is used as the popup ID (key), and its geometry is used for popup placement. For video annotations, the pending annotation is attached in the same way.

The [two issues](https://github.com/biigle/core/pull/1478#issuecomment-4602265918) mentioned previously should now be fixed. I also noticed that you, @mzur, already opened issue #1510 regarding the video screen crash.

Another issue I found was related to tracking annotations. When creating an ROI in a video, then seeking to another time, and finally clicking "Finish and Track Annotation" while LabelBOT is enabled, the video screen crashes. This happens because there are no available annotations to use as an anchor at the new current time:

[Screencast from 2026-07-31 12-17-16.webm](https://github.com/user-attachments/assets/c4181103-d099-4546-a9b2-65b99fd7d963)

This was already fixed for non-tracking annotations, and I have now applied the same fix for tracking annotations as well.

When clicking the play button while a LabelBOT popup is open, the popup remains visible until the video is paused or the user interacts with the popup. If the popup is empty, pausing the video will remove the entire pending annotation if no label has been entered. If the popup is not empty, the Top-1 is already saved.

resources/assets/sass/annotations/components/_labelbot-popup.scss
```diff
@@ -28,6 +28,15 @@
     border-radius: 50%;
 }
 
+.labelbot-popup__message {
+    list-style-type: none;
+    max-width: 300px;
+    margin: 0;
+    padding: $padding-small-vertical;
+    background-color: $body-bg;
+    color: $text-muted;
```
mzur, line 37: Most is already handled by `.labelbot-labels`.

```suggestion
    padding: $padding-small-vertical;
    color: $text-muted;
```

resources/assets/js/annotations/components/labelbotPopup.vue
```diff
@@ -294,7 +344,13 @@ export default {
 
         },
         createOverlay(annotationCanvas) {
-            const annotationFeature = annotationCanvas.annotationSource.getFeatureById(this.annotation.id);
+            let annotationFeature;
+            if (this.noLabels) {
+                annotationFeature = this.annotation.feature;
+                this.pendingAnnotation = true;
```
mzur, line 350: This case should automatically focus the label typeahead when the popup opens.

resources/assets/js/videos/videoContainer.vue
```diff
@@ -329,6 +336,12 @@ export default {
             }
         },
         async createAnnotation(pendingAnnotation) {
+            const pendingAnnotationFeature = pendingAnnotation.feature;
+            pendingAnnotation.feature = undefined;
+            delete pendingAnnotation.feature;
+
+            const track = pendingAnnotation.track
```
mzur, line 343: Tracking video annotations is broken (and was before this PR) when LabelBOT is not enabled or did not return results. The `track` attribute is not passed to `newAnnotation` below.

resources/assets/js/videos/videoContainer.vue
```diff
@@ -357,6 +370,11 @@ export default {
 
             try {
                 newAnnotation.labelbotImage = await pendingAnnotation.screenshotPromise;
+                // We need to add the feature, the pending annotation and track to the new annotation
+                // in case LabelBOT returns no results
+                newAnnotation.feature = pendingAnnotationFeature;
+                newAnnotation.pendingAnnotation = pendingAnnotation;
+                newAnnotation.track = track;
```
mzur, line 377: This can be removed when the `track` attribute is added to `newAnnotation` above.
```suggestion
```

resources/assets/js/annotations/stores/annotations.js
```diff
@@ -79,16 +79,20 @@ class Annotations {
...
-            .then(this.setDeselected.bind(this))
-            .then((annotation) => {
-                this.cache[imageId].then(function (annotations) {
-                    annotations.unshift(annotation);
-                });
-
-                return annotation;
-            });
+            .then((response) =>  {
+                if (response.status === 204) return null;
```
mzur, line 83: Please add a comment what this case means.

Thinking about this, maybe we should not use 204 after all, because it signals some kind of success. Maybe we should use 422 with a special error message for the `feature_vector` and then catch this and check the message here? A 422 will have better semantics here too because the promise chain will be skipped on an error response.

resources/assets/js/videos/videoContainer.vue
```diff
@@ -380,8 +398,14 @@ export default {
 
         },
         saveAnnotation(newAnnotation, pendingAnnotation) {
+            let pendingEmptyLabelBOTAnnotation = false;
             return VideoAnnotationApi.save({id: this.videoId}, newAnnotation)
                 .then((res) => {
+                    if (res.status === 204) {
+                        pendingEmptyLabelBOTAnnotation = true;
+                        return null;
+                    }
```
mzur, line 407: Same than above.

resources/assets/js/annotations/components/labelbotPopup.vue
```diff
@@ -182,6 +197,27 @@ export default {
...
+                    this.$parent.labelbotSource.removeFeature(feature);
+                } catch (e) {
+                    // ignore
+                }
+            };
+
+            delete annotation.feature;
+            delete annotation.labels;
+
+            this.$emit('new', { newAnnotation: annotation, rmvCallback: removeCallback});
```
mzur, line 218: The `rmvCallback` key is not what the annotatorContainer expects. I get an error:
```
Uncaught (in promise) TypeError: removeCallback is not a function
handleNewAnnotation http://127.0.0.1:5173/resources/assets/js/annotations/annotatorContainer.vue:472
```
Pass the annotation and remove callback as two separate arguments instead.

resources/assets/js/annotations/annotatorContainer.vue
```diff
@@ -439,8 +439,11 @@ export default {
             annotation.confidence = 1;
 
             let promise;
-
-            if (this.labelbotIsActive) {
+            let pendingAnnotation = false;
```
mzur, line 442: It's also a little unclear what "pending annotation" means in this context. Usually a pending annotation is one that is drawn but not saved yet. But here it means `labelBotReturnedNoResults`.

resources/assets/js/annotations/annotatorContainer.vue
```diff
@@ -449,18 +452,26 @@ export default {
...
                 .catch(handleErrorResponse)
-                // Remove the temporary annotation if saving succeeded or failed.
-                .finally(removeCallback);
+                // Remove the temporary annotation if saving succeeded or failed,
+                // if it's not a pending annotation
+                .finally(() => {
+                    if (!pendingAnnotation) {
+                        removeCallback();
+                    }
+                });
```
mzur, line 474: Maybe also do this instead of adding the guard in `setLastCreatedAnnotation`

```js
.then((annotation) => {
   if (!labelBotReturnedNoResults) {
      this.setLastCreatedAnnotation(annotation);
   }
})

File: resources/assets/js/annotations/components/annotationCanvas/labelBot.vue [hunk 15]
```diff
@@ -55,6 +57,9 @@ export default {
         },
     },
     methods: {
+        createNewLabelBOTAnnotation(event) {
```
mzur, line 60: Please use consistent casing:

```suggestion
        createNewLabelbotAnnotation(event) {
```

resources/assets/js/annotations/components/labelbotPopup.vue
```diff
@@ -182,6 +197,27 @@ export default {
...
     },
     methods: {
+        createAndClose(label) {
+            this.pendingAnnotation = false;
+
+            const annotation = { ...this.annotation };
+            annotation.label_id = label.id;
+            const feature = annotation.feature;
+
+            let removeCallback = () => {
```
mzur, line 207: Instead of passing the feature to the popup via the annotation prop, you could pass the original `removeCallback` (which was not used in annotatorContainer). This will keep the actual logic of removing the pending annotation in annotationCanvas. Also you will know when not to use a removeCallback in the video annotation tool.

Let me know if this doesn't make any sense 😉 

As to the popupKey, we don't actually need this any more. We now only allow a single popup at a time so we could refactor this from the array of popups (requiring keys) to only a single popup without key.

resources/assets/js/annotations/components/annotationCanvas.vue
```diff
@@ -590,6 +590,8 @@ export default {
                 e.feature.set('color', '5bc0de');
                 e.feature.setStyle(Styles.editing);
 
+                newAnnotation.feature = e.feature, // we need this temporarily if LabelBOT returns no results
```
mzur, line 593: See above.

resources/assets/js/videos/components/videoScreen/drawInteractions.vue
```diff
@@ -339,6 +346,11 @@ export default {
             const points = this.getPointsFromGeometry(e.feature.getGeometry());
             this.pendingAnnotation.points.push(points);
 
+            // We add temporarily the feature to the pending annotation
+            // so in case LabelBOT did not return any results, the unique id of 
+            // the feature could be used as LabelBOT popup id
+            this.pendingAnnotation.feature = e.feature;
```
mzur, line 352: See above.

resources/assets/js/annotations/components/annotationCanvas/labelBot.vue
```diff
@@ -67,6 +72,9 @@ export default {
         handleDeleteLabelbotAnnotation(annotation) {
             this.$emit('delete', [annotation]);
         },
+        handleDeleteLabelbotPendingAnnotation(annotation) {
+            this.$emit('delete-pending', annotation);
```
mzur, line 76: Where is this actually handled in the image annotation tool? I don't see it anywhere. Instead the pending annotation is removed with the callback.

resources/assets/js/annotations/components/labelbotPopup.vue
```diff
@@ -344,7 +400,11 @@ export default {
             const line = new LineString([popupPosition, popupPosition]);
             this.lineFeature = markRaw(new Feature(line));
             this.lineFeature.set('unselectable', true);
-            this.lineFeature.set('color', this.labels[0].color);
+            if (this.noLabels) {
+                this.lineFeature.set('color', '5bc0de');
```
mzur, line 404: Add a comment that this is the "info" color.

resources/assets/js/annotations/components/annotationCanvas.vue
mzur, line 593:
```suggestion
                newAnnotation.feature = e.feature; // we need this temporarily if LabelBOT returns no results
```

resources/assets/js/annotations/components/labelbotPopup.vue
```diff
@@ -105,6 +112,7 @@ export default {
             shouldHaveProgressBar: true,
             maybeGetsAttention: false,
             typeaheadFocused: false,
+            pendingAnnotation: false,
```
mzur, line 115: Better use:

```suggestion
            isPendingAnnotation: false,
```

tests/php/Http/Controllers/Api/ImageAnnotationControllerTest.php
```diff
@@ -331,6 +331,16 @@ public function testStoreWithFeatureVectorWithoutHNSW()
...
+        // Query an empty DB
+        $response = $this->post("/api/v1/images/{$this->image->id}/annotations", [
+            'shape_id' => Shape::pointId(),
+            'feature_vector' => range(1, 384),
+            'confidence' => 0.5,
+            'points' => [10, 11],
+        ]);
+        // We expect no content
+        $response->assertStatus(204);
+
```
mzur, line 343: This should be added as a separate test case.

tests/php/Http/Controllers/Api/VideoAnnotationControllerTest.php
```diff
@@ -864,6 +864,16 @@ public function testStoreWithFeatureVectorWithoutHNSW()
...
+        // Query an empty DB
+        $response = $this->json('POST', "api/v1/videos/{$this->video->id}/annotations", [
+            'shape_id' => Shape::pointId(),
+            'feature_vector' => range(1, 384),
+            'points' => [[10, 11]],
+            'frames' => [0.0],
+        ]);
+        // We expect no content
+        $response->assertStatus(204);
+
```
mzur, line 876: Same.

mzur (changes requested):
I merged the current master so the code works with my new DINO ONNX (generated with a newer Torch version).

I think the popup should close when there are no labels and the user clicks play. It closes anyway once the user clicks pause, so we can just close it earlier.

If the popup timeout is enabled, two <kbd>Esc</kbd> presses are needed to close the popup. It should close right away.

Maybe the whole logic of remove callbacks, direct calls to the OpenLayers source and `delete-pending` events can be consolidated a little more. They all do similar things but it feels very inconsistent.

There are several missing semicolons in the code. Please rerun the AI review in the next iteration before requesting my own review.

resources/assets/js/annotations/components/labelbotPopup.vue
mzur: Also the remove callback is ignored in the next step when the `new` event is forwarded.

PR #1480, opened 2026-06-03
Make typahead scrollable

### LINKED ISSUE #1474: Make label typeahead scrollable
In the [label typeahead](https://github.com/biigle/core/blob/04f5acc728bc016b278460732c7edd7a7769b55e/resources/assets/js/label-trees/components/labelTypeahead.vue), more than the default 5 results may be relevant. It should activate the `scrollable` state of the underlying typehead component so users get up to 50 scrollable results.

dbrembilla:
Resolves #1474 

Makes the typeahead for labels scrollable

https://github.com/user-attachments/assets/e7433739-6efd-47d8-880e-6f8235f75fba


mzur (approved):
Thanks!

PR #1491, opened 2026-06-17
Pause/resume for lawnmower

### LINKED ISSUE #547: Feature request: Pause / Resume for Lawnmower Mode
Dear Martin,

whilst scanning a large image with the lawnmower mode to make sure we do not miss anything, we often need to zoom in for investigating details of individual objects. Alas, lawnmower mode and zooming do not go well together and cause the lawnmower to reposition the current field of view erratically (or if not erratically, at least not in a helful way).

Our suggestion to solve this issue (and bring some other benefit) would be to have a pause / resume function for the lawnmower mode: Interrupt it either with a specific keyboard shortcut or (optionally, but nice to have) automatically when the view is zoomed in or out, and then continue it at the exact position and zoom level used before pausing with another specific keyboard shortcut. 

Another nice featere would be to be able to optionally start the lawnmower mode from the current field of view instead of the lower left corner, to be able to continue analysing after a long break or in another session, or even to log the progress and resume from the last used position.

Thank you very much for BIIGLE any our constant support,
Michael

yannik131:
Closes #547

### DIFF
--- a/resources/assets/js/annotations/annotatorContainer.vue
+++ b/resources/assets/js/annotations/annotatorContainer.vue
@@ -599,17 +599,17 @@ export default {
                     break;
             }
         },
-        handleLawnmowerStateTransitionRequest(targetState) {
+        setLawnmowerState(targetState) {
             const transition = `${this.currentLawnmowerState}->${targetState}`;
 
             switch(transition) {
-                case 'playing->paused':
+                case 'active->paused':
                     this.lawnmowerSaveState = LawnmowerSaveState.SAVE;
                     break;
-                case 'paused->playing':
+                case 'paused->active':
                     this.lawnmowerSaveState = LawnmowerSaveState.LOAD;
                     break;
-                case 'paused->stopped':
+                case 'paused->inactive':
                     this.lawnmowerSaveState = LawnmowerSaveState.DISCARD;
                     break;
             }
@@ -624,7 +624,7 @@ export default {
         },
         handleAnnotationModeChange(mode, data) {
             if (mode !== 'default' && mode !== 'lawnmower') {
-                this.handleLawnmowerStateTransitionRequest(PlayPauseState.INACTIVE);
+                this.setLawnmowerState(PlayPauseState.INACTIVE);
             }
             this.annotationMode = mode;
             this.annotationModeCarry = null;
--- a/resources/assets/js/annotations/annotatorContainer.vue
+++ b/resources/assets/js/annotations/annotatorContainer.vue
@@ -97,7 +97,6 @@ export default {
             imageFilenames: {},
             labelTrees: [],
             projectIds: [],
-            currentLawnmowerState: PlayPauseState.INACTIVE,
             lawnmowerSaveState: "",
             volare: null,
             pauseLawnmowerAutomatically: false,
@@ -137,7 +136,7 @@ export default {
             return this.filteredAnnotations.filter((a) => a.selected);
         },
         isDefaultAnnotationMode() {
-            return this.annotationMode === 'default';
+            return this.annotationMode === 'default' || this.annotationMode === 'volarePaused' || this.annotationMode === 'lawnmowerPaused';
         },
         isLawnmowerAnnotationMode() {
             return this.annotationMode === 'lawnmower';
@@ -544,32 +543,15 @@ export default {
         enableAutomaticLawnmowerPausing() {
             this.pauseLawnmowerAutomatically = true;
         },
+        disableAutomaticLawnmowerPausing() {
+            this.pauseLawnmowerAutomatically = false;
+        },
         pauseLawnmowerAfterViewportChanged() {
             if (!this.isLawnmowerAnnotationMode || !this.pauseLawnmowerAutomatically) {
                 return;
             }
 
-            this.setLawnmowerState('paused');
-        },
-        setLawnmowerState(targetState) {
-            const transition = `${this.currentLawnmowerState}->${targetState}`;
-            if (targetState === 'active') {
-                this.pauseLawnmowerAutomatically = false;
-            }
-
-            switch (transition) {
-                case 'active->paused':
-                    this.lawnmowerSaveState = LawnmowerSaveState.SAVE;
-                    break;
-                case 'paused->active':
-                    this.lawnmowerSaveState = LawnmowerSaveState.LOAD;
-                    break;
-                case 'paused->inactive':
-                    this.lawnmowerSaveState = LawnmowerSaveState.DISCARD;
-                    break;
-            }
-
-            this.currentLawnmowerState = targetState;
+            this.$refs.annotationModesTab.pauseLawnmower();
         },
         showImageWithId(imageId) {
             const index = this.imagesIds.indexOf(imageId);
@@ -709,6 +691,15 @@ export default {
                 this.openTab = '';
             }
         },
+        annotationMode(oldMode, newMode) {
+            if (newMode === 'lawnmowerPaused') {
+                this.lawnmowerSaveState = LawnmowerSaveState.SAVE;
+            } else if (oldMode === 'lawnmowerPaused' && newMode === 'lawnmower') {
+                this.lawnmowerSaveState = LawnmowerSaveState.LOAD;
+            } else {
+                this.lawnmowerSaveState = LawnmowerSaveState.DISCARD;
+            }
+        }
     },
     created() {
         this.allImagesIds = biigle.$require('annotations.imagesIds');
@@ -796,6 +787,7 @@ export default {
             image: computed(() => this.image),
             mapResolution: computed(() => this.mapResolution),
             showImageWithId: this.showImageWithId,
+            annotationMode: computed(() => this.annotationMode),
         });
     },
 };
--- a/resources/assets/js/annotations/annotatorContainer.vue
+++ b/resources/assets/js/annotations/annotatorContainer.vue
@@ -647,7 +647,7 @@ export default {
                 this.annotations = annotations;
 
                 // Wait for the canvas to be updated, otherwise lawnmower mode
-                // could compute the current section using the old viewport
+                // may compute the current section using the old viewport
                 await this.$nextTick();
 
                 this.maybeUpdateAnnotationMode();
--- a/resources/assets/js/annotations/annotatorContainer.vue
+++ b/resources/assets/js/annotations/annotatorContainer.vue
@@ -290,6 +290,9 @@ export default {
                 x: Math.round(viewport.center[0]),
                 y: Math.round(viewport.center[1]),
             });
+            
+        },
+        handleMapMovestart() {
             this.pauseLawnmowerAfterViewportChanged();
         },
         // Handler for the select event fired by the global event bus.
@@ -691,7 +694,7 @@ export default {
                 this.openTab = '';
             }
         },
-        annotationMode(oldMode, newMode) {
+        annotationMode(newMode, oldMode) {
             if (newMode === 'lawnmowerPaused') {
                 this.lawnmowerSaveState = LawnmowerSaveState.SAVE;
             } else if (oldMode === 'lawnmowerPaused' && newMode === 'lawnmower') {
--- a/resources/assets/js/annotations/annotatorContainer.vue
+++ b/resources/assets/js/annotations/annotatorContainer.vue
@@ -23,6 +23,8 @@ import {CrossOriginTiffError} from './stores/images.js';
 import {debounce} from '@/core/utils.js';
 import {handleErrorResponse} from '@/core/messages/store.js';
 import {urlParams as UrlParams} from '@/core/utils.js';
+import { PlayPauseState } from '../core/components/playPause.vue';
+import { LawnmowerSaveState } from './components/annotationCanvas/lawnmower.vue';
 
 const asyncAnnotationCanvas = defineAsyncComponent({
     loader: function () {
@@ -95,6 +97,8 @@ export default {
             imageFilenames: {},
             labelTrees: [],
             projectIds: [],
+            currentLawnmowerState: PlayPauseState.STOPPED,
+            lawnmowerSaveState: ""
         };
     },
     provide() {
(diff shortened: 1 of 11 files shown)


yannik131:
Architectural sketch of what I implemented:

- introduce a PlayPause component for the two buttons
- annotatorContainer defines 
  - currentLawnmowerState, which is passed as a prop to annotationModesTab -> playPause
  - lawnmowerSaveState (load, save, discard), passed as a prop to annotationCanvas/lawnmower.vue, watcher for changes
- data flow is then: 
  1. user requests change in playPause.vue (emit lawnmowerStateChangeRequested)
  2. annotationModesTab.vue re-emits
  3. annotatorContainer.vue changes the state if possible and updates both props
  4. change is reflected in playPause via the currentLawnmowerState prop, annotationModesTab.vue updates the mode
  5. annotationCanvas/lawnmower.vue loads/saves/discards the state (current image, current section) from local storage tied to volume ID
- then do the same for label review mode

yannik131:
@mzur It's working now, but the architecture is not really consistent (mixin/options API for lawnmower, composable for volare). I managed to move the volare-logic from the AnnotatorContainer into the composable and would like to do the same for the lawnmower, but would first like to get some feedback from you regarding the overall structure. I'll mark it ready for review anyways. 

Edit: Oh, and also I've run out of AI credits, so I can't use copilot for reviews right now. Since the switch to the new credits-based pricing model a single larger AI code review uses up about 25% of my $2.00 worth of free credits. 

Edit 2: I also noticed a bug in the original behaviour that is still present: Focus annotation in volare, filter annotations that keeps the focussed annotation on the screen, remove filter -> wrong annotation is focussed. I will resume development once I get your feedback regarding the structure, though.

mzur:
I'll skip the review until we finished the discussion in the issue. Let me know if I should request an AI review for you here.

yannik131:
@mzur I don't think volare should be paused when the viewport is changed, that would likely break workflows where users zoom in while in volare. Otherwise this is as we discussed. Let me know what you think about the volare.js structure and I will turn `lawnmower.vue` into a similar composable.

mzur:
Do you need/want an AI review first?

yannik131:
@mzur Sure, but I'd also like feedback from you on the composable structure overall. I will use that feedback for the upcoming lawnmower composable.

resources/assets/js/annotations/annotatorContainer.vue
```diff
@@ -471,7 +416,7 @@ export default {
                     confidence: 1,
                 };
 
-                if (this.isVolareAnnotationMode) {
+                if (this.volare.volareModeIsActive) {
```
mzur: I don't get how this is exposed by `useVolareMode`. It's a private const.

resources/assets/js/annotations/annotatorContainer.vue
```diff
@@ -780,6 +749,8 @@ export default {
         Events.on('annotations.delete', this.handleDeleteAnnotation);
         Events.on('annotations.focus', this.focusAnnotation);
 
+        // TODO Are these necessary? Should they be moved to the new volare composable?
+
```
mzur: No, this is independent from Volare. If the url param is present (e.g. from clicking on a thumbnail in Largo) then the annotation tool should focus the annotation once but it should not enable Volare.

mzur (commented):
This is not a full review, just my take on the composable approach with volare. I like a composable approach like this in general. The main benefit could be that logic can be shared across similar parts of BIIGLE (e.g. image and video annotation tools). All the current mixins that are used more than once are possible candidates for refactoring. Also the mixins of the annotationCanvas and videoScreen could maybe be (partially) merged and reused.

However in the case of volare it "just" moves code from one file to another. It makes the annotatorContainer slightly smaller but also introduces inconsistent architecture now (as the other annotation modes don't use the composable approach). And since the annotation modes are exclusive to the image annotation tool, I don't really see a benefit in making them all composable. It will be mostly refactoring for it's own sake which we should only do with our little resources if there is a real benefit in terms of code complexity.

So in summary I like the composable approach but volare and lawnmower mode may be the wrong features to pilot this. But please feel free to argue your point. We can also discuss this in a VC if you like.

yannik131:
@mzur I like the composable approach because of the separation of concerns it offers. Decoupling things from the rest of the codebase makes code testable and easier to reason about. This helps other developers to understand the code, including myself in the future. I do agree that the time investment is probably not worth it here. But I will start considering separating code into composables more often from now on and might default to it for new functionality.
I'll leave lawnmower as it is for now. Ready for review.

resources/assets/js/annotations/components/annotationModesTab.vue
```diff
@@ -103,19 +176,46 @@ export default {
...
                 case 'regularSampling':
-                    this.$emit('change', mode, [this.regularSamplingRows, this.regularSamplingColumns]);
+                    this.$emit('annotation-mode-changed', newMode, [this.regularSamplingRows, this.regularSamplingColumns]);
                     break;
                 default:
-                    this.$emit('change', mode);
+                    this.$emit('annotation-mode-changed', newMode);
             }
         },
+        timeAgo(timestamp) {
```
mzur, line 192: This looks like it could be a good util method.

resources/views/annotations/show/tabs/annotationModes.blade.php
```diff
@@ -1,33 +1,65 @@
...
 @push('scripts')
 <script type="text/html" id="annotation-modes-tab-template">
     <div class="annotator-tab annotator-tab--annotation-modes">
         <div class="sidebar-tab__section">
             <h5 title="Cycle through all annotations">Volume Label Review<br><small>cycle through annotations</small></h5>
-            <power-toggle :active="isVolareActive" title-off="Start cycling through all annotations" title-on="Stop cycling through all annotations 𝗘𝘀𝗰" v-on:on="startVolare" v-on:off="resetMode"></power-toggle>
+            <play-pause
+                playButtonTitleText="Start cycling through all annotations"
+                stopButtonTitleText="Stop cycling through all annotations 𝗘𝘀𝗰"
+                pauseButtonTitleText="Pause Volare"
```
mzur, line 17: The convention is to use kebab-case for attributes.

resources/views/annotations/show/tabs/annotationModes.blade.php
```diff
@@ -1,33 +1,65 @@
...
         </div>
 
         <div class="sidebar-tab__section">
             <h5 title="Cycle through image sections">Lawnmower Mode<br><small>cycle through image sections</small></h5>
-            <power-toggle :active="isLawnmowerActive" title-off="Start cycling through image sections" title-on="Stop cycling through image sections 𝗘𝘀𝗰" v-on:on="startLawnmower" v-on:off="resetMode"></power-toggle>
+            <play-pause
+                playButtonTitleText="Start cycling through image sections"
+                stopButtonTitleText="Stop cycling through image sections 𝗘𝘀𝗰"
+                pauseButtonTitleText="Pause Lawnmower"
+                :state="currentLawnmowerState"
```
mzur, line 33: same

resources/assets/js/annotations/annotatorContainer.vue
```diff
@@ -813,6 +782,19 @@ export default {
         Keyboard.on('C', this.selectLastAnnotation, 0, this.listenerSet);
 
         this.initLabelBot();
+
+        this.volare = useVolareMode({
+            filteredAnnotations: computed(() => this.filteredAnnotations),
+            selectedAnnotations: computed(() => this.selectedAnnotations),
+            focusAnnotationInCanvas: this.focusAnnotation,
+            fitImageInCanvas: (...args) => this.$refs.canvas.fitImage(...args),
+            annotationFilter: this.annotationFilter,
```
mzur, line 791: This is usually `null`, so the volare watcher would not pick up any changes.


```suggestion
            annotationFilter: computed( () => this.annotationFilter),
```

resources/assets/js/annotations/components/annotationCanvas/lawnmower.vue
```diff
@@ -78,13 +98,17 @@ export default {
         },
         showImageSection(section) {
             if (section[0] < this.imageSectionSteps[0] && section[1] < this.imageSectionSteps[1] && section[0] >= 0 && section[1] >= 0) {
+                this.$emit('lawnmower-pre-viewport-change');
                 this.imageSection = section;
                 // Don't make imageSectionCenter a computed property because it
                 // would automatically update when the resolution changes. But we
                 // need the old value to compute the new image section in the
                 // resolution watcher first!
                 this.imageSectionCenter = this.getImageSectionCenter(section);
```
mzur, line 107: This comment is stale.

resources/assets/js/annotations/components/annotationCanvas/lawnmower.vue
```diff
@@ -118,33 +142,104 @@ export default {
...
-        // Update the current image section if either the resolution or the map size
-        // changed. viewExtent depends on both so we can use it as watcher.
-        viewExtent() {
-            if (!this.isLawnmowerAnnotationMode || !Number.isInteger(this.imageSectionSteps[0]) || !Number.isInteger(this.imageSectionSteps[1])) {
+        getLawnmowerStorageKey() {
+            const volumeId = biigle.$require('annotations.volumeId');
+            return `lawnmower-state-${volumeId}`;
+        },
+        saveCurrentLawnmowerState() {
+            if (this.getSavedLawnmowerState()) {
```
mzur, line 150: Why not overwrite the saved state? Presumably, if a user uses lawnmower mode a second time, this is the new "relevant" state that should be saved.

resources/assets/js/annotations/components/annotationCanvas/volare.js
```diff
@@ -0,0 +1,209 @@
...
+        return annotationMode.value === 'volare';
+    });
+
+    function getVolareStorageKey() {
+        const volumeId = biigle.$require('annotations.volumeId');
+        return `volare-state-${volumeId}`;
+    }
+
+    function saveCurrentVolareState() {
+        if (getSavedVolareState()) {
```
mzur, line 36: Same question.

resources/assets/js/annotations/components/annotationCanvas/volare.js
```diff
@@ -0,0 +1,209 @@
...
+        if (state.imageId !== image.value.id) {
+            showImageWithId(state.imageId);
+            return;
+        }
+
+        const savedID = state.focussedAnnotationId;
+        discardSavedVolareState();
+        resuming = false;
+
+        focussedAnnotationIndex.value = null;
```
mzur, line 153: Loading/resuming always focuses on annotation 0 first before switching to the actual saved annotation (after nextTick). Maybe this can be updated to avoid the unneeded focus.

resources/assets/js/annotations/components/annotationCanvas/lawnmower.vue
```diff
@@ -118,33 +142,104 @@ export default {
...
+            }
+
+            this.$emit('lawnmower-pre-viewport-change');
+
+            const view = this.map.getView();
+            view.setResolution(state.resolution);
+
+            this.$nextTick(() => {
+                this.imageSectionCenter = state.center;
+                this.imageSection = state.imageSection;
```
mzur, line 204: Does this also work as expected if the state is restored on a resized map (e.g. if the sidebar is now expanded or the window size changed)?

resources/assets/js/annotations/components/annotationCanvas/lawnmower.vue
```diff
@@ -1,17 +1,37 @@
...
+        'restore-lawnmower-image',
+        'lawnmower-pre-viewport-change',
+        'lawnmower-post-viewport-change',
+        'restore-lawnmower-pause-state',
+    ],
+    props: {
+        lawnmowerSaveState: {
+            type: String,
+            required: true
+        }
```
mzur, line 24: I'm not sure about this. This prop really acts like a command. Elsewhere this is done with direct method calls on a ref. If we don't want this, maybe this could be a hint that the lawnmower logic should also be moved up to the annotatorContainer which just tells the canvas which image section to show? Maybe migrate lawnmower mode to a composable after all?

resources/assets/js/annotations/components/annotationCanvas/volare.js
mzur, line 1: This file is no longer a canvas mixin but logic of the container. Put it into a "composables" dir?

resources/assets/js/annotations/annotatorContainer.vue
```diff
@@ -131,14 +134,8 @@ export default {
         selectedAnnotations() {
             return this.filteredAnnotations.filter((a) => a.selected);
         },
-        focussedAnnotation() {
-            return this.filteredAnnotations[this.focussedAnnotationIndex];
-        },
         isDefaultAnnotationMode() {
```
mzur, line 137: This name is now misleading. Since it's only used for the highlight, maybe negate it and call it "should HighlightAnnotationMode", then update the use in the template.

mzur (commented):
The time ("40s ago") should appear immediately if pause is activated. It can show "just now" if the time is <60s. Also it should tick/update (maybe once a minute).

The title of the play buttons should change if the mode is paused (e.g. "continue...").

The title of the pause buttons should change if the mode is paused (e.g. "cancel...").

The title of the inactive pause buttons should explain what it does (save the state so user can resume here later).

This behavior should also be explained in the manual article.

Images can be missing if they were deleted from the volume or a volume filter is active. What happens if a volare/lawnmower state is restored but the image is missing?

<kbd>Esc</kbd> doesn't seem to cancel an active volare/lawnmower mode.

Collapsing the sidebar or resizing the browser window (i.e. resizing the map) pauses lawnmower mode but it should not.

mzur (commented):
One more thing: If volare is enabled but the currently focussed annotation disappears because of a filter in the annotations tab, the focussed annotation and viewport is updated correctly. But if the filter is then disabled again, the original annotation is focussed but the viewport does not jump back.

https://github.com/user-attachments/assets/e3e7e957-5035-4ebf-93d7-bd10eed55fa7

PR #1529, opened 2026-07-27
1240 improve control button active state

### LINKED ISSUE #1240: Improve control button active state
A user reported that it's difficult to distinguish between the hover and the active state of control buttons (the blue buttons in the annotation tools). The two states can occur next to each other which is most confusing. Think about an improved active state, maybe with a different background color or with a brightly colored (second) border. I'd prefer a minimal approach as these buttons are iconic for the annotation tools and the users are used to the UI. So I'd favor a new border over the background color it it works alright.

Confusing examples:

Image

Image

parimalgavali:
Fixes #1240

## Summary

- Use the annotation-selection orange as the background for active control buttons.
- Kept the grouped parent buttons unchanged while highlighting the selected sub-control.
- Preserved the existing styling of danger controls.

### DIFF
--- a/resources/assets/sass/annotations/components/_control-button.scss
+++ b/resources/assets/sass/annotations/components/_control-button.scss
@@ -1,3 +1,5 @@
+$control-button-active-color: #ff5e00;
+
 .control-button {
     position: relative;
     border-radius: 2px;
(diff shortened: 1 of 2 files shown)


resources/assets/sass/annotations/components/_control-button.scss
```diff
@@ -1,3 +1,5 @@
+$control-button-active-color: #ff5e00;
```
mzur: Please try `$brand-info`. I think it might work after all.

The orange color looks too much like "warning/danger".

resources/assets/sass/annotations/components/_control-button.scss
```diff
@@ -13,6 +15,12 @@
         background-color: fade-out($ol-control-color, 0.3);
     }
 
+    &:not(.control-button--danger):not(.control-button--has-sub-controls) {
```
mzur: Buttons with sub controls can also be active themselves.

```suggestion
    &:not(.control-button--danger) {
```

mzur (changes requested):

mzur (approved):

Issue #391, opened 2021-11-02
Seeking can produce different video frames in Firefox and Chrome


mzur:
Seeking to the time of a video annotation can produce different vieo frames in Firefox and Chrome (e.g. [video 6018](https://biigle.de/videos/6018/annotations?x=960&y=540&r=188&t=70225)). This can result in erroneously displayed annotation locations in the video.

This seems to be a problem caused by the video encoding (presumably through conversion with VLC). Judging by the Largo patches generated by ffmpeg with the same annotation timestamp, Firefox seems to show the wrong video frame.

mzur:
I produced a minimal working example. Based on this, Firefox would be correct.
```html
<video id="video" src="http://distribution.bbb3d.renderfarming.net/video/mp4/bbb_sunflower_1080p_60fps_normal.mp4"></video>

<script type="text/javascript">
   let video = document.getElementById('video');
   video.currentTime = 501.95;
</script>
```

This renders as follows:

![chrome](https://user-images.githubusercontent.com/2457311/139894425-71541ef9-b358-48ff-b419-496a0f59dacf.jpg)
(Chrome)

![firefox](https://user-images.githubusercontent.com/2457311/139894524-80c392a1-074a-4407-886c-59b4de85cf0f.jpg)
(Firefox)

![output](https://user-images.githubusercontent.com/2457311/139894628-265b8f5e-916c-4c5c-ac2a-0facf4e76079.jpg)
(`ffmpeg -ss 00:08:21.95 -i bbb_sunflower_1080p_60fps_normal.mp4 -vframes 1 -q:v 2 output.jpg`)

mzur:
Example video:

https://user-images.githubusercontent.com/2457311/140070681-fe8042f0-73c7-47ce-b62b-e0ff995769b5.mp4

Updated code:
```html
<video id="video" src="bbb-clip.mp4"></video>


<script type="text/javascript">
   let video = document.getElementById('video');
   video.currentTime = 53.05;
</script>
```

mzur:
Filed a bug for Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1739125

mzur:
This issue was confirmed to be a Firefox bug. Maybe we should show a warning for users who use Firefox in a version that is affected by this.

mzur:
[...]

PR #1531, opened 2026-07-28
Always show video frame at or before timestamp

### LINKED ISSUE #1492: Largo and annotation tool can produce different frames in different video decoders
The largo-thumbnail of a video annotation doesn't always match the frame of the actual annotation. Example:

Use this video as input

https://github.com/user-attachments/assets/2d727ddb-c267-4fd0-b457-0b88b7391b50

Create annotation at 00:01.460 where the circle is touching a line on the left (frame 14)

Largo left, annotation right:

Image

Image

Largo produced the thumbnail with something similar to `ffmpeg -i video.mp4 -ss 00:01.460 -frames:v 1 screenshot.png`. `ffmpeg` rounds to the next frame. Most browsers don't. Thus the mismatch.

(See #391 for more details). @mzur This is the first child issue from that issue.

yannik131:
Closes #1492


yannik131:
@mzur 
- mediabunny displays frame for paused video. Small visual glitch/delay when the video is paused and mediabunny displays a different frame than the last one the browser rendered, but that is as expected
- mediabunny-like ffmpeg command for thumbnail generation
- some unrelated tests had a division by zero error on my machine, quick fixes

Since my free plan can barely handle 2-3 reviews, from now on I'll just request a review from you and you can request an AI review if your budget allows it, I'll revisit the PR on my own if you do that.

resources/assets/js/videos/components/videoScreen/videoPlayback.vue
```diff
@@ -76,11 +82,13 @@ export default {
...
-                this.renderCurrentTime = this.video.currentTime;
-                this.videoContext.drawImage(this.video, 0, 0, this.videoCanvas.width, this.videoCanvas.height);
-                this.videoSource.changed();
+            if (!force && this.renderCurrentTime === this.video.currentTime) {
+                return;
             }
+
+            this.renderCurrentTime = this.video.currentTime;
+            this.videoContext.drawImage(this.video, 0, 0, this.videoCanvas.width, this.videoCanvas.height);
+            this.videoSource.changed();
```
mzur, line 91: Reuse `drawBitmap()` here?

resources/assets/js/videos/components/videoScreen/videoPlayback.vue
```diff
@@ -138,8 +153,12 @@ export default {
             // Update the layer (dimensions) if a new video is loaded.
             this.video.addEventListener('loadedmetadata', this.updateVideoLayer);
         },
-        handleSeeked() {
-            this.renderVideo(true);
+        async handleSeeked() {
+            if (this.video.paused) {
+                await this.renderPausedFrame();
+            } else {
+                this.renderVideo(true);
```
mzur, line 160: Is the `else` case still needed for anything? The old workaround was to seek to `currentTime` to force the browser to jump to the frame that will be displayed at that time (see `setPausedAndSeek()`). This was done to make sure the same frame will be shown next time based on the timestamp that was stored for an annotation.

This didn't fix the cross-browser and decoding inconsistencies that we want to fix with this PR. But maybe the `currentTime = currentTime` trick is no longer required with this PR, too? If we remove this we have to call renderPausedFrame directly on the pause event. Also we should make the process of decoding with Mediabunny visible in the UI somehow (currently the seeking state does this).

Ideally we can close https://github.com/biigle/core/issues/433 as part of this PR too.

resources/assets/js/videos/components/videoScreen/videoPlayback.vue
```diff
@@ -243,6 +262,62 @@ export default {
...
+                this.mediabunnyInput = new Input({
+                    source: new UrlSource(this.video.src),
+                    formats: ALL_FORMATS
+                });
+                const track = await this.mediabunnyInput.getPrimaryVideoTrack();
+                this.mediabunnySink = new VideoSampleSink(track);
+            } catch (e) {
+                this.useMediabunnyFallback = false;
+                throw e;
+            }
```
mzur, line 284: When can this fail? I'd rather fail hard if Mediabunny is not available than to display wrong frames again as fallback (i.e. Mediabunny should be the default method, not the fallback). What do you think?

resources/assets/js/videos/components/videoScreen/videoPlayback.vue
```diff
@@ -243,6 +262,62 @@ export default {
...
+        async getFrameBitmap(time) {
+            if (!this.useMediabunnyFallback || !this.mediabunnySink) {
+                return;
+            }
+
+            const roundedTime = Math.round(time * 100) / 100;
+            if (this.cachedTime === roundedTime && this.cachedBitmap) {
+                this.drawBitmap(this.cachedBitmap);
+                return;
+            }
```
mzur, line 295: This rounds to 10s of ms, right? If decoding is fast, shouldn't we skip rounding/caching and use the raw timestamp for the sake of correctness instead?

app/Jobs/ProcessAnnotatedVideo.php
```diff
@@ -119,43 +124,104 @@ protected function updateOrCreateFeatureVectors(Collection $annotations, \Genera
...
+     * @param string $sourcePath
      * @param float $time
      * @param int $trySeek
      *
      * @return Image
      */
-    protected function getVideoFrame(Video $video, float $time, int $trySeek = 60)
+    protected function getVideoFrame(string $sourcePath, float $time, int $trySeek = 60)
     {
+        $fps = max($this->getVideoFps($sourcePath), 0.0001);
```
mzur, line 144: This should be done outside the loop in `processAnnotationChunk` and maybe even before the chunking in `handleFile`. The fps can be cached once for the whole job.

app/Jobs/ProcessAnnotatedVideo.php
```diff
@@ -119,43 +124,104 @@ protected function updateOrCreateFeatureVectors(Collection $annotations, \Genera
...
+                        $decodeDuration,
+                        escapeshellarg($sourcePath),
+                        $timeString,
+                        escapeshellarg($outputPath)
+                    ))
+                    ->throw();
+
+                $buffer = File::get($outputPath);
+            } catch (Throwable $e) {
+                $buffer = '';
```
mzur, line 180: Should this really be handled as empty buffer instead of failing immediately? If there is something wrong with the video, the job retries 60 times before giving up.

resources/assets/js/videos/components/videoScreen/videoPlayback.vue
```diff
@@ -252,6 +327,15 @@ export default {
...
             }
         },
+        video: {
+            immediate: true,
+            deep: true,
+            handler() {
+                this.video.addEventListener('loadedmetadata', () => {
+                    this.initMediabunny();
+                });
+            }
```
mzur, line 337: Since the same video object is reused and only `src` is updated, it should be enough to add the listener once in `created()`.

app/Jobs/ProcessAnnotatedVideo.php
```diff
@@ -119,43 +124,104 @@ protected function updateOrCreateFeatureVectors(Collection $annotations, \Genera
...
+                return Image::newFromBuffer($buffer);
+            }
+
             $trySeek -= 1;
             // Roughly estimated framerate of 30 fps. With 60 iterations, we seek back up
             // to 2 s by default (this is based on what was required for edge cases in
             //  1.5M annotations on 16k videos).
-            $time = max(0, $time - 0.033333333);
-        } while (empty($buffer) && $trySeek > 0);
+            $time = max(0, $time - (1 / 30.0));
```
mzur, line 195: We now know the actual framerate so we can use exact steps.

mzur (commented):
This looks very promising and decoding with Mediabunny seems as fast as the old "seek to currentTime on pause" trick. I hope that we can close all the issues related to frame-accuracy as part of this PR. The warning message for Firefox should be removed as part of this PR, too.

Did you test for correctness with the artificial videos again?

When I reload the page and the video is seeked to a previous time, I sometimes get the first video frame instead. Looks like an async race issue:

https://github.com/user-attachments/assets/238b6b72-9237-4b91-903b-258f9f38ae60

More comments below.

mzur:
Please also remove the error message for Firefox users ([ref](https://github.com/biigle/core/issues/391#issuecomment-5795815479)) as part of this PR.

PR #1533, opened 2026-07-31
235 annotation guideline backend annotation controllers

dbrembilla:
This PR adds, as specified in PR #1390, a validate method for `AnnotationGuideline` as well as the backend for `AnnotationsControllers` and `AnnotationLabelControllers`. Adds tests for each of them, multiple edge cases.

### DIFF
--- a/app/Http/Controllers/Api/ImageAnnotationBulkController.php
+++ b/app/Http/Controllers/Api/ImageAnnotationBulkController.php
@@ -2,6 +2,7 @@
 
 namespace Biigle\Http\Controllers\Api;
 
+use Biigle\AnnotationGuideline;
 use Biigle\Http\Requests\StoreImageAnnotations;
 use Biigle\ImageAnnotation;
 use Biigle\ImageAnnotationLabel;
@@ -61,7 +62,8 @@ class ImageAnnotationBulkController extends Controller
      *        "shape_id": 1,
      *        "label_id": 1,
      *        "confidence": 1.00,
-     *        "points": [10, 11]
+     *        "points": [10, 11],
+     *        "annotation_guideline_id": 48
      *     },
      *     {
      *        "image_id": 321,
@@ -78,9 +80,24 @@ class ImageAnnotationBulkController extends Controller
      */
     public function store(StoreImageAnnotations $request)
     {
-        $annotations = collect($request->all())->map(function ($input) {
+        //Avoid having to find guideline repeatedly
+        $annotationGuidelines = [];
+        $annotations = collect($request->all())->map(function ($input) use ($annotationGuidelines) {
+            $annotationGuideline = null;
+            if (($annotationGuidelineId = $input['annotation_guideline_id'] ?? null) !== null) {
+                /** @phpstan-ignore nullCoalesce.offset */
+                $annotationGuidelines[$annotationGuidelineId] ??= AnnotationGuideline::findOrFail($annotationGuidelineId);
+                $annotationGuideline = $annotationGuidelines[$annotationGuidelineId];
+            }
+            $shapeId = $input['shape_id'];
+            $labelId = $input['label_id'];
+
+            if (!is_null($annotationGuideline) && !$annotationGuideline->validate($labelId, $shapeId)) {
+                abort(422, "Annotation uncompatible with enforced Annotation Guideline");
+            }
+
             $annotation = new ImageAnnotation;
-            $annotation->shape_id = $input['shape_id'];
+            $annotation->shape_id = $shapeId;
 
             try {
                 $annotation->validatePoints($input['points']);
@@ -88,10 +105,11 @@ class ImageAnnotationBulkController extends Controller
                 throw ValidationException::withMessages(['points' => [$e->getMessage()]]);
             }
 
+
             $annotation->points = $input['points'];
             $annotation->image_id = $input['image_id'];
             /** @phpstan-ignore property.notFound */
-            $annotation->label_id = $input['label_id'];
+            $annotation->label_id = $labelId;
             /** @phpstan-ignore property.notFound */
             $annotation->confidence = $input['confidence'];
 
--- a/app/Http/Controllers/Api/ImageAnnotationController.php
+++ b/app/Http/Controllers/Api/ImageAnnotationController.php
@@ -114,7 +114,6 @@ class ImageAnnotationController extends Controller
      * }
      */
 
-
     /**
      * Displays the annotation.
      *
--- a/app/Http/Controllers/Api/ImageAnnotationController.php
+++ b/app/Http/Controllers/Api/ImageAnnotationController.php
@@ -96,6 +96,26 @@ class ImageAnnotationController extends Controller
     }
 
     /**
+     * @api {get} annotations/:id Get an annotation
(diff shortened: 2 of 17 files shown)


app/Http/Controllers/Api/ImageAnnotationBulkController.php
```diff
@@ -78,20 +80,36 @@ class ImageAnnotationBulkController extends Controller
...
+            if (($annotationGuidelineId = $input['annotation_guideline_id'] ?? null) !== null) {
+                /** @phpstan-ignore nullCoalesce.offset */
+                $annotationGuidelines[$annotationGuidelineId] ??= AnnotationGuideline::findOrFail($annotationGuidelineId);
+                $annotationGuideline = $annotationGuidelines[$annotationGuidelineId];
+            }
+            $shapeId = $input['shape_id'];
+            $labelId = $input['label_id'];
+
+            if (!is_null($annotationGuideline) && !$annotationGuideline->validate($labelId, $shapeId)) {
+                abort(422, "Annotation uncompatible with enforced Annotation Guideline");
```
mzur, line 96: This must be more detailed. Users need to know if the shape or the label was wrong and which annotation was invalid (the index in the input array).

In general, this validation step should be performed in the `StoreImageAnnotations` request class.

app/Http/Controllers/Api/ImageAnnotationBulkController.php
```diff
@@ -78,20 +80,36 @@ class ImageAnnotationBulkController extends Controller
...
+            $annotationGuideline = null;
+            if (($annotationGuidelineId = $input['annotation_guideline_id'] ?? null) !== null) {
+                /** @phpstan-ignore nullCoalesce.offset */
+                $annotationGuidelines[$annotationGuidelineId] ??= AnnotationGuideline::findOrFail($annotationGuidelineId);
+                $annotationGuideline = $annotationGuidelines[$annotationGuidelineId];
+            }
+            $shapeId = $input['shape_id'];
+            $labelId = $input['label_id'];
+
+            if (!is_null($annotationGuideline) && !$annotationGuideline->validate($labelId, $shapeId)) {
```
mzur, line 95: Also this does not fail if a project has an enforced guideline but the client did not pass the guideline ID, right? If an annotation should be created and all the parent projects have enforced guidelines, the request must specify one of the guidelines.

app/Http/Controllers/Api/ImageAnnotationController.php
```diff
@@ -223,10 +226,19 @@ public function store(StoreImageAnnotation $request, LabelBotService $labelBotSe
...
+
+        if ($request->has('annotation_guideline_id')) {
+            $annotationGuideline = AnnotationGuideline::findOrFail($request->input('annotation_guideline_id'));
+        }
+
         if ($request->has('label_id')) {
             $label = Label::findOrFail($request->input('label_id'));
+            if (!is_null($annotationGuideline) && !$annotationGuideline->validate($label->id, $annotation->shape_id)) {
+                abort(422, "Annotation uncompatible with enforced Annotation Guideline");
+            }
```
mzur, line 239: Same here. Validation should be done in the request, it needs a better error message and it should fail if all projects have an enforced guideline and none were chosen.

app/Http/Controllers/Api/ImageAnnotationController.php
```diff
@@ -300,7 +316,19 @@ public function update(Request $request, $id)
 
         // from a JSON request, the array may already be decoded
         $points = $request->input('points', $annotation->points);
-        $annotation->shape_id = $request->input('shape_id', $annotation->shape_id);
+        $shapeId = $request->input('shape_id', $annotation->shape_id);
+
+        $annotationGuideline = null;
+        if ($request->has('annotation_guideline_id')) {
```
mzur, line 322: Same.

app/Http/Controllers/Api/ImageAnnotationBulkController.php
```diff
@@ -78,20 +80,36 @@ class ImageAnnotationBulkController extends Controller
...
     public function store(StoreImageAnnotations $request)
     {
-        $annotations = collect($request->all())->map(function ($input) {
+        //Avoid having to find guideline repeatedly
+        $annotationGuidelines = [];
+        $annotations = collect($request->all())->map(function ($input) use ($annotationGuidelines) {
+            $annotationGuideline = null;
+            if (($annotationGuidelineId = $input['annotation_guideline_id'] ?? null) !== null) {
+                /** @phpstan-ignore nullCoalesce.offset */
+                $annotationGuidelines[$annotationGuidelineId] ??= AnnotationGuideline::findOrFail($annotationGuidelineId);
```
mzur, line 89: You don't check if the guideline actually belongs to the project of the annotation.

app/Http/Controllers/Api/ImageAnnotationBulkController.php
```diff
@@ -78,20 +80,36 @@ class ImageAnnotationBulkController extends Controller
      */
     public function store(StoreImageAnnotations $request)
     {
-        $annotations = collect($request->all())->map(function ($input) {
+        //Avoid having to find guideline repeatedly
+        $annotationGuidelines = [];
+        $annotations = collect($request->all())->map(function ($input) use ($annotationGuidelines) {
```
mzur, line 85: `$annotationGuidelines` is called per-value during the map so the changes do not persist across method calls. Use:


```suggestion
        $annotations = collect($request->all())->map(function ($input) use (&$annotationGuidelines) {
```

app/Services/LabelBot/LabelBotService.php
```diff
@@ -45,6 +48,9 @@ public function getLabelsForAnnotation(
         Cache::increment($cacheKey);
         try {
             $topNLabels = $this->performVectorSearch($featureVector, $treeIds, $model);
+            if (!is_null($annotationGuideline)) {
+                $topNLabels = array_filter($topNLabels, fn ($labelId) => $annotationGuideline->validate($labelId, $shapeId));
```
mzur, line 52: Ideally the vector search could take the allowed shape and labels into account right away, so users can still get 3 suggestions (just more unlikely) instead of 3 minus the disallowed ones.

app/Http/Controllers/Api/ImageAnnotationController.php
mzur, line 315: `annotation_guideline_id` is not validated at all.

mzur (commented):
I didn't do a detailed review yet because there are some general issues that have to be addressed first (see below). This affects all controllers; I didn't add a comment everywhere.

Also, did you intentionally leave out changes required for Largo? Will you do these in a separate PR?

This needs a lot of care because it requires changes in the most important code paths of BIIGLE. Do you still have time for this? Otherwise I can take this over, too.

dbrembilla:
Hi @mzur 

> Also, did you intentionally leave out changes required for Largo? Will you do these in a separate PR?

Yes I did! This was already quite massive and did not want to miss important aspects within BIIGLE because, as you said, it is quite critical.

> Do you still have time for this? Otherwise I can take this over, too.

I don't think I will have time to look at it for at least a month between vacation days and focus. Maybe at the end of October I will have time to look deeper into it. But if you have time, you can of course do!

Issue #255, opened 2020-07-09
Improved annotation filtering


mzur:
Michael Kloster requested an improved annotation filtering mechanism in the annotation tool. He wanted to create filter rules like "all annotations with label *a* or *b*". The current filtering mechanism allows only one of label, user, shape or annotation session. Tim had a nice idea for a possible UI for this. The "label" option is removed from the current filter dropdown in the annotations tab. Instead, the user can click on the colored circle of the label in the annotation list below (see image below) to make this label "invisible". This is similar to toggling the visibility of layers in Gimp or Photoshop. The visibility can be indicated by a [closed eye icon](https://fontawesome.com/icons/eye-slash) and muted text. There should be a "show all" and "hide all" button, too. The underlying code for filtering should stay the same with the addition of supporting multiple labels at once. It should be possible to combine the label filtering with one of the user, shape or annotation session filters.

![image](https://user-images.githubusercontent.com/2457311/45219565-9b552380-b2ab-11e8-925d-fc33fc26b258.png)

We can't implement this right now but it might be an important feature in WP 3.2 of MiningImpact2. If we implement a display of image mosaics and then show annotations on this mosaic, a filtering mechanism as described above may be necessary.

mzur:
With biigle/geo#18 this is no longer relevant to MiningImpact2. It is solely related to the filtering in the annotation tool.

mzur:
A user also asked for filtering my label *and* shape for example. So if the filtering mechanism is improved, it should allow arbitrary combinations of label, user, shape or annotation session (and/or) similar to the volume overview.

Issue #241, opened 2020-07-08
Filter by time/altitude


mzur:
If the `taken_at` or `gps_altitude` information is available for (all) images of a volume, offer appropriate filters for these attributes. Users should be able to filter the images by the time (range) and by the altitude (range) they were taken.

mzur:
This should be generalized to all metadata attributes that BIIGLE supports.

PR #1549, opened 2026-08-18
Largo filter annotation date

### LINKED ISSUE #1166: Filter annotations in Largo

dbrembilla:
Resolves #1166 
Creates a `created_at` and `updated_at` filter for Largo. It was not straightforward as an implementation. The solution I found was to use an array 
```
[
  created_at => [
    ref => 'annotation'
    operator => 'gt'
    date=>'2025-01-02'
]
]
```
with ref as `annotation` or `annotation_label`,  'gt', 'lt', 'eq', 'neq' to explain whether the annotation is after, before, on a day or not on a day. For `updated_at` is the same.

### DIFF
--- a/app/Http/Controllers/Api/Projects/FilterImageAnnotationsByLabelController.php
+++ b/app/Http/Controllers/Api/Projects/FilterImageAnnotationsByLabelController.php
@@ -7,7 +7,6 @@ use Biigle\Http\Requests\FilterProjectAnnotationsRequest;
 use Biigle\ImageAnnotation;
 use Biigle\Project;
 use Biigle\Traits\CompileFilters;
-use Illuminate\Http\Request;
 
 class FilterImageAnnotationsByLabelController extends Controller
 {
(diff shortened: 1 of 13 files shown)


mzur (commented):
The UI looks great but I still have two high-level comments before I look at the code:

- I'm not sure if the "updated" variants of the filter are useful. IIRC an annotation is updated only if the shape is modified. Do we really anticipate cases where users want to filter by the date when an annotation shape was modified? An annotation label is never actually updated. There is an update API endpoint (for the confidence) but this is a relict of the past and we never implemented the confidence in the UI.

- I think we can stick to the way the other filters are encoded. They use e.g. `shape_id=5` or `shape_id=-5` (for "is" and "is not"). Here we could use:
    - `annotation_created_at=<2026-09-02` (less than)
    - `annotation_created_at=2026-09-02` (equals)
    - `annotation_created_at=>2026-09-02` (greater than)
    - `annotation_created_at=-2026-09-02` (not equals)
    
    The "ref" can be encoded in the original key like "annotation_created_at" or "annotation_label_created_at". I don't know about if this makes the code more complicated but to me this is more consistent with the existing filter interface. And it may still be a good template to implement more advanced filters in the volume overview or the annotation tool (https://github.com/biigle/core/issues/241, https://github.com/biigle/core/issues/255).

PR #1551, opened 2026-08-25
Remove static models

### LINKED ISSUE #1048: Pull "static models" out of the DB
There are several "static models" such as \Biigle\Role that are never changed. They only have an ID and a name and exist as foreign keys on several tables. To speed things up, these models are completely cached. But they don't have to exist in the database at all. They could basically be enums with fixed IDs and values. The IDs are put into the database as well but not as foreign keys to a table. This way, no foreign key checks are required, no queries are required and no caching is required.

Implement the static models as enums (or similar) with ID and name. Implement a migration that first updates all existing foreign key colums so the previous IDs of the static models match the new hard-coded IDs. Then drop the static model database tables.

Static models in biigle/core:
- \Biigle\MediaType
- \Biigle\Role
- \Biigle\Shape
- \Biigle\Visibility
- \Biigle\ReportType

Static models in biigle/maia:
- MaiaJobState

yannik131:
Closes #1048


yannik131:
### Migration guide for me

Turning eloquent model `X` into enum `X` requires these changes:

### Enum implementation
Assuming all possible model values are written into the corresponding `X` table upon migration

- Find values by connecting to the db (`docker compose exec database psql -U biigle -d biigle -c "select * from X;"`) and by looking at `DB::table('X')->insert` statements, insertion order corresponds to values, 1-indexed
- Assign integer values to each existing case using the same values and implement
- `cId()` returning the value of case `c` for each case
- `c()` returning the case `c`for each case
- `label()` returning the case name like it was previously written in the DB
- `toArray()` returning `['id' => xId(), 'name' => label()]`

### Creating the migration
Find all foreign keys and their constraint names using `docker compose exec database psql -U biigle -d biigle -c "\d+ X"`, then

- (make sure enum and database table are 1:1 congruent on present IDs/values as a failsafe: MAYBE if migrations don't fill the tables completely?)
- drop all fk constraints
- drop the table itself

### Editing old migrations
Instead of removing the tables from the migrations, it's easier to keep them and replace eloquent model method calls `X::method()` (`all()`, `insert()`, ...) with `DB::table('X')->method()`.

### Common needed replacements
`X` is the name of the model and enum, `x` is lowercase name, `x_id` is (usually) the name of the FK
- `X::all()` -> `collect(X::cases())->map->toArray()` (add `->values()` if PHPStan is pedantic)
- `$val = X::findOrFail($caseId)` -> `$val = X::tryFrom((int) $caseId); abort_if($val === null, 404);`
- similarly, other model-specific query patterns X::find(), whereHas(), with(), whereRelation()
- `collect([X::case1(), X::case2(), ..])`-> `collect(...)->map->toArray()`
- `FormRequest` validation rules like `'x_id' => 'integer|exists:x,id'` -> `'x_id' => ['integer', Rule::in(array_column(X::cases(), 'value'))]`
- `$this->belongsTo(X::class)` -> `X::from($this->x_id)`
- Eloquent model has `x()` method -> `getXAttribute(): X`. This way `$object->x` still works
- blade files: `{!! Biigle\X::x() !!}` -> `@json(Biigle\X::x()->toArray())`
- `$x->id` -> `$x->value`
- `$x->name` -> `$x->label()`
- `$object->x()->associate(Biigle\X::case())` -> `$object->x_id = X::caseId()`
- remove DB-specific `X`-tests
- `XTest::create()` -> `X::case()` (find out default case or just choose one)
- `X::pluck('id', 'name')` -> `collect(X::cases())->mapWithKeys(fn (X $x) => [$x->label() => $x->value])`
- Delete factory `XFactory`, replace `X::factory()` calls with some default `X::case1Id()`
- Remove eager loading statements `->with('x')` from DB queries

yannik131:
@mzur I've started working on the maia code, but this PR is ready for review. It shouldn't be merged before the maia PR though. Most of the changes are the result of the above migration guide I created for myself.  There are a few TODOs scattered through the code, but they could maybe just be removed. And I can't really test the entirety of the application, but the tests pass and the basic things I tried locally (annotations, reports, admin area, ...) worked without issues. I've long been thinking about creating an E2E regression test with something like [Cypress](https://www.cypress.io/#create) for regression cases like this, what is your opinion on that? Just a single test that logs in and tests some basic things (create a project, open image/video annotation tool, create an annotation, generate a report, ...). 
I would like to put the enums into a separate Enums/ directory. I just didn't do it yet because it would affect hundreds of files and make the PR difficult to review, so I'll do it last when everything else is settled.

app/Http/Requests/StoreImageAnnotation.php
```diff
@@ -52,7 +53,7 @@ function ($attribute, $value, $fail) {
                 },
             ],
             'confidence'  => 'required|numeric|between:0,1',
-            'shape_id' => 'required|integer|exists:shapes,id',
+            'shape_id' => ['required', 'integer', Rule::in(Shape::pluckById()->keys()->all())],
```
mzur: Isn't this equivalent to:
(also for other validation cases)


```suggestion
            'shape_id' => ['required', 'integer', Rule::enum(Shape::class)],
```

app/Http/Requests/StoreImageAnnotations.php
```diff
@@ -81,7 +81,7 @@ public function authorize()
     public function rules()
     {
         // Image annotations cannot have the whole frame shape.
-        $shapeIds = Shape::whereKeyNot(Shape::wholeFrameId())->pluck('id');
+        $shapeIds = Shape::pluckById(Shape::wholeFrame())->keys();
```
mzur: A little confusing syntax. Maybe use:


```suggestion
        $shapeIds = Shape::pluckById(except: Shape::wholeFrame())->keys();
```

app/MediaType.php
```diff
@@ -2,31 +2,74 @@
...
+            'id' => $this->value,
+            'name' => $this->label()
+        ];
+    }
+
+    #[Override]
+    public function jsonSerialize(): mixed
+    {
+        return $this->toArray();
+    }
```
mzur: Maybe put this into a trait (or Laravel already has something like this)?

app/MediaType.php
```diff
@@ -2,31 +2,74 @@
...
-     *
-     * @var array<string, string>
-     */
-    const INSTANCES = [
-        'image' => 'image',
-        'video' => 'video',
-    ];
+    // Values previously used in the DB model, kept for compatibility
+    case IMAGE = 3;
+    case VIDEO = 4;
```
mzur: You can't rely on these IDs in the DB. The migration must override the existing IDs in the DB to match the enums in all cases. The enums can have new IDs starting at 1.

database/migrations/2026_08_25_191931_replace_roles_with_enum.php
mzur, line 1: This should be called something like `replace_role_table_with_enum`. If it's called `remove_roles` it implies something completely different.

database/migrations/2026_08_25_191931_replace_roles_with_enum.php
```diff
@@ -0,0 +1,77 @@
...
+    ];
+
+    /**
+     * Run the migrations.
+     */
+    public function up(): void
+    {
+        // TODO Discuss if we want to "be safe" by mapping here or if it's enough
+        // to check in psql/with the migrations that the enum used the same numbers
+        // as the db table, in which case mapping is unnecessary
```
mzur: The mapping is essential for all static models (see above).

database/migrations/2026_08_25_191931_replace_roles_with_enum.php
```diff
@@ -0,0 +1,77 @@
...
+
+        // Replace foreign keys with the above IDs
+        foreach ($this->foreignKeys as [$table, $column]) {
+            foreach ($map as $oldId => $newId) {
+                DB::table($table)
+                    ->where($column, $oldId)
+                    ->update([$column => $newId]);
+            }
+
+            Schema::table($table, fn (Blueprint $t) => $t->dropForeign([$column]));
```
mzur: Isn't this required before you update the IDs to something potentially invalid above?

resources/views/label-trees/show/members.blade.php
```diff
@@ -4,7 +4,7 @@
 <script type="module">
     biigle.$declare('labelTrees.members', {!! $members !!});
     biigle.$declare('labelTrees.roles', {!! $roles !!});
-    biigle.$declare('labelTrees.defaultRole', {!! Biigle\Role::editor() !!});
+    biigle.$declare('labelTrees.defaultRole', @json(Biigle\Role::editor()->toArray()));
```
mzur, line 7: Maybe also add `implements Stringable` to the enums?

mzur (commented):
I'm thinking if we switch to this new pattern we could just as well embrace it completely although it may be a lot of work. On the other hand, this could be a lot of work with only marginal performance gains (the performance impact of cached models could be minimal although it saves network roundtrips to the cache). It could be mostly a logical refactor without tangible gains...

Since you already put some work into this: Do you feel the change is worth it? We could still pull the plug before putting even more work into this.

If we go ahead:

I think the cleanest approach would be to rename columns like `volumes.media_type_id` to `volumes.media_type` and then use [enum casting](https://laravel.com/framework/docs/eloquent-mutators#enum-casting) for this property. This way we don't have to implement custom accessors for every enum property.

Instead of implementing the enums to mimic the old static models (with `MediaType::video()` or `MediaType::videoId()` methods, for example) I'd fully switch to enum syntax (i.e. `MediaType::VIDEO` and `MediaType::VIDEO->value`). This would require updates to every place a static model was used (including modules).

Let's discuss the high-level comments first. I didn't go into too much detail below.

mzur:
Regarding E2E integration tests: I'd like to stick with testing the backend/API only because I feel full UI/E2E tests would quickly become a burden to maintain with our little resources. If people start paying for BIIGLE to be reliable in business-critical applications, we can reconsider this.

Besides, since I just noticed this: In the future, please try to align your commit messages to the [convention](https://chris.beams.io/git-commit) we use.

app/MediaType.php
yannik131: I was wondering about that. Given the migrations execute changes in a defined order and that `insert` statements result in 1-indexed IDs also with a defined order I assumed the DB-values are the same everywhere? Would be nice for backwards-compatibility.

mzur: It can be different if migrations are rolled back and reapplied because the sequence for the incrementing IDs is not reset during rollback.

yannik131:
@mzur I agree with your suggestions, and I don't think it will take more than a couple of hours to add these other changes. Shouldn't be a problem.
But do feel free to change priorities of issues on the [roadmap](https://github.com/orgs/biigle/projects/4/views/9) if you think some issues are more important than others. I always pick issues with higher priority first, and no issue currently has "low" priority.

mzur:
I think this is not an issue of priorities (besides that "low" issues are basically never worked on), it's more my habit of writing down ideas as issues without always giving much though about gain/effort. So we have issues with medium priority that we definitely need sitting next to issues with the same priority that we should be more careful with. I'm not sure how to deal with it other than asking you to think about implications first and talking to me if you think gain/effort is too low.

P.S.: You find issues with low priority in the [full roadmap](https://github.com/orgs/biigle/projects/4/views/1). The task list is already the version filtered by me.

yannik131:
**Migration guide for removing `_id` suffixes**

Use VS Code regex search+replace:

1. Add enum casting to `casts()`: `'x_id' => X::class`
2. Search for `->x_id` and check if `->value` is still applicable after enum casting
3. `X::(\w+)Id\(\)` -> `X::\U$1->value` (`\U` for uppercasing the match)
4. `X::(\w+)\(\)` -> `X::\U$1` 
5. `x_id` -> `x`

I only replaced`XId` -> `X` and `xId` -> `x` in case of error because they used to refer to integer-only values, whereas `x` and `X` are often int or `X`.

yannik131:
@mzur I've been working on this for several days now and I noticed:
1. Replacing `X::case()` with `X::CASE`: Easy
2. Replacing `X::caseId()` with `X::CASE->value`: Easy
3. Adding enum casting `'x_id' => X::class`: Moderate
4. Removing the `_id` suffix from the DB column and subsequently changing thousands of variables, comments, interfaces, DB queries, JSON keys, etc.: Very hard.

I propose not to rename the columns but to do the other things. Of course then the variable name will be kind of unfortunate..
However, I did manage to do all 4 steps for `Role` and `MediaType` (but I'm not 100% sure it will work in all circumstances) and if you like, I'll keep working on renaming the other columns.

mzur:
Can we write a guide how to migrate the `_id` attributes and then break these down into individual issues (one for each model or even attribute?). We don't have to change everything at once in this PR. With a thorough guide we can work through the issues step by step and be reasonably confident that all relevant code was updated and works.

yannik131:
@mzur Nvm I did it! It's done! I renamed all afflicted columns, applied enum casting and removed the enum helpers. Including the `MaiaJobState` enum (https://github.com/biigle/maia/pull/241). In some cases, due to generic search-replace,`X::case->value` is used where `X::case` might be more appropriate, but that's just a cosmetic problem. However, there might still be frontend issues or other things not covered by the tests. Ready for review. 
I suggest to test this PR (together with the Maia PR) on the staging area.
Also, while I did include `down()` methods for all migrations, I'm not sure we actually need them since this is designed to be destructive. If we do want them, they should also be tested somehow (revert the migrations, then checkout a previous commit). 

Edit: According to ChatGPT, the modules ananas, user-disks, user-storage and metadata-coco would be broken by this PR. I'll create the necessary PRs in these modules as well, and all of them should be tested together in the staging area. I'll try to hurry to avoid future merge-conflicts, as any PR is probably going to conflict with these changes.

Edit 2: Some pivot tables (such as `project_user` or `label_tree_user`) could also have enum casting enabled by doing something like 

```php
class ProjectUser extends Pivot
{
    protected function casts(): array
    {
        return [
            'project_role' => Role::class,
        ];
    }
}

class Project 
{
    public function users()
    {
        return $this->belongsToMany(User::class)
            ->using(ProjectUser::class)
            ->withPivot('project_role');
    }
}
```

Not sure if we need that though as these pivots aren't instantiated anywhere.

resources/views/label-trees/show/members.blade.php
yannik131: I don't see the usecase here and it's less explicit than calling `->label()`, or where do you think it would be nice to have?

yannik131:
@mzur Related PRs are:
- ananas: https://github.com/biigle/ananas/pull/69
- metadata-coco: https://github.com/biigle/metadata-coco/pull/16
- user-disks: https://github.com/biigle/user-disks/pull/60
- user-storage: https://github.com/biigle/user-storage/pull/90
- maia: https://github.com/biigle/maia/pull/241

Not sure if I can list them natively in github somewhere

mzur:
Wow, that's great work. I'll look at the PRs now and try to merge them asap to avoid conflicts.

Regarding the pivot models: As far as i remember we don't use pivot models anywhere so there is no need to introduce them here.

app/Http/Controllers/Api/Projects/FilterImageAnnotationsByLabelController.php
```diff
@@ -54,7 +54,7 @@ public function index(Request $request, $pid, $lid)
 
         $take = $request->input('take');
         $filters = [
-            'shape_id' => $request->input('shape_id'),
+            'shape' => $request->input('shape'),
```
mzur: I would try to keep the public API unchanged as much as possible. Here, the public argument can stay `shape_id` which is internally mapped to `shape` in the `$filters` array.

mzur: I thought about keeping the old attributes (e.g. `shape_id`) as deprecated fallback for all API endpoints. But on the other hand maybe we should do a clean break so (script) users will notice that they may have to use different IDs now. What's your opinion?

app/Http/Controllers/Api/MediaTypeController.php
```diff
@@ -54,6 +54,8 @@ public function index()
      */
     public function show($id)
     {
-        return MediaType::findOrFail($id);
+        $mediaType = MediaType::tryFrom((int) $id);
+        abort_if($mediaType === null, 404);
+        return $mediaType;
```
mzur: Maybe implement `findOrFail` as a convenience method for all enums like this?

```php
public static function findOrFail(int|string $id): static
{
    return static::tryFrom((int) $id) ?? abort(404);
}
```

app/Http/Controllers/Api/ProjectsAttachableVolumesController.php
```diff
@@ -65,12 +65,14 @@ public function index(Request $request, $id, $name)
...
-        $volumes->each(function ($item) use ($hidden) {
+
+        return $volumes->map(function ($item) use ($hidden) {
             $item->append('thumbnailUrl')
                 ->append('thumbnailsUrl')
                 ->makeHidden($hidden);
+            $data = $item->toArray();
+            $data['media_type'] = MediaType::from($item->media_type_id)->toArray();
+            return $data;
         });
```
mzur: I think the `media_type_id` was originally required to to make `thumbnailUrl` work (because it checks if it's an image or video volume). Now the check is done using `media_type` so we don't need the `media_type_id` alias in the query. In fact this causes a bug because `$volume->media_type` is null when `thumbnailUrl` should be appended.

We should add a test case that covers the bug scenario.

app/Http/Controllers/Views/Annotations/AnnotationToolController.php
```diff
@@ -56,7 +56,7 @@ public function show(Request $request, $id)
             })
             ->get();
 
-        $shapes = Shape::pluck('name', 'id');
+        $shapes = Shape::pluckById();
```
mzur: Maybe add `pluckById` to all new enums via trait?

app/Http/Controllers/Views/Projects/ProjectsController.php
```diff
@@ -42,14 +42,18 @@ protected function show(Request $request, $id)
...
             ->each(function ($item) use ($hidden) {
                 $item->append('thumbnailUrl')
                     ->append('thumbnailsUrl')
                     ->makeHidden($hidden);
+            })
+            ->map(function ($volume) {
+                $data = $volume->toArray();
+                $data['media_type'] = $volume->media_type->toArray();
+                return $data;
             });
```
mzur: I feel something like this could be problematic. The `media_type` can sometimes be an int and sometimes an array. Maybe we should rather update the frontend to compare the ID instead of the name? Or we append an extra `media_type_name` here.

app/Http/Controllers/Views/Projects/ProjectStatisticsController.php
```diff
@@ -32,11 +32,14 @@ public function show(Request $request, $id)
...
+            ->select('id', 'name', 'updated_at', 'media_type')
             ->orderBy('created_at', 'desc')
-            ->get();
-
+            ->get()
+            ->map(function ($volume) {
+                $data = $volume->toArray();
+                $data['media_type'] = $volume->media_type->toArray();
+                return $data;
+            });
```
mzur: same

app/Http/Controllers/Views/Volumes/VolumeReportsController.php
```diff
@@ -24,10 +24,7 @@ public function show(Request $request, $id)
         $volume = Volume::findOrFail($id);
         $this->authorize('access', $volume);
         $sessions = $volume->annotationSessions()->orderBy('starts_at', 'desc')->get();
-        $types = ReportType::when($volume->isImageVolume(), fn ($q) => $q->where('name', 'like', 'Image%'))
-            ->when($volume->isVideoVolume(), fn ($q) => $q->where('name', 'like', 'Video%'))
-            ->orderBy('name', 'asc')
-            ->get();
+        $types = ReportType::getSortedTypes($volume->isImageVolume(), $volume->isVideoVolume());
```
mzur: Could also be something like:


```suggestion
        $types = ReportType::getTypesForVolume($volume);
```

app/Support/EnumMigrationHelper.php
```diff
@@ -0,0 +1,73 @@
...
+            [$table, $column] = $foreignKey;
+            $constraint = $foreignKey[2] ?? "{$table}_{$column}_foreign";
+
+            Schema::table($table, fn (Blueprint $t) => $t->dropForeign($constraint));
+
+            foreach ($map as $oldId => $newId) {
+                DB::table($table)
+                    ->where($column, $oldId)
+                    ->update([$column => $newId]);
+            }
```
mzur: The sequential update can be problematic. Say we have a map like this:

```php
[
  "1" => 2,
  "2" => 3,
]
```
During the first loop the ID 1 is updated to 2. During the second loop, 2 is updated to 3 **but this now also includes the updated rows from the first loop**.

Instead of the loop, do a single atomic update for each table.

app/Support/EnumMigrationHelper.php
```diff
@@ -0,0 +1,73 @@
...
+     */
+    public static function replaceStaticTableWithEnum(array $map, string $tableName, array $foreignKeys)
+    {
+        foreach ($foreignKeys as $foreignKey) {
+            [$table, $column] = $foreignKey;
+            $constraint = $foreignKey[2] ?? "{$table}_{$column}_foreign";
+
+            Schema::table($table, fn (Blueprint $t) => $t->dropForeign($constraint));
+
+            foreach ($map as $oldId => $newId) {
```
mzur: If `$oldId === $newId` the update should be skipped for better performance during the migration.

app/Http/Requests/StorePendingVolume.php
```diff
@@ -44,7 +43,7 @@ public function rules(): array
         ];
 
         $parserClass = $this->input('metadata_parser', false);
-        if ($this->has('media_type') && $parserClass && ParserFactory::has($this->input('media_type'), $parserClass)) {
+        if ($this->has('media_type') && $parserClass && ParserFactory::has(MediaType::from($this->input('media_type'))->label(), $parserClass)) {
```
mzur: This should use `tryFrom` with a null check because the `media_type` was not validated yet.

resources/views/search/label-trees-content.blade.php
```diff
@@ -14,7 +14,7 @@
                     <a href="{{route('label-trees', $tree->id)}}" title="Show {{$tree->name}}">{{$tree->name}}</a>
                 @endif
             </span>
-            @if ($tree->visibility_id === Biigle\Visibility::privateId())
+            @if ($tree->visibility === Biigle\Visibility::PRIVATE->value)
```
mzur:
```suggestion
            @if ($tree->visibility === Biigle\Visibility::PRIVATE)
```

app/Support/EnumMigrationHelper.php
```diff
@@ -0,0 +1,73 @@
...
+     * foreign key constraint name is not supplied, the default name is constructed
+     * @return void
+     */
+    public static function replaceStaticTableWithEnum(array $map, string $tableName, array $foreignKeys)
+    {
+        foreach ($foreignKeys as $foreignKey) {
+            [$table, $column] = $foreignKey;
+            $constraint = $foreignKey[2] ?? "{$table}_{$column}_foreign";
+
+            Schema::table($table, fn (Blueprint $t) => $t->dropForeign($constraint));
```
mzur: Since the foreign key constraint will be dropped now, we should consider to add a `CHECK (col IN (...))` or `CHECK (col BETWEEN xxx AND yyy)` to the relevant tables.

Since adding the check could apply an exclusive lock the AI suggests to do it in two steps, e.g.:

```php
DB::statement('ALTER TABLE image_annotations ADD CONSTRAINT image_annotations_shape_check CHECK (shape BETWEEN 1 AND 6) NOT VALID');
DB::statement('ALTER TABLE image_annotations VALIDATE CONSTRAINT image_annotations_shape_check');
```
(note that we can use different checks for image and video annotation shapes now)

The first step applies the check which takes effect for new inserts but existing rows are not validated. The second step validates existing rows but with a shared lock.

app/Http/Controllers/Api/Projects/FilterImageAnnotationsByLabelController.php
mzur: This is also true for the user/label tree/volume export and import. Here we should recommend to update the source BIIGLE instance first before transferring data to the target BIIGLE instance.

tests/php/RoleTest.php
```diff
@@ -3,65 +3,56 @@
...
-    }
-
-    public function testNameRequired()
+    public function testAdmin(): void
     {
-        $this->model->name = null;
-        $this->expectException(QueryException::class);
-        $this->model->save();
+        $this->assertSame(Role::ADMIN, Role::ADMIN);
+        $this->assertSame(Role::ADMIN->value, Role::ADMIN->value);
```
mzur: Tests like these no longer make sense, please update all enum tests accordingly.

tests/php/RoleTest.php
```diff
@@ -3,65 +3,56 @@
 namespace Biigle\Tests;
 
 use Biigle\Role;
-use Illuminate\Database\QueryException;
-use ModelTestCase;
+use PHPUnit\Framework\TestCase;
 
-class RoleTest extends ModelTestCase
+// TODO Create similar tests for the other enums?
```
mzur: Yes, the enums can be tested for `toArray`, `jsonSerialize` and any other special method. The label doesn't have to be tested IMO.

run_tests.py
mzur: Should be removed

.gitignore
```diff
@@ -28,4 +28,5 @@ yarn-error.log
 /*.phar
 /*.php-cs-fixer.cache
 /.idea
-**/.DS_Store
\ No newline at end of file
+**/.DS_Store
+/passed_tests.txt
```
mzur:
```suggestion
```

database/migrations/2026_08_25_191931_replace_roles_with_enum.php
```diff
@@ -0,0 +1,54 @@
...
+    {
+        $oldIds = DB::table('roles')->pluck('id', 'name');
+        $map = [
+            $oldIds['admin'] => Role::ADMIN->value,
+            $oldIds['editor'] => Role::EDITOR->value,
+            $oldIds['guest'] => Role::GUEST->value,
+            $oldIds['expert'] => Role::EXPERT->value,
+        ];
+
+        EnumMigrationHelper::replaceStaticTableWithEnum($map, 'roles', $this->foreignKeys, true);
```
mzur: There is no fourth argument. Please update all migrations.


```suggestion
        EnumMigrationHelper::replaceStaticTableWithEnum($map, 'roles', $this->foreignKeys);
```

database/migrations/2026_08_25_191931_replace_roles_with_enum.php
```diff
@@ -0,0 +1,54 @@
...
+        });
+
+        DB::table('roles')->insert([
+            ['id' => Role::ADMIN->value, 'name' => 'admin'],
+            ['id' => Role::EDITOR->value, 'name' => 'editor'],
+            ['id' => Role::GUEST->value, 'name' => 'guest'],
+            ['id' => Role::EXPERT->value, 'name' => 'expert'],
+        ]);
+
+        EnumMigrationHelper::createForeignKeys($this->foreignKeys, 'roles', true);
```
mzur: Same


```suggestion
        EnumMigrationHelper::createForeignKeys($this->foreignKeys, 'roles');
```

app/Http/Controllers/Api/UserController.php
mzur: The role is not returned as array here, right?

resources/assets/js/videos/models/Annotation.js
```diff
@@ -15,7 +15,7 @@ export default class Annotation {
         this._frames = ref(args.frames);
         this._points = args.points;
         this.video_id = args.video_id;
-        this.shape_id = args.shape_id;
+        this.shape_id = args.shape;
```
mzur: Maybe also rename `shape_id` to `shape` in the UI?

app/Services/MetadataParsing/Annotation.php
```diff
@@ -11,19 +11,19 @@ abstract class Annotation
...
+    public int $shape;
 
     /**
      * @param Shape $shape
      * @param array<float>|array<array<float>> $points
      * @param array<LabelAndUser> $labels
      */
     public function __construct(
-        public Shape $shape,
+        Shape $shape,
```
mzur: You should discard `shape_id` here and use `shape` directly.

app/Http/Requests/StoreVolumeFile.php
```diff
@@ -39,7 +39,7 @@ public function rules()
             'files' => [
                 'required',
                 'array',
-                new VolumeFiles($this->volume->url, $this->volume->media_type_id),
+                new VolumeFiles($this->volume->url, $this->volume->media_type->value),
```
mzur: Update the rule to accept the plain media_type instead.

mzur (commented):
I applied an (AI assisted) PR that resolves the `X::case->value` cases (https://github.com/biigle/core/pull/1569). Below is a partial review (50/379 files). More will follow.

PR #1552, opened 2026-08-27
Fixed validation for concurrent metadata imports

### LINKED ISSUE #1256: Error 500 when importing multiple metadata files at once
There is an error 500 when importing metadata from two volumes (I tried with the coco import) at the same time.

parimalgavali:
## Changes

- Wrap pending-volume deletion and creation in a database transaction.
- Preserve the existing sequential replacement behavior.
- Convert concurrent unique-constraint violations into a validation
  response.
- Display the validation message on the volume metadata page.
- Add regression coverage for sequential and concurrent requests.

Closes #1256

### DIFF
--- a/app/Http/Controllers/Api/PendingVolumeController.php
+++ b/app/Http/Controllers/Api/PendingVolumeController.php
@@ -7,11 +7,11 @@ use Biigle\Http\Requests\StorePendingVolumeFromVolume;
 use Biigle\Http\Requests\UpdatePendingVolume;
 use Biigle\Jobs\CreateNewImagesOrVideos;
 use Biigle\PendingVolume;
-use Biigle\Project;
-use Biigle\Role;
 use Biigle\Volume;
 use DB;
+use Illuminate\Database\UniqueConstraintViolationException;
 use Illuminate\Http\Request;
+use Illuminate\Validation\ValidationException;
 use Queue;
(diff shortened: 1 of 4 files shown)


mzur (changes requested):
The original code intentionally deleted all pending volumes before creating the new one. The existing pending volumes are likely to be leftovers of aborted "new volume" actions. We should keep the logic of deleting existing pending volumes for a better UX.

I think it should be enough to wrap the "delete old PV then create the new PV" in a transaction and throw a validation exception on a unique constraint exception with try/catch like you already implemented. This catches the case where a user wants to actually create two metadata imports at the same time. We could reword the message to "only one metadata import can be performed at a time".

We should not reject the request as part of the validation in the form request. The user will have a hard time to find out how to fix this (i.e. go back to the create volume page and click cancel to delete the PV). The only real issue should be concurrent requests. Sequential requests should be fine even if there is a dangling PV.

app/Http/Controllers/Api/PendingVolumeController.php
mzur, line 63: Looks good now. Please apply the same transaction guard here, too.

mzur (approved):
The transaction wasn't really required in store() as it's only a single statement (my bad). I removed it again. But the try/catch is fine there.

PR #1562, opened 2026-09-08
Disallow own domain for remote volumes

### LINKED ISSUE #1445: Disallow own domain for remote volumes
Someone managed to set a volume URL to `https://biigle.de/volumes/VOLUME_ID`. First, this should not be possible after https://github.com/biigle/core/pull/1430 because this URL prepended to a filename will always return 404. Second, add a validation rule to explicitly deny the own domain in volume URLs.

parimalgavali:
## Summary

- Rejects remote volume URLs that use the hostname configured in `APP_URL`.
- Compare hostnames case-insensitively and independently of scheme, port, and path.
- Keep external domains and subdomains valid.
- Added regression tests for own-domain URLs.

### DIFF
--- a/tests/php/Rules/VolumeUrlTest.php
+++ b/tests/php/Rules/VolumeUrlTest.php
@@ -214,10 +214,6 @@ class VolumeUrlTest extends TestCase
     public function testRemoteOwnDomain()
     {
         $mock = new MockHandler([
-            new Response(200),
-            new Response(200),
-            new Response(200),
-            new Response(200),
         ]);
 
         $handler = HandlerStack::create($mock);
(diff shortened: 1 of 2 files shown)


tests/php/Rules/VolumeUrlTest.php
```diff
@@ -210,6 +211,33 @@ public function testRemoteProviderBlacklist()
         $this->assertFalse($validator->passes(null, 'https://drive.google.com'));
     }
 
+    public function testRemoteOwnDomain()
+    {
+        $mock = new MockHandler([
+            new Response(200),
+            new Response(200),
+            new Response(200),
+            new Response(200),
```
mzur: These are never used.


```suggestion
```

